PR #5, opened 2015-07-07
Hello world rpc

joshm-3b6b58:
So... I've never made a pull request. Here goes:
- refactored client side RPC so that the generic RPC functionality is contained within client_rpc.ts and called from the other scripts. 
- Added dumb_server.ts as an example of a super simple rpc implementation to a simple server. 
- Cleaned out registry.ts so that only specific code relating the the implementatio of that RPC is present.
- Added approprite call in registryControl.scala to execute the call. 

The server being called only exists on my machine. I don't know if I should add that somewhere so we have a complete working example.


DanielSank:
Since ur a n00b I am going to pick on you a bit :-)

In general, it's nice to submit pull requests for "one thing" at a time. Seems like the first two items could be their own things. Not a big deal this time but it's nicer for reviewers if PR's are small.

.idea/sbt.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ScalaSbtSettings">
+    <option name="customLauncherEnabled" value="true" />
+    <option name="customLauncherPath" value="$USER_HOME$/bin/sbt-launch.jar" />
+    <option name="customVMPath" value="$PROJECT_DIR$/../../../../../buildtools/java/jdk8-google-v7/bin" />
+  </component>
+</project>
```
DanielSank: Newline at eof.

DanielSank:
What is .idea/sbt.xml? Can we have a comment line or something saying what this is?

client-js/app/scripts/client_rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+      var message = {
+        jsonrpc: "2.0",
+        method: method,
+        params: params
+      };
+      this.socket.send(JSON.stringify(message));
+    });
+  }
+
+}
```
DanielSank: newline

client-js/app/scripts/dumb_server.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    foo(params:{}): Promise<string>{
+        return client_rpc.sendRequest('foo',params,this.socket,"dumb_server");    
+    }
+    dumb_echo(params:{inp: string}): Promise<string>{
+        return client_rpc.sendRequest('dumb_echo',params,this.socket,"dumb_server")    
+    }
+      
+}
```
DanielSank: newline

scalabrad-web.iml
```diff
@@ -0,0 +1,11 @@
...
+<module type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="false">
+    <output url="file://$MODULE_DIR$/bin" />
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" name="scala-sdk-2.11.7" level="application" />
+  </component>
+</module>
```
DanielSank: newline. You can probably tell your editor to do this for you automatically.

DanielSank:
What's supposed to be in `client-js/app/scripts/app.ts`? Why does that file exist?

joshm-3b6b58:
Thanks for the feedback.

1) .idea is something that intellij added on it's own, I think? I didn't mean to commit it, can I just pluck files from the commit individually?
2) in dumb_server and client_rpc, where exactly do I need the newlines? I can't ascertain this from the comment.
3) scalabrad-web.iml I never purposely `git add`-ed this, so i don't know how it got there. I should remove this file from the pull request.
4) app.ts is basically the logic for the web app as a whole. The idea right now is that a common web-app will handle registry browsing, data-vault browsing, etc, where those functionalities are sub components. app.ts will handle the commong backend between the lot of them. It's basically the logic for index.html. Right now it mostly handles the websocket connection for the entire app and the overall layout which encapsulates the components.

client-js/app/scripts/app.ts
```diff
@@ -4,6 +4,8 @@ import page = require("page");
 
 import registry = require("./registry");
 import datavault = require("./datavault");
+import dumb_server = require("./dumb_server");
+import client_rpc = require("./client_rpc");
```
maffoo: The convention in js is to use `camelCase` for variable names. We should stick with that. Files named with underscores are ok, but I think the convention in the js world is to use `hyphen-separated-names` for files.

DanielSank:
Josh, come get me and I can show you the true path to git-fu.

client-js/app/scripts/client_rpc.ts
```diff
@@ -0,0 +1,94 @@
+import promise = require('es6-promise');
+
+var Promise = promise.Promise;
+
+export function sendRequest(method: string, params: Object, socket, server: string): Promise<string> {
+        console.log("sending dumb RPC named: "+ method+" argument: "+ JSON.stringify(params));
+        return socket.call("org.labrad."+server+"." + method, params)
+                      .then((result) => <string>result);
+}
```
maffoo: indentation should be 2 spaces. This looks like tabs? Also, I think this function should be part of the `JsonRpcSocket` class, and the class can have a member called `prefix` that gets added to the front of the method name for every request that is sent. That way even "org.labrad" does not need to be hardcoded and we keep things maximally flexible, but still encapsulated within JsonRpcSocket.

client-js/app/scripts/dumb_server.ts
```diff
@@ -0,0 +1,21 @@
+
+import promise = require('es6-promise');
+
+import client_rpc = require("./client_rpc");
+
+export class DumbServiceJsonRpc {
+    socket: client_rpc.JsonRpcSocket;
```
maffoo: indentation should be 2 spaces.

client-js/app/scripts/dumb_server.ts
```diff
@@ -0,0 +1,21 @@
...
+    
+    constructor(socket: client_rpc.JsonRpcSocket){
+        this.socket = socket;  
+        console.log("Dumb RPC starting");
+    }
+
+    foo(params:{}): Promise<string>{
+        return client_rpc.sendRequest('foo',params,this.socket,"dumb_server");    
+    }
+    dumb_echo(params:{inp: string}): Promise<string>{
```
maffoo: should use `camelCase`

client-js/app/scripts/client_rpc.ts
```diff
@@ -0,0 +1,94 @@
+import promise = require('es6-promise');
```
maffoo: I would suggest just calling this file `rpc.ts`, instead of `client_rpc.ts`.

client-js/app/scripts/registry.ts
```diff
@@ -9,182 +11,10 @@ export interface RegistryListing {
   vals: Array<string>;
 }
 
-export interface RegistryApi {
```
maffoo: why did all this registry api stuff get deleted?

server/src/main/scala/org/labrad/browser/RegistryController.scala
```diff
@@ -387,6 +393,11 @@ class RegistryApi @Inject() (cxnHolder: LabradConnectionHolder) {
   }
 
   // callable RPCs
+  
+  @Call("org.labrad.dumb_server.dumb_echo")
+  def dumb_echo(inp: String): Future[String] = {
```
maffoo: use `camelCase` for scala names. The string names for json rpc or labrad settings are fine with underscores.

maffoo:
I would rather not include intellij project files like the `*.iml` and `.idea/` in the repo, but if we do want them we should add them in a separate change and check the intellij docs to see which files are ok to include because they are useful for other people just checking out a project, and which should be excluded because they are machine-specific. The latter should be added to `.gitignore`.

maffoo:
Another thought; since no one else has your labrad `dumb_server`, that rpc call is going to be broken for anyone else who tries check out and run this code. Maybe make the example rpc call a setting on the manager instead, such as its own echo setting.

client-js/app/scripts/registry.ts
joshm-3b6b58: I thought all of that was being unused and now exists in `registry.ts` in the function `RegistryServiceJsonRpc`

joshm-3b6b58:
I removed all the intellij stuff in the rebase, I think intellig `git add`-ed this on it's won

client-js/app/scripts/registry.ts
maffoo: The `RegistryApi` just defined the abstract interface, while `RegistryServiceJsonRpc` is an implementation of that interface. I'd like to keep that separation of interface and implementation.

DanielSank:
`git add` only adds what you tell it to add. If you do `git add -A` it adds everything.

client-js/app/scripts/registry.ts
joshm-3b6b58: I don't really understand the distinction. And both seem to contain the same information, the command, the type to send and the expected type to return. Having what looked like the same information in two places in the same file, made the code harder for me to read (since I guess I didn't understand I was deleting something that had a purpose)

maffoo: It's the same distinction as interface and class in java; an interface is just a collection of methods, but you can't actually create instances of an interface. A class is something you can create instances of, and it actually implements those methods. Previously, we had `RegistryService` and `RegistryServiceJsonRpc`, which both implemented the `RegistryApi` interface, so if some piece of code required a `RegistryApi` to work, you could pass an instance of either of those classes and it would be perfectly happy because the both satisfy the interface that it depends on. In a case like that where there are multiple implementations, it's clear that you want to separate the interface from the implementation, but even if we only have one implementation (the json rpc version), it's nice to have this separation. For example, if we want to test some code that needs a `RegistryApi`, we can write a little test class that implements this interface but doesn't actually make remote calls to a server and the code under test will be none the wiser.

joshm-3b6b58: OK, thanks for explaining that. What I don't understand is if the interface is defined somewhere, why do the methods have to be defined again within the class, it looks to me like the same information is being written twice.

maffoo: the methods just have to implemented in the class, but yes, it does mean repeating the type signature of the method, which here looks like a lot because the method bodies themselves are so short.

joshm-3b6b58: Oh I think I get it now, becuase hte actual sendRequest is specifc to the websocket jsonRPC

joshm-3b6b58: Yup, I didn't understand what was specific to the old http RPC protocol and what was not. I just kept hacking away everthing I could that kept the app still running.

client-js/app/scripts/client_rpc.ts
joshm-3b6b58: done

client-js/app/scripts/dumb_server.ts
joshm-3b6b58: done

client-js/app/scripts/app.ts
joshm-3b6b58: got it.

client-js/app/scripts/client_rpc.ts
joshm-3b6b58: I've added `sendRequest` to `JsonRpcSocket`, I defined the prefix in app.ts when new JsonRpcSocket is defined. Let me know if that makes sense...

client-js/app/scripts/registry.ts
joshm-3b6b58: OK it's back and I'm adding an interface in the echo example

joshm-3b6b58:
~~How do I call the manager echo setting?~~
-nevermind

client-js/app/scripts/rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+  nextId: number;
+  socket: WebSocket;
+  prefix: string;
+
+  openPromise: Promise<void>;
+  // A promise that will be resolved as soon as the socket is connected.
+  // We chain call promises off this promise so that the user can immediately
+  // make calls but we don't actually send any messages to the server until the
+  // socket is ready.
+  sendRequest(method: string, params: Object, socket, server: string): Promise<string> {
```
maffoo: We should change this to not take the final `server` parameter, since it's the caller's responsibility to construct the full method string. (registry service has been updated to reflect the change of responsibility, but this method and the echo service have not)

maffoo: Also, I'd suggest moving this to after the constructor, since that's typically how we order things in classes.

client-js/app/scripts/rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+  console.log("sending dumb RPC named: "+ method+" argument: "+ JSON.stringify(params));
+  return socket.call(this.prefix+server+"." + method, params)
+    .then((result) => <string>result);
+}
```
maffoo: method body needs to be indented two more spaces.

client-js/app/scripts/rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+      var message = {
+        jsonrpc: "2.0",
+        method: method,
+        params: params
+      };
+      this.socket.send(JSON.stringify(message));
+    });
+  }
+
+}
```
maffoo: add newline at end of file

server/src/main/scala/org/labrad/browser/RegistryController.scala
```diff
@@ -76,6 +76,12 @@ class RegistryController @Inject() (cxnHolder: LabradConnectionHolder) extends C
     pkt.cd(absPath(path), create = create)
     pkt
   }
+  
+  private def dumbPacket(path: Seq[String], create: Boolean = false) = {
+    val pkt = cxnHolder.cxn.registry.packet()
+    //pkt.cd(absPath(path), create = create)
+    pkt
+  }
```
maffoo, line 84: it seems this is not used anywhere; can remove

client-js/app/scripts/registry.ts
```diff
@@ -226,7 +73,9 @@ export class RegistryServiceJsonRpc implements RegistryApi {
   }
 
   sendRequest(method: string, params: Object): Promise<RegistryListing> {
+      console.log("sending registry RPC: "+ method+" argument: "+ JSON.stringify(params));
```
maffoo: outdent to match rest of method body

client-js/app/scripts/rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+  prefix: string;
+
+  openPromise: Promise<void>;
+  // A promise that will be resolved as soon as the socket is connected.
+  // We chain call promises off this promise so that the user can immediately
+  // make calls but we don't actually send any messages to the server until the
+  // socket is ready.
+  sendRequest(method: string, params: Object, socket, server: string): Promise<string> {
+  console.log("sending dumb RPC named: "+ method+" argument: "+ JSON.stringify(params));
+  return socket.call(this.prefix+server+"." + method, params)
```
maffoo: `socket.call` should be `this.call`

client-js/app/scripts/echo-server.ts
```diff
@@ -0,0 +1,25 @@
...
+  socket: clientRpc.JsonRpcSocket;
+
+  constructor(socket: clientRpc.JsonRpcSocket){
+    this.socket = socket;  
+    console.log("Echo RPC starting");
+  }
+
+  foo(params:{}): Promise<string>{
+    return this.socket.sendRequest('foo',params,this.socket,"manager");    
+  }
```
maffoo: looks like there's no matching implementation on the server?

client-js/app/scripts/app.ts
```diff
@@ -23,10 +25,11 @@ window.addEventListener('WebComponentsReady', function() {
     page(e.detail.path);
   });
 
-  var socket = new registry.JsonRpcSocket('ws://localhost:9000/api/socket2');
+  var socket = new clientRpc.JsonRpcSocket('ws://localhost:9000/api/socket2');
+  socket.prefix = "org.labrad."
```
maffoo: prefix should be passed as a constructor parameter. it can default to the empty string.

client-js/app/scripts/registry.ts
```diff
@@ -26,165 +28,10 @@ export interface RegistryApi {
...
-export class RegistryService implements RegistryApi {
-  baseUrl: string;
-
-  constructor(baseUrl: string) {
-    this.baseUrl = baseUrl + '/api/registry';
-  }
-
-  dir(params: {path: Array<string>}): Promise<RegistryListing> {
-    return this.sendRequest('/dir', params.path);
-  }
```
maffoo, line 38: with all these method implementations removed, I don't see how this service can work. These methods are not actually implemented in the `RegistryApi` interface, just the signatures are defined.

client-js/app/scripts/app.ts
```diff
@@ -23,10 +25,11 @@ window.addEventListener('WebComponentsReady', function() {
...
   var reg = new registry.RegistryServiceJsonRpc(socket);
+  var echo = new echoServer.EchoServiceJsonRpc(socket);
```
maffoo, line 30: given that these both share the same socket and may not have the same prefix, it seems wrong to make prefix a property of the socket; rather, I think these services should build the full method string and then just pass that to the socket.

maffoo:
mades some fixes; see #9

client-js/app/scripts/registry.ts
joshm-3b6b58, line 38: It definitely works. I wouldn't have commited it otherwise. All the implementations are in `RegistryServiceJsonRpc`

client-js/app/scripts/app.ts
joshm-3b6b58, line 30: When I made the prefix a property of the socket I was thinking it would be cleaner if different sockets were opened to make calls with different prefixes, so all calls made to the manager would have the `org.labrad` prefix. If I were going to make calls to a different service I might as well open a new socket.

client-js/app/scripts/registry.ts
maffoo, line 38: yeah, when I looked at the code these implementations are still there, so I guess it's just something screwy in the way github is displaying the diff

client-js/app/scripts/app.ts
maffoo, line 30: prefix is just for namespacing method names, so it would certainly be possible to serve requests with entirely different prefixes on the same server over the same websocket. It just seems like not a good idea to divide the task of constructing the full method name between this jsonrpc socket and the service implementations which use it.

client-js/app/scripts/registry.ts
joshm-3b6b58, line 38: There were two implementations before, one for the HTTP requests, one for the WebSocket, I just removed the HTTP ones

client-js/app/scripts/app.ts
joshm-3b6b58, line 30: I think I misunderstood what you mean by prefix. I thought you meant `org.labrad` as prefix, but you mean, `org.labrad.registry` etc? If we were sending a request to an address other than `org.labrad` I would say open a new socket, but if you mean the whole things then I agree.

maffoo, line 30: yeah, I wasn't very clear. see #9 for what I was thinking, putting the prefix in a service base class, rather than the socket class. I also made it easier to cast return values to the expected types, which will help because in most situations not all methods have the same return type; the registry api currently does make all methods return the same type, but actually I think that should change, because we're transferring lots of extra data that's not in fact needed.

maffoo:
LGTM

PR #7, opened 2015-07-08
Remove duplicate dependence.

DanielSank:
Based on my understanding of `package.json`, there was no reason for `"node-libs-browser"` to be in dependencies and devDependencies.


maffoo:
LGTM

PR #11, opened 2015-07-15
Remove http api and allow server->client calls/notifications over jsonrpc

maffoo:
This touches a lot of code, but in the end I think it will be a big simplification. We remove all the http interfaces and leave just the jsonrpc interfaces, which will mean less code duplication, especially on the server side. Also fixes how server->client calls and notifications work from the server perspective; still need to do some work to wire this up nicely on the client side, but for now they're logged and least so we can verify that they're coming through as expected.

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -25,7 +25,7 @@ window.addEventListener('WebComponentsReady', function() {
     page(e.detail.path);
   });
 
-  var socket = new rpc.JsonRpcSocket('ws://localhost:9000/api/socket2');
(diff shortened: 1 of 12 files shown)


DanielSank:
I don't think I have much more to say about this PR. Given our situation with respect to number of developers and scala literacy, LGTM.

maffoo:
Thanks for looking this over, @DanielSank. I hope reviews will get easier over time as we improve the quality of our frontend code and after folks have had more experience with scala as well.

DanielSank:
It'll get easier when people are using the code. Nothing like wanting to use a lib to get you to learn a language.

PR #12, opened 2015-07-15
Remove gwt client and java code previously shared with server.

maffoo:
Remove the gwt code and go all-in with client-js. This builds on #11, which removes server dependencies on code in the shared project. For now, until that PR is in and this can be rebased, only the last commit is relevant.


client-js/gulpfile.js
```diff
@@ -148,6 +148,8 @@ gulp.task('images', function () {
 // Copy All Files At The Root Level (app)
 gulp.task('copy', function () {
   var app = gulp.src([
+    '.tmp/**/*.js',
+    '.tmp/**/*.js.map',
```
joshm-3b6b58, line 152: Now I'm paranoid about things hapening to gulp. What's going on here? What is gulp copy supposed to do?

maffoo, line 152: this task copies all the files that are part of the final built app into the `dist/` directory to create a distributable version (e.g. just a bunch of static files). The task that we inherited from the polymer starter kit was just copying scripts directly from `app/`, but we added typescript compilation which produces additional javascript files (and source maps) in the `.tmp/` folder. This just makes sure those also get included in the dist package.

joshm-3b6b58:
So I've switched into this branch and while most of scalabrad-web/client is gone there's still a bunch of stuff in there. Why not remove the whole folder?

maffoo:
I thought I did remove the whole folder; I presume whatever is still left is untracked files. What are you seeing in there?

joshm-3b6b58:
client/target/gwt/labradbrowser has 4 files in it..
client/target has a bunch of stuff. Is this auto generated from the build?

maffoo:
yeah, all the stuff in `client/target/` comes from the build. You can just delete the `client/` directory; these files should already be in `.gitignore`

server/src/main/scala/org/labrad/browser/Controllers.scala
```diff
@@ -1,21 +0,0 @@
...
-    val a = request.body.as[A]
-    f(a).map { b =>
-      val headers = Seq.newBuilder[(String, String)]
-      for (origin <- originOpt) {
-        headers += "Access-Control-Allow-Origin" -> origin
-      }
-      Ok(Json.toJson(b)).withHeaders(headers.result: _*)
-    }
-  }
-}
```
joshm-3b6b58, line 21: So has this just been broken out into separate controllers for each sub-application? When I look in the folder now, there's no Controllers.scala, but theres the WebsocketController and a bunch of other ones.

maffoo, line 21: This was just a little utility for creating actions that handle single http requests, taking a json object as the request body and returning a json object as the response body. Since the api calls now all use real jsonrpc over websockets, this is no longer needed. "Controller" is a play framework thing, and we're now only using it for a few things (serving the index.html page and other static files, and handling the websocket connection request). Most of those other files which implement apis are no longer "controllers" in the sense of the play framework, so those files could be renamed.

joshm-3b6b58:
Branch removes gwt stuff. Current implementation of server works on my machine. LGTM.

PR #14, opened 2015-07-17
Add note about javascript code style to README.

### LINKED ISSUE #13: Do we need to require documentation in the code?
I'm having a little bit of a hard time reviewing #11 because I don't know the intent of the code. Do we need some basic requirements on inline documentation in this project?

maffoo:
In particular, we want to make explicit the requirement that code should include JSDoc comments for documentation.

Fixes #13


DanielSank:
LGTM. Thanks for doing this.

PR #17, opened 2015-07-22
Cosmetic changes

joshm-3b6b58:
Made some cosmetic changes to the client. Hooked up labrad-grapher-dataset and started experimenting with a layout


client-js/app/elements/labrad-grapher-dataset.html
```diff
@@ -18,15 +30,17 @@
...
+        type: String,
+        notify: true
       }
     },
-    ready: function() {
-      this.items = [
-        'A single dataset'
-      ];
+    back: function(){
+      window.location.href='/grapher/'+this.path.toString().replace(/,/g,"/");
```
maffoo: I would prefer to have this location passed in as a property, so the logic for building urls is centralized, not spread out among various elements.

client-js/app/elements/dir-item.html
```diff
@@ -0,0 +1,30 @@
+
+<dom-module id='dir-item'>
+  <template>
+    <paper-icon-item><iron-icon icon="{{dir_icon}}" item-icon></iron-icon><content></content></paper-icon-item>
```
maffoo: can you break this across multiple lines to make the structure clearer?

client-js/app/elements/dir-item.html
```diff
@@ -0,0 +1,30 @@
...
+
+
+  </template>
+
+  <script>
+    Polymer({
+        is:"dir-item",
```
maffoo: indentation should be 2 spaces. also, whitespace after ':'

client-js/app/elements/labrad-grapher-dataset.html
```diff
@@ -5,11 +5,23 @@
...
   </style>
   <template>
-    <ul>
-      <template is="dom-repeat" items="{{items}}">
-        <li><span class="paper-font-body1">{{item}}</span></li>
-      </template>
-    </ul>
+  <paper-button on-click="back" raised class="colorful">BACK</paper-button>
+    <paper-tabs selected="{{selected}}" noink> 
+        <paper-tab>Plot</paper-tab>
```
maffoo: indentations should be two spaces

client-js/app/elements/dir-item.html
```diff
@@ -0,0 +1,30 @@
...
+        properties: {
+            dir_name: {type: String},
+            prop: {type: String},
```
maffoo: maybe call this 'kind' or 'type' instead of 'prop'?

client-js/app/elements/dir-item.html
```diff
@@ -0,0 +1,30 @@
...
+            dir_icon:{
+                type: String,
+                computed: 'assignIcon(prop)'
+            }
+        },
+        assignIcon: function(entry_type){
+            if (entry_type == "folder"){
```
maffoo: I like having a space before the opening curly bracket, both on control structures and function defs. This is also the recommendation in the google javascript style guide: http://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting

client-js/app/elements/labrad-grapher-dataset.html
joshm-3b6b58: I tried, but honestly, I couldn't figure out how it worked.

client-js/app/elements/labrad-registry.html
```diff
@@ -1,30 +1,45 @@
...
-      <template is="dom-repeat" items="{{keys}}">
-        <tr>
-          <td><span class="paper-font-body1">{{item.name}}</span></td>
-          <td><span class="paper-font-body1">{{item.value}}</span></td>
-        </tr>
-      </template>
+          <tr>
+            <td>
+              <a is="app-link" path="{{item.url}}" href="{{item.url}}">
+                <span class="paper-font-body1">{{item.name}}</span>
```
maffoo: would be nice to use icons here too, as in dir-item and the delphi registry editor.

client-js/app/elements/labrad-registry.html
```diff
@@ -1,30 +1,45 @@
...
-          <td></td>
-        </tr>
-      </template>
-      <template is="dom-repeat" items="{{keys}}">
-        <tr>
-          <td><span class="paper-font-body1">{{item.name}}</span></td>
-          <td><span class="paper-font-body1">{{item.value}}</span></td>
-        </tr>
-      </template>
+          <tr>
```
maffoo: indentation is 2 spaces

client-js/app/scripts/echo-server.ts
```diff
@@ -1,3 +1,4 @@
+
```
maffoo: any particular reason for all the extra blanks lines here and in registry.ts and rpc.ts?

joshm-3b6b58: No, this was probably a mistake when merging.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: Do you want this for this pull request or just in general? I can make a reg-item along the lines of dir-item, or add extra cases to dir-item?

client-js/app/elements/labrad-grapher-dataset.html
joshm-3b6b58: Actually scratch that, I figured it out and it would involve a change to app.ts function `loadDataset` but I wanted to talk to you about what data structures we were sending back and forth for the grapher before changing this. I wanted to limit this pull request to cosmetic changes, not touching the underlying logic.

maffoo: well, this is new logic (being able to navigate back to the directory listing), so I'd say let's just fix it in `loadDataset`

joshm-3b6b58: OK

client-js/app/elements/labrad-registry.html
maffoo: It seems like reusing dir-item would make sense

client-js/app/elements/dir-item.html
joshm-3b6b58: Yes, thanks for picking that up this is a vestige from when I was trying to learn polymer.

client-js/app/elements/labrad-grapher-dataset.html
maffoo: Also, by setting the href here it causes a full page reload, which is unnecessary. Instead, we just want to transition to the previous view without reloading the app. For now, can use the same technique as the `app-link` element for this and emit an "app-link-click" event: `this.fire('app-link-click', {path: this.parentPath});` (replacing parentPath by whatever property name you decide on).

joshm-3b6b58: I'm just going to put the paper button in`app-link`  anchor tags so it's exactly like all the other links. That way we can more easily play with the styles and appearance when we're polishing the site.

client-js/app/elements/dir-item.html
joshm-3b6b58: like this? 

```
    <paper-icon-item>
      <iron-icon icon="{{dir_icon}}" item-icon></iron-icon>
      <content></content>
    </paper-icon-item>
```

maffoo: exactly

client-js/app/elements/labrad-grapher-dataset.html
maffoo: sounds good. do buttons inside anchor tags still render nicely?

joshm-3b6b58:
I'm just noticing that the web components you wrote are nested in an extra ``function` than mine are. For example you have 

```
<script>
(function () {
  Polymer({
.
.
.
  });
})();
</script>
```

Where I just have

```
  <script>
    Polymer({
.
.
.
    });
  </script>
```

What's the reason for that? Should I nest my Polymer in another function?

client-js/app/elements/labrad-grapher-dataset.html
joshm-3b6b58: They render how you would expect. The text turns alink of vlink color

maffoo:
That's called an [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression) (Immediately-invoked function expression). It's used generally to avoid polluting the global namespace, since variables declared inside a function are local to that function. But given that these scripts just call Polymer and don't declare any top-level variables, it's not needed. I probably got it from copy-pasting some other component, but we could remove it for consistency.

maffoo:
A couple minor comments, otherwise LGTM

PR #22, opened 2015-07-29
Fix up needlessly obscure code in jsonrpc message parsing.

maffoo:


DanielSank:
LGTM.

Do we have tests for this code?

maffoo:
we do not have tests. TODO.

PR #25, opened 2015-07-31
Remove hardcoded websocket url and allow overriding in dev mode.

maffoo:
When we deploy to production, the url for the websocket connection should be relative to the page (the host and port from which we loaded the html). I dev mode, we insert a script to set a global variable to override this so that the api connection goes to the scala server running on port 9000, instead of the js dev server running on port 3000.

### DIFF
--- a/client-js/app/index.html
+++ b/client-js/app/index.html
@@ -33,6 +33,7 @@
   <link rel="stylesheet" href="/styles/main.css">
   <script src="/bower_components/webcomponentsjs/webcomponents-lite.js"></script>
   <link rel="import" href="/elements/elements.html">
+  <!-- PROD_MODE -->
 </head>
 
 <body unresolved class="fullbleed layout vertical">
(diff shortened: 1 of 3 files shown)


client-js/app/index.html
```diff
@@ -33,6 +33,7 @@
   <link rel="stylesheet" href="/styles/main.css">
   <script src="/bower_components/webcomponentsjs/webcomponents-lite.js"></script>
   <link rel="import" href="/elements/elements.html">
+  <!-- PROD_MODE -->
```
DanielSank, line 36: Not sure to what this comment refers.

maffoo, line 36: It's just a marker so the dev mode code knows where to insert its script tag. I'm happy to change this to something more meaningful. Originally I wrote PROD_MODE thinking that we wouldn't want something like DEV_MODE_SPECIAL_CODE_GOES_HERE in the html when we do serve the file in prod, but in fact the `html` gulp task strips all the comments anyway, so we can put whatever we want.

joshm-3b6b58:
LGTM

client-js/app/index.html
DanielSank, line 36: Would have been nice to include a comment to explain what that comment does.

maffoo, line 36: oh right, sorry, forgot to address that. See #28

PR #28, opened 2015-08-03
Explain spot where dev-mode websocket url gets inserted.

maffoo:


client-js/app/index.html
```diff
@@ -33,7 +33,8 @@
   <link rel="stylesheet" href="/styles/main.css">
   <script src="/bower_components/webcomponentsjs/webcomponents-lite.js"></script>
   <link rel="import" href="/elements/elements.html">
-  <!-- PROD_MODE -->
+
+  <!-- Websocket url for dev mode -->
```
DanielSank: `PROD_MODE` became `Websocket url for dev mode`. Seems suspicious.

client-js/gulpfile.js
```diff
@@ -181,7 +181,7 @@ gulp.task('html', function () {
 // Inject an alternate url for connecting to the api backend in dev mode.
 gulp.task('websocket-url', function () {
   return gulp.src(['app/index.html'])
-    .pipe($.replace('<!-- PROD_MODE -->',
+    .pipe($.replace('<!-- Websocket url for dev mode -->',
```
DanielSank: ~~Again "PROD" turned into "dev". Is this intentional?~~ This is explained in the other comment.

client-js/app/index.html
maffoo: Let's go over once again what this is doing. The comment here appears in index.html, which is served AS IS\* in prod mode. In dev mode, this comment gets _replaced_ by a script which overrides the url where we should connect to the api backend. Originally I wrote PROD_MODE because if you look at the source of this file in PROD_MODE it would then not look out of place, but of course this does not then convey any information about why the comment is there at all.

The change here is meant to have the comment describe why it is there, e.g. to be replaced by the dev mode websocket url when running in dev mode. As it happens, we minify the html before using it in prod mode, so comments get stripped (ie, that AS IS above was not quite correct). Hence it won't happen that someone looking at the source in prod mode will see the comment either way, so my original reasoning for writing PROD_MODE is kind of moot.

In the end, though, this is just a placeholder token. We can write here _whatever we want_. If you tell me what you think would be least confusing to a new reader of the code, I'd be happy to use that. It just has to match between here and the spot in the gulpfile where we do the replacement.

DanielSank: This makes perfect sense.

DanielSank:
LGTM

DanielSank:
The one thing I might do (not rquired) is add a comment just above the existing one explaining that the existing one is a place holder to be filled in by the dev mode build.

maffoo:
@DanielSank, I tried to make this more clear with an additional comment, also more general in case there is ever any other configuration we need to inject besides just the websocket url. Could you take another quick look?

DanielSank:
That is very clear. Thanks.

φαίνεται καλό για μένα

DanielSank:
I picked on this because, as we mentioned the other day, I find the build systems and tooling to be the hardest part of any software project, so I think it's worth being really clear there. Anyone can decipher a tricky line of code but build tooling is so loosely coupled that it can be really hard to figure out what's going on.

maffoo:
Looks Greek To Me? :) Yeah, I agree it can be tricky to figure out the build tooling. I'm glad you're looking at things and trying to make sense of them; that definitely helps make things better.

DanielSank:
> Looks Greek To Me? :)

Heh, yeah. I'm getting a little silly today and some times it just takes another language to satisfy that. I was considering writing you a Russian review but got annoyed trying to install a Russian keyboard layout, and doing it all through Google translate is cheating.

> Yeah, I agree it can be tricky to figure out the build tooling. I'm glad you're looking at things and trying to make sense of them; that definitely helps make things better.

It's symbiotic because I get to learn with somewhat less effort than if were to do it all from scratch :)

maffoo:
> It's symbiotic because I get to learn with somewhat less effort than if were to do it all from scratch :)

I just feel bad because it's sort of hacked together and will (hopefully) change a lot as we improve it; in other words, it's not really the best code to learn from. But hey, I guess working code is better than no code.

DanielSank:
> I just feel bad because it's sort of hacked together and will (hopefully) change a lot as we improve it

Kinda like... the web?

> in other words, it's not really the best code to learn from.

Dude, if you can honestly point to a single web application from which you think I _should_ learn, I'll not only receive the reference happily, I'll eat my headphones.

maffoo:
> Dude, if you can honestly point to a single web application from which you think I should learn, I'll not only receive the reference happily, I'll eat my headphones.

Challenge accepted :)

DanielSank:
If you mean that I'm giddy with excitement. I've found a lot of crap out there and despite my relative inexperience I've already developed a deep distrust of other "web programmers". For example, the moment I read "publish-subscribe" my brain more or less turns off.

PR #30, opened 2015-08-07
U/joshmutus/registry features alt

joshm-3b6b58:
This adds the web interface for the registry editor.

Lot of new code here, but it's kind of the bare minimum to get a working registry interface. @maffoo and I talked about features to add next which should be filed as issues after this PR is sorted.


DanielSank:
Can we separate the html and js in `labrad-registry.html`? Also, some doc strings there would seem appropriate.

client-js/app/elements/dir-item.html
```diff
@@ -16,6 +16,9 @@
         dirIcon: {
           type: String,
           computed: 'assignIcon(kind)'
+        },
+        link: {
```
DanielSank, line 20: Can this go on one line?

joshm-3b6b58, line 20: It could. But the examples I've seen all separate the code like this. I don't know if it's according to a style guide or just to make the examples easier to read.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
+          <tbody is="selectable-table" is="keyList" attr-for-selected="name"  selected="{{selKey}}" on-tap="selectKey"> 
+            <template is="dom-repeat" items="{{keys}}">
+              <tr name="{{item.name}}">
+                <td>
+                  <iron-icon icon = "more-vert"></iron-icon>
+                </td>
+                <td>
+                  <dir-item kind="key">{{item.name}}</dir-item>
+                </td>
+                <td>
```
maffoo: need to style this third column so it expands to maximum width to show the data.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
+    <section onclick="clickHandler(event)">
+      <menu-button data-dialog="doDelete" name="delete" icon-name="delete" class = "done"></menu-button>
+      <menu-button data-dialog="newFolderDialog" name="new folder" icon-name="folder-open" class = "done"></menu-button>
+      <menu-button data-dialog="copyDialog" name="copy" icon-name="content-copy" class = "done"></menu-button>
+      <menu-button data-dialog="renameDialog" name="rename" icon-name="code" class= "done"></menu-button>
+      <menu-button data-dialog="newKeyDialog" name="new key" icon-name="create"  class = "done"></menu-button>
+     </section>
     </div>
+    <div id="reg-area"> 
+      <table>
```
maffoo: need to style this table to extend to full width.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
+                </td>
+                <td>
+                  <dir-item kind="key">{{item.name}}</dir-item>
+                </td>
+                <td>
+                  <form is="iron-form" id="{{item.name}}">
+                      <iron-a11y-keys keys="tab" on-keys-pressed="incrementSelector">
+                       <paper-input-container>
+                        <label>{{item.value}}</label>
+                        <input is="iron-input" name="{{item.name}}">
```
maffoo: the input here doesn't deal with multiline text when input gets large. maybe use [iron-autogrow-textarea](https://github.com/PolymerElements/iron-autogrow-textarea/blob/master/iron-autogrow-textarea.html) instead?

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
+                      </iron-a11y-keys>
+                  </form>
+                </td>
+              </tr>
+            </template>
+          </tbody>
+      </div>
+    </table>
+    <!-- These are the dialogs -->
+    <paper-dialog id="doRename">
```
maffoo: It would be nice if pressing enter in dialogs would work, instead of having to click on ok. Maybe it's possible to make that work by setting some attribute, but it not we could add it manually.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
+      <form>
+        <input></input>
+      </form>
+      <div class="buttons">
+        <paper-button dialog-confirm autofocus>Submit</paper-button>
+      </div>
+    </paper-dialog>
+
+    <paper-dialog id="doDelete">
+      <p>Delete "<span class="nothing">{{selDir}}</span><span class="nothing">{{selKey}}</span>" of type <span>{{selectType}}</span>?</p>
```
maffoo: what does class "nothing" do here? I don't see any styling for that.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
-      </tbody>
-    </table>
+    <div id="reg-menu"> 
+    <section onclick="clickHandler(event)">
+      <menu-button data-dialog="doDelete" name="delete" icon-name="delete" class = "done"></menu-button>
+      <menu-button data-dialog="newFolderDialog" name="new folder" icon-name="folder-open" class = "done"></menu-button>
+      <menu-button data-dialog="copyDialog" name="copy" icon-name="content-copy" class = "done"></menu-button>
+      <menu-button data-dialog="renameDialog" name="rename" icon-name="code" class= "done"></menu-button>
+      <menu-button data-dialog="newKeyDialog" name="new key" icon-name="create"  class = "done"></menu-button>
+     </section>
```
maffoo: It would be nice if these buttons could float over the page so they are still accessible when you scroll down in a directory with a lot of entries.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
+    <div id="reg-area"> 
+      <table>
+          <tbody class="table-header" is="selectable-table"> 
+            <th></th>
+            <th>Key/Dir Name</th>
+            <th>Key Value</th>
+          </tbody>
+          <tbody is="selectable-table"  id="dirList" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">       
+            <template is="dom-repeat" items="{{dirs}}">
+                <tr name="{{item.name}}">
```
maffoo: The indentation levels here are somewhat inconsistent. We should stick to two spaces. You can set this up in eclipse, just be aware that the editor preferences for different languages have to be set individually.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
+                  <iron-icon icon = "more-vert"></iron-icon>
+                </td>
+                <td>
+                  <dir-item kind="key">{{item.name}}</dir-item>
+                </td>
+                <td>
+                  <form is="iron-form" id="{{item.name}}">
+                      <iron-a11y-keys keys="tab" on-keys-pressed="incrementSelector">
+                       <paper-input-container>
+                        <label>{{item.value}}</label>
```
maffoo: inconsistent indentation here, too.

client-js/app/elements/labrad-registry.html
```diff
@@ -59,8 +207,213 @@
...
+      },
+      selectType: {
+        type: String,
+        notify: true
       }
-    }
+    },
+    ready: function() {
+      console.log(Polymer.version);
+      this.addEventListener('iron-form-submit',this.updateKey);
```
maffoo: whitespace between arguments:

``` javascript
this.addEventListener('iron-form-submit', this.updateKey);
```

client-js/app/elements/labrad-registry.html
```diff
@@ -59,8 +207,213 @@
...
+    selectDir: function(){
+      this.selKey = null;
+      this.selectType = 'dir';
+    },
+    incrementSelector: function() {
+      console.log(this.selKey);
+      //TODO increment selected key on tab
+    },
+    repopulateList: function(resp) {
+      this.selDir=null;
```
maffoo: whitespace around '='

client-js/app/elements/labrad-registry.html
```diff
@@ -59,8 +207,213 @@
...
+      this.splice('dirs',0,this.dirs.length);
+      this.splice('keys',0,this.keys.length);
```
maffoo: whitespace between arguments.

client-js/app/elements/labrad-registry.html
```diff
@@ -59,8 +207,213 @@
...
+        //bw = resp.dirs.length-1-i; //run through list backwards so it orders correctly
+        //console.log(resp.dirs[i]);
+        this.push('dirs',{name: resp.dirs[i], url: this.createUrl(resp.path,resp.dirs[i])});
+      }
+      for (var j in resp.keys){ 
+        //bw = resp.keys.length-1-i; //run through list backwards so it orders correctly
+        this.push('keys',{name: resp.keys[j], value: resp.vals[j]});
+      }
+    },
+    createUrl: function(path,dir) {
```
maffoo: As we've discussed, it would be nice to avoid duplicating logic like this between the typescript code and individual polymer elements. We need to figure out how to have one scheme for modularity of the code, not have both typescript module imports and html imports. But that's clearly out of scope for this change.

client-js/app/elements/labrad-registry.html
```diff
@@ -59,8 +207,213 @@
...
+      console.log(error);
+    },
+    /////////////////////
+    //Interface Functions
+    ////////////////////
+    doDelete: function() {
+      var self = this;
+      var p1;
+      
+      if (self.selectType==='dir') {
```
maffoo: whitespace around '==='

client-js/app/elements/labrad-registry.html
```diff
@@ -59,8 +207,213 @@
...
+        p1 = self.socket.rmDir({path: self.path,dir: self.selDir});
+        p1.then(function(resp) {
+          self.repopulateList(resp);
+        }, 
+        function(reason) {
+          self.handleError(reason); 
+        });
```
maffoo, line 299: the indentation here is a bit misleading, making it unclear that this second function is the error callback. I would suggest something like:

``` javascript
p1.then(
  function(resp) { self.repopulateList(resp); },
  function(reason) { self.handleError(reason); }
);
```

client-js/app/elements/labrad-registry.html
```diff
@@ -59,8 +207,213 @@
...
+      console.log(event.target[0]);
+      event.target.reset();
+      event.target[0].value = null; 
+      //todo fix rendering of input after repopulatelist
+    },
   });
 })();
+
+function clickHandler(e) {
+  //lint complains about this but it's used. This is what makes the dialogs popup
```
maffoo: You should put this inside the IIFE above. Putting it here makes it a global function (e.g. defined on the global window object for the entire page); putting it in the IIFE keeps it scoped to this element only.

client-js/app/elements/labrad-registry.html
```diff
@@ -6,40 +6,188 @@
...
-        <template is="dom-repeat" items="{{keys}}">
-          <tr>
-            <td><dir-item kind="key">{{item.name}}</dir-item></td>
-            <td><span class="paper-font-body1">{{item.value}}</span></td>
-          </tr>
-        </template>
-      </tbody>
-    </table>
+    <div id="reg-menu"> 
+    <section onclick="clickHandler(event)">
```
maffoo: this references `clickHandler` as a global function. Might be better to make click handler just be defined locally (as suggested below) and wire up this click handler in the element's `ready` callback.

client-js/app/elements/selectable-table.html
```diff
@@ -0,0 +1,22 @@
...
+  <script>
+    Polymer({
+      is: 'selectable-table',
+      extends: 'tbody',
+      behaviors: [
+        Polymer.IronSelectableBehavior
+      ]
+    });
+  </script>
+</dom-module>
```
maffoo, line 22: newline at end of file

client-js/app/scripts/app.ts
```diff
@@ -94,6 +94,10 @@ window.addEventListener('WebComponentsReady', function() {
       app.path = path;
       app.registryDirs = dirs;
       app.registryKeys = keys;
+      app.reg = reg;
+      console.log(dirs);
+      app.test = {name:"josh", url:"HWENG"};
```
maffoo: still needed?

maffoo:
It'd be good to clean up the code style issues (indentation, whitespace, global functions, etc) before merging. Changes to element visual styling need not block this change however, so after those code issues are addressed, LGTM.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: Vestige from debugging. I was having trouble formatting this, but fixed it and forgot to take the class away.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: I'll make sure to add this before merging the PR. There's presumably some toolbar thingy I can just put this into.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: Yeah, I have the 2 spaces set up, the extra spaces are left over from playing the adding extra divs, et cetera, to debug the layout. Just missed fixing the indentation here and a few other places.

client-js/app/elements/labrad-registry.html
joshm-3b6b58, line 299: Yeah, that looks much clearer.

client-js/app/scripts/app.ts
joshm-3b6b58: Nope, good find.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: @maffoo can you take a look at https://github.com/labrad/scalabrad-web/commit/aa33d66e8d666b81898d643761471ea7db95c021 it's got a floating toolbar, but theres a weird bug with the icons showing up on top of it.

maffoo: @joshmutus, I'm guessing you could fix this by adjusting the z-index in css. Unfortunately, I can't get the revision you mention to build on my machine. I think the problem is related to webpack and typescript versions, because I've seen a similar problem on linux that prevents the current master from building on gubuntu if you do a fresh clone. I've made some changes to the build system in #31 that fix all this, so maybe we can hold off on this particular styling issue until you've merged the rest of your change. Then we can get #31 in and go from there.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: I can't immediately figure this out. Might take a little exploring or we can create a custom component with a11y keys to bind enter and escape. I'll leave this out of this PR if that's OK since we have a more major keyboard FR to handle.

maffoo: no problem to leave that out of this PR

client-js/app/elements/labrad-registry.html
joshm-3b6b58: I'm having trouble getting this to work. It doesn't immediately work if I just put it inside the 

```
function() {
  function clickHandler(e) {
...
```

before (or after) the `Polymer` definition

client-js/app/elements/labrad-registry.html
joshm-3b6b58: I'm a bit confused about how to accomplish this. I still don't have a firm grasp of scope in JS :/

maffoo: it looks like if you use `on-click` (a special polymer attribute) instead of `onclick` (a standard DOM attribute) then polymer will wire this up to call an instance method on the element itself, so you could avoid having `clickHandler` be a global function. (https://www.polymer-project.org/1.0/docs/devguide/events.html#annotated-listeners)

joshm-3b6b58:
So, I made all the changes except I couldn't get the `clickHandler` in the correct place. Can I get a hand with this quickly tomorrow?

maffoo:
sure thing

client-js/app/elements/labrad-registry.html
joshm-3b6b58: Ahhhhhhhhhh ok. I thought I was going crazy.

PR #35, opened 2015-08-14
Add UI for controlling node servers

maffoo:
I've split this into two pieces. In the first commit there's some infrastructure, a class called `Notifiable` that the various api classes create for the messages that they expect from the server which allows adding and removing message handlers for that message type. @DanielSank and I discussed the design of this, but I'd like to get some feedback now that I have a first cut working.

The second commit adds the node ui, meant to replace the old GWT-based version we currently use. To test this out, you'll probably want to spin up a few node servers. In a new shell you can pop into your python virtualenv and then run:

```
twistd --pidfile=foo.pid -n labradnode --name=foo
```

This will start up a node called "node foo" and then you can start a few others with different names. You have to specify a different pidfile name for each copy of the node you run, or else run them in separate directories because the default pid file name will collide. (Note that you're supposed to be able to run the node source file as a script, but the command line parsing is broken; see: https://github.com/labrad/pylabrad/pull/144)

You'll see that the node control UI updates automatically when node servers connect or disconnect, and also when you configure their search paths (you'll want to add your labrad servers directory to the `directories` key in the registry at `["", "Nodes", "<name>"]`.

### DIFF
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -34,6 +34,7 @@
 <link rel="import" href="labrad-grapher.html">
 <link rel="import" href="labrad-grapher-dataset.html">
 <link rel="import" href="labrad-manager.html">
+<link rel="import" href="labrad-nodes.html">
 <link rel="import" href="labrad-registry.html">
 <link rel="import" href="dir-item.html">
 <link rel="import" href="labrad-server.html">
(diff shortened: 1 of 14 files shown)


DanielSank:
I think `Notifiable` should not live in the rpc file. I would break this out and write it as an entirely independent interface for registering callbacks.

maffoo:
That's a good point. I'll move it out and rename to make the observable stuff more generic.

DanielSank:
Let's get into the habit of documenting our interfaces. A lay user may not understand how `Observable` should be used.

By the way, seeing the simplicity of the observable code here made me realize that all of the complexity in [observed](https://github.com/DanielSank/observed) comes from trying to make it work as a decorator on existing functions/methods, and in dealing with weak refs.

DanielSank:
@maffoo this all looks pretty good. I would go for a more explicit documentation format with all of the arguments and class attributes listed, etc. but that's an issue which shouldn't hold this back.

LGTM.

maffoo:
Thanks @DanielSank for the careful, thoughtful review

PR #38, opened 2015-09-01
Clean up registry dialogs

maffoo:
Use paper elements to give the registry editor dialogs a common style and fix up their behavior. Also pull a few styles out of main.css and into elements themselves.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -1,3 +1,4 @@
+<link rel="import" href="../bower_components/paper-input/paper-textarea.html">
 <link rel="import" href="app-link.html">
 <link rel="import" href="dir-item.html">
 
@@ -21,6 +22,7 @@
       background-color: white;
       display: inline-block;
       @apply(--shadow-elevation-2dp);
+      height: 100%;
     }
     .iron-selected {
       background-color: #ddd;
@@ -36,6 +38,9 @@
       display: table; 
       border: 1px solid;
     }
+    table {
+      width: 100%;
+    }
     tbody {
       display:table-header-group;
     }
@@ -49,7 +54,7 @@
       display:table-header-group;
 
     }
-    .label-wide{
+    .label-wide {
       width: 100%;
     }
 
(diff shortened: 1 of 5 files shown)


joshm-3b6b58:
The menu buttons don't work for me anymore. For example, when I click "new folder" the dialog doesn't pop up.
`[labrad-registry::_createEventHandler]: listener method `newFolderClicked` not defined`

joshm-3b6b58:
I get these scrollbars on the registry entry table now, that I don't think do anything:
screenshot from 2015-09-01 16 34 26

maffoo:
`newFolderClicked` definitely exists; did you clean and rebuild everything? As for the scrollbars; I'll see if I can fix that; which browser are you on?

maffoo:
oops; I had a syntax error in there that was breaking the typescript compiler. Should work now.

joshm-3b6b58:
Scrollbars are gone and the buttons work. I really like what you did here.

client-js/app/elements/labrad-registry.html
```diff
@@ -102,87 +102,74 @@
...
 
-    <!-- These are the dialogs -->
-    <paper-dialog id="doRename">
-      <p>Rename "<span>{{selItem}}</span>" to something else</p>
+    <paper-dialog id="newKeyDialog" modal>
+      <h1>Create new key</h1>
       <form>
-        <input></input>
+        <paper-input id="newKeyInput" label="Key"></paper-input>
+        <paper-textarea id="newValueInput" label="Value"></paper-textarea>
```
joshm-3b6b58, line 126: Textarea is a nice multiline input but if I try and input something with newlines in between input, it fails in some unexpected way. As far as I can tell it never even sends the RPC command. 

Sample input:

```
"123123123

123123123"
```

client-js/app/elements/labrad-registry.ts
```diff
@@ -91,43 +91,129 @@ export class LabradRegistry extends polymer.Base {
...
 
-  /////////////////////
-  //Interface Functions
-  ////////////////////
-  doDelete() {
-    var self = this;
+  onKeyPress(event) {
+    console.log('onKeyPress', event);
+    switch (event.key) {
+      case 'j':
```
joshm-3b6b58: What's going on here? Are these hotkeys?

maffoo: oops, that was an experiment; removed for now.

client-js/app/elements/labrad-registry.html
maffoo, line 126: The error message is not great, but there is an error in the console. The problem is that's not valid input; you have to escape newlines in string literals: `"123123123\n123123123"`. There's no equivalent to python's triple-quoted strings, though we could add that. Newlines do work fine between, say, list elements:

```
[1, 2, 3,
 4, 5, 6
]
```

joshm-3b6b58, line 126: If I remember correctly the old delphi registry editor would basically supress or revert newlines it didn't like. Is this behavior we want to add here eventually?

joshm-3b6b58:
LGTM

PR #42, opened 2015-09-08
Add searchbar for filtering registry entries

joshm-3b6b58:
For long lists of registry keys and directories, you can now type a substring of the key/dir you want into the search bar and the key/dir lists will be filtered based on that substring.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -73,6 +73,18 @@
     iron-autogrow-textarea {
       max-width: 400px;
     }
+    #search-bar {
+      padding: 0px 10px;
+    }
+    #search-bar input {
+      font-family: var(--primary-font-family);
+      font-size: 15px;
+      width: 50%;
+      padding: 10px;
+      border: 0;
+      border-radius: 2px;
+      -webkit-appearance: none;
+    }
   </style>
   <template>
     <div id="reg-menu">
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-registry.ts
```diff
@@ -26,14 +26,31 @@ export class LabradRegistry extends polymer.Base {
...
+  reloadMenu() {
+    //triggers re-render of dir, key lists when filterText is changed
+    this.$.dirList.render();
+    this.$.keyList.render();
+  }
+
+  filterFunc(item) {
+    // called when dir, key lists are populated. Returns entries that contain
+    // substring in filterText
+    return item.name.match(new RegExp(this.filterText, 'i'));
```
maffoo: It would be nice to avoid recompiling this regular expression for every item. I would suggest creating it once in the `reloadMenu` method and storing it in a member variable.

client-js/app/elements/labrad-registry.html
```diff
@@ -98,8 +99,8 @@
             </tr>
           </template>
         </tbody>
-        <tbody is="selectable-table" is="keyList" attr-for-selected="name"  selected="{{selKey}}" on-tap="selectKey">
-          <template is="dom-repeat" items="{{keys}}">
+        <tbody is="selectable-table"s attr-for-selected="name"  selected="{{selKey}}" on-tap="selectKey">
```
maffoo: typo: extra s after `is` attribute

maffoo:
a couple of small comments. otherwise LGTM

joshm-3b6b58:
So it's cool if I merge this?

joshm-3b6b58:
Otherwise LsGTY?

maffoo:
I think it would be nice to stick with a polymer input element to maintain a consistent style, e.g.: https://github.com/labrad/scalabrad-web/commit/faa33557ac69c9cc04e11b2ab37026cffc3c2446

joshm-3b6b58:
The search bar is a different kind of input that a registry key, I was trying to differentiate it from that input. Paper-input is more styled to be a form input which is kind of different than this application. Also this is more consistent with the search inputs on other websites like Inbox, GitHub, Polymer, Scala's tutorial site, etc..

maffoo:
Well, I personally like the polymer style better than the blue outline when clicking on the search box, but seeing as how we are totally in the realm of style here, I don't think it's worth pressing the point :) LGTM

joshm-3b6b58:
Blue outline? I have a grey outline. I can suppresse the outline entirely if you like. 

Anyhow, you're right we can tune the CSS when we have the app more fleshed out so things are consistent among pages. My goal here was to emulate the search bar that appears on other sites.

PR #44, opened 2015-09-17
Create Expandable dialog

joshm-3b6b58:
This is a branch off a previous commit to finish up the expandable dialog using javascript instead of TypeScript. I couldn't figure out how to resolve the detached head, so I created a new branch :/ 

Added a Polymer Tooltip component to hint at keyboard shortcuts, for this to work correctly you have to install by running
`$ bower install --save PolymerElements/paper-tooltip`
from /client-js

### DIFF
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -1,11 +1,13 @@
 <!-- Paper, Iron and Platinum elements -->
 <link rel="import" href="../bower_components/iron-a11y-keys/iron-a11y-keys.html">
 <link rel="import" href="../bower_components/iron-flex-layout/classes/iron-flex-layout.html">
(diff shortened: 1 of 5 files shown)


maffoo:
The tooltips are a nice touch!

LGTM

maffoo:
One other thing; be sure to include the modifications to bower.json since you're adding another element.

joshm-3b6b58:
OK, added bower.json. I just noticed in the scrum of reverting to a previous version, I forgot to include the variable `cols` that is the hard coded (I know, I know, I'll make myself an issue to dynamically size this) length before shortening the value to a single line terminating in `...`

Everything still look good?

maffoo:
LGTM

Issue #33, opened 2015-08-13
Ctrl-click on intra-app links does not work properly


maffoo:
For app-links, Ctrl-clicking just follows the link, whereas it should open the link in a new tab and leave do nothing in the current tab (just like conventional links). The links in the menu tray also do not handle Ctrl-click properly; in that case the link is opened in a new tab as expected, but the current tab also gets messed up (it seems to update the ui as though it were making a screen transition, but does not actually update the page url).

joshm-3b6b58:
So, in looking at existing [Polymer projects](https://github.com/Polymer/polymer-element-catalog/blob/master/app/elements/app-shell/app-shell.html) from Google, the way that they handle this is to use the [app-router](https://github.com/erikringsmuth/app-router) web component. 

I don't know if this also is related to the life-cycle of the components conversation I had with you (briefly) on Friday, so I don't know if adding this is a good idea or if it would conflict with the refactor already in progress.

joshm-3b6b58:
Nevermind, can simply be fixed by stealing this code: 

https://github.com/Polymer/polymer-element-catalog/blob/master/app/elements/app-link/app-link.html

maffoo:
Funny that we picked the same name `app-link` for this component. Thanks for the tip about the `app-router` component. That's certainly something we can try out.

joshm-3b6b58:
This should probably be re-opened since the lifecycle issue isn't fixed.

joshm-3b6b58:
I'm closing this since https://github.com/labrad/scalabrad-web/pull/49 fixes this

PR #45, opened 2015-09-19
enable ctrl-click for opening links in new tabs

joshm-3b6b58:
Fixes https://github.com/labrad/scalabrad-web/issues/33

### DIFF
--- a/client-js/app/elements/app-link.html
+++ b/client-js/app/elements/app-link.html
@@ -2,14 +2,17 @@
 (function () {
   Polymer({
     is: 'app-link',
-    
+
     extends: 'a',
-    
+
     listeners: {
       'click': 'onClick'
     },
 
     onClick: function(e) {
+      if (e.ctrlKey || e.metaKey) {
+        return true;
+      }
       this.fire('app-link-click', {path: this.path});
       e.preventDefault();
     }


joshm-3b6b58:
To clarify, this actually doesn't fix all of #33, just the part about ctrl-click not working. The sidebar links are still wonky, but I suspect this has something to do with the component lifecycle and may be fixed by other things in progress.

client-js/app/elements/app-link.html
```diff
@@ -10,7 +10,11 @@
     },
 
     onClick: function(e) {
-      this.fire('app-link-click', {path: this.path});
+      if (e.ctrlKey || e.metaKey) {
+        return true;
+      } else { 
+        this.fire('app-link-click', {path: this.path});
+      }
```
maffoo: I would suggest just leaving out the else clause since the if clause returns. This makes it clear that preventDefault goes together with the custom event, in response to clicks with no special keys.

``` javascript
if (e.ctrlKey || e.metaKey) {
  return true;
}
this.fire('app-link-click', {path: this.path});
e.preventDefault();
```

maffoo:
one code style comment. Otherwise LGTM.

PR #49, opened 2015-09-22
Create page elements explicitly with typescript static methods.

maffoo:
This has the advantage that we control when elements are created (so that elements do not share state between page transitions), and we can statically type the properties being passed to the static initialization functions.

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -10,6 +10,7 @@ import * as rpc from "./rpc";
 
 import {LabradNodes, LabradInstanceController, LabradNodeController} from "../elements/labrad-nodes";
 import {LabradRegistry} from "../elements/labrad-registry";
+import * as pages from "../elements/labrad-pages";
 
 // autobinding template which is the main ui container
 var app: any = document.querySelector('#app');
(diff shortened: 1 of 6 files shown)


client-js/app/scripts/app.ts
```diff
@@ -141,20 +153,16 @@ window.addEventListener('WebComponentsReady', function() {
...
-  function loadDataset(path: Array<string>, dataset: String) {
+  function loadDataset(path: Array<string>, dataset: string) {
     console.log('loading dataset:', path, dataset);
     dv.dir(path).then((listing) => {
       app.route = 'dataset';
-      app.path = path;
-      app.dataset = dataset;
-      app.parentUrl = '/grapher/' + pathStr(path);
+      var parentUrl = '/grapher/' + pathStr(path);
+      setContent(pages.DatasetPage.init(path, dataset, parentUrl);
```
joshm-3b6b58: Missing parenthesis

maffoo: fixed.

joshm-3b6b58:
The page renders differently than before, which is a bit concerning because I didn't see you changing any CSS:
This PR:
screenshot from 2015-09-22 15 22 43
Master:
screenshot from 2015-09-22 15 25 04

I'm fine with the aesthetic change if you meant to make it, but I'm concerned if there's something a wrong if the layout is changing unintentionally.

maffoo:
The change was unintentional, but my ignoring style issues for now was intentional. The problem is that the paper-material elements were being styled by rules in app-theme.html and when I moved those elements into their own components to encapsulate the various pages, those rules are no longer getting applied, presumably because the paper-material element is now hidden inside the shadow DOM of the page element. I guess the autobinding template we were using before in index.html works differently than other custom elements and does not introduce a shadow DOM.

Polymer used to have a method to reach inside the shadow DOM and apply styles, but those mechanisms (the `::shadow` and `/deep/` css selectors) have been deprecated, and this area is currently in flux for web components generally (see [here](https://www.polymer-project.org/1.0/docs/devguide/styling.html#background) for some discussion).

joshm-3b6b58:
I see. LGTM. 

I like this, the app is much more well behaved now.

PR #50, opened 2015-09-22
fix menu-bar rendering issue

### LINKED ISSUE #34: Floating buttons on registry ui cover the navigation menu
The floating toolbar buttons on the registry editor ui cover the navigation menu which slides in from the left; this makes it hard to navigate from the registry editor to other parts of the site.

joshm-3b6b58:
Fixes #34 

Basically re-ordering the divs removes the need to use z-index. avoiding z-index plays nice.
The div that is placed last, shows up on top of the pile rendering wise.

The lesson apparently: never use z-index.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -8,14 +8,17 @@
     :host {
       display: block;
     }
+
     #reg-menu {
       position: fixed;
-       @apply(--shadow-elevation-3dp);
-       padding-top: 5px;
-       padding-bottom: 5px;
-       background-color: white;
-       z-index: +1;
+      top: 140px;
+      left: 33px;
+      @apply(--shadow-elevation-2dp);
+      padding-top: 5px;
+      padding-bottom: 5px;
+      background-color: white;
     }
+
     #reg-area {
       margin-top: 85px;
       padding-top: 12px;
@@ -24,6 +27,7 @@
       @apply(--shadow-elevation-2dp);
       height: 100%;
     }
+
     .iron-selected {
       background-color: #ddd;
     }
@@ -33,11 +37,7 @@
       border: 1px solid #ccc;
       border-top: none;
     }
-    .reg-list { 
-      padding-top: 12px;
-      display: table; 
-      border: 1px solid;
-    }
+    
     table {
       width: 100%;
     }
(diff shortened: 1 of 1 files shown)


client-js/app/elements/labrad-registry.html
```diff
@@ -8,15 +8,24 @@
...
     }
-    #reg-menu {
+
+    .z-container {
+      background: white;
+      height: 100%;
+      margin: 0px;
+    }
+
+    .reg-menu {
```
maffoo: why using classes here instead of ids?

client-js/app/elements/labrad-registry.html
```diff
@@ -8,15 +8,24 @@
...
       position: fixed;
-       @apply(--shadow-elevation-3dp);
-       padding-top: 5px;
-       padding-bottom: 5px;
-       background-color: white;
-       z-index: +1;
+      top: 140px;
+      left: 33px;
```
maffoo, line 15: this is unfortunate; I mean, it works for now, but magic numbers like this are a pain to tweak, e.g. if the toolbar changes  or fonts or any number of other things.

client-js/app/elements/labrad-registry.html
```diff
@@ -87,57 +97,53 @@
...
-        <menu-button on-click="copyClicked" name="copy" icon-name="content-copy" disabled={{!selected}}></menu-button>
-        <menu-button on-click="renameClicked" name="rename" icon-name="code" disabled={{!selected}}></menu-button>
-        <menu-button on-click="deleteClicked" name="delete" icon-name="delete" disabled={{!selected}}></menu-button>
-      </section>
-    </div>
-    <div id="reg-area">
-      <div id="search-bar">
-        <iron-icon icon="search"></iron-icon>
-        <input type="search" id="search" value="{{filterText::input}}" placeholder="search registry">
+    <div class="z-container">
```
maffoo: is this outer container needed? I don't see any effect if I remove it. I only ask because it would be nice to keep non-semantic markup to a minimum, but not a big deal.

maffoo:
A few comments, otherwise LGTM. Good find figuring out a way to fix the rendering glitch.

client-js/app/elements/labrad-registry.html
joshm-3b6b58, line 15: Yeah, I agree. I'm sure there's a correct way to do this, I just don't know it. Things like this I can keep in mind and fix when I learn it.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: Chronologically, I had to add it to get it to work when I was debugging this, but I didn't try just removing it at the end. It it works without it, I'll get rid of it.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: These might make more sense as as classes as they may be re-usable in other parts of the app, of course it would then make sense to remove the `reg` identifier. They're not selected anywhere in the code so they don't need to be ids

maffoo: Since these are inside a DOM module, the rules only apply to this tree (due to the same style encapsulation that bit us with paper-material). So no need to worry about making these styles reusable.

PR #51, opened 2015-09-23
remove magic numbers from menu alignment

joshm-3b6b58:
Edited css:

added a css variable to define the height of the menu in the maintoolbar. Locating registry editor menu based on that. 
Also, fixed text in maintoolbar so it appears.


DanielSank:
LGTM

Issue #61, opened 2015-09-24
Make keys editable


DanielSank:
It does not appear possible to edit registry key values.

maffoo:
Select a key and then press rename. It might be nice to have direct editing like we do for the values, but that can't work with directories (they are links so clicking on them navigates).

joshm-3b6b58:
It works for directories just click where the value would be in the table to select it

DanielSank:
I'm not talking about renaming keys; I want to set the value of a key and this does not work. If I click on a key's row then the row turns grey and that's it.

DanielSank:
When I click on a row for the key `f10` the chrome dev console window shows

```
selectType key selDir null selKey f10
```

but nothing happens.

joshm-3b6b58:
Click on the value of the key and it should be editable. Just start typing. Press "shift+enter" to submit and you should change your key.

It should be clear that the key is editable, the field will have focus and your cursor will apear in the value field

DanielSank:
When I click on the key value the row turns grey and nothing else happens. Typing on the keyboard does nothing. There's no cursor.

joshm-3b6b58:
So in my browser, before clicking anything it looks like this:
![screenshot from 2015-09-24 17 10 38](https://cloud.githubusercontent.com/assets/8173925/10090047/76714c70-62df-11e5-9c47-b3bd2c7af741.png)
When I click on a value (in this case clicking on the string "five2r" it looks like this:
![screenshot from 2015-09-24 17 11 16](https://cloud.githubusercontent.com/assets/8173925/10090053/82312d00-62df-11e5-9965-9f357e6b9c91.png)

Is that not how it works for you?

DanielSank:
@joshmutus Great! Mine does not do that.

joshm-3b6b58:
:/  Are you using Chrome?

joshm-3b6b58:
Did you try refreshing the page? 
Are there any weird compile errors in gulp or sbt?

DanielSank:
@joshmutus 
- I'm using Chrome.
- I checked out the repo this afternoon and am on revision ab30f1709a06deed69fb0642514860a9f7b531e0.
- I am coordinated enough to click on the target area. I have clicked on everything imaginable and nothing gives me the opportunity to edit the item.
- There are no errors in the console window.
- No gulp errors.
- There is an error from the sbt process when I refresh the page:

```
[...]

PR #66, opened 2015-09-25
Edit registry values in dialog box instead of inline.

### LINKED ISSUE #60: Registry editor doesn't work, provides no feedback.
When I modify a registry value, it shows up as changed, when I navigate to a different folder and back, the change disappears.

maffoo:
This is much simpler than the inline editing due to the modal interface and generally it just works (does not rely on so many quirks of polymer element behavior). If we have time in the future to tweak and test, we can revisit the inline editing.

Fixes #60 and #61.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -113,11 +113,7 @@
                 <dir-item kind="key">{{item.name}}</dir-item>
               </td>
               <td class="label-wide">
-                <form is="iron-form" id="{{item.name}}">
-                  <iron-a11y-keys keys="tab" on-keys-pressed="incrementSelector">
-                      <expandable-input name="{{item.name}}" value="{{item.value}}"></expandable-input>
-                  </iron-a11y-keys>
-                </form>
+                <paper-item key-name="{{item.name}}" on-dblclick="editValueClicked">{{item.value}}</paper-item>
               </td>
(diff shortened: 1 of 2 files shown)


joshm-3b6b58:
I'm a little bummed we're going away from the inline editing. Hopefully we can circle back and fix it. 

Also there's a bug where if I have a first key selected, then click on a second key, the dialog opens prompting me to edit the first key.

maffoo:
@joshmutus, I'm unable to reproduce the issue you mentioned. If I select key 'A' and then click on the value for key 'B', the dialog shows me the value of 'B' to edit. Any other repro instructions?

One thing that might be happening is that after trying to figure out the problem @DanielSank was having with not being able to edit at all, I blew away and reinstalled all my bower components. It's possible there's some difference in our bower packages that is causing issues here too.

maffoo:
Also, re: direct editing, I also think it'd be nice to resurrect it; one of these days when we have the time to get it right.

maffoo:
@joshmutus, when you get a chance, can you try again? Fixed up the way data binding was working and it no relies on selection to pick the right key name.

joshm-3b6b58:
Cool, bug fixed. On my end it brings up the dialog on single click. It would be nice to make this on double click so we can select without editing keys.

joshm-3b6b58:
LGTM

PR #70, opened 2015-09-25
Display breadcrumbs in main toolbar and clean up secondary toolbar in registry view.

### LINKED ISSUE #56: Keep registry path still while scrolling
The registry path disappears when we scroll down in a registry directory. The path should stay put like the copy/rename/etc. buttons.

maffoo:
With these changes, the registry view looks as follows:

screen shot 2015-09-25 at 2 57 58 pm

Fixes #56

### DIFF
--- a/client-js/app/elements/labrad-breadcrumbs.html
+++ b/client-js/app/elements/labrad-breadcrumbs.html
@@ -3,6 +3,7 @@
 <dom-module id="labrad-breadcrumbs">
   <style>
     ul {
+      display: inline-block;
       padding: 0;
     }
     li {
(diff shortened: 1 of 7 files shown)


joshm-3b6b58:
I really like how this looks, but I think it makes more sense to use a `paper-header-panel` here, like in `index.html` that way the top of the scroll bar won't be covered by the toolbar. I had the floating menu before because I was just avoiding covering up the breadcrumbs before. 

Also, the buttons are no longer disabled or enabled depending on what is selected.

maffoo:
Added the button enable/disable styles. I tried a number of variations of using paper-header-panel, but couldn't get things to work; it doesn't seem to want to take up the full screen when a paper-header-panel is nested inside another paper header panel. I propose to delay fixing that issue to a subsequent PR. @joshmutus, thoughts?

joshm-3b6b58:
Agreed. LGTM.

PR #71, opened 2015-09-25
Only trigger app-link event if left button is pressed.

### LINKED ISSUE #67: middle click doesn't work
middle-clicking on a link should open a new tab/window.  Instead it operates like a left click.

maffoo:
If using the middle button, the default link behavior will occur, which typically opens a link in a new tab.

@ejeffrey, I don't have a mouse with a middle button. Feel free to test this is you like, or else I'll be able to get to this on monday.

Fixes #67

### DIFF
--- a/client-js/app/elements/app-link.html
+++ b/client-js/app/elements/app-link.html
@@ -10,7 +10,7 @@
     },
 
     onClick: function(e) {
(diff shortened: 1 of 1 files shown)


ejeffrey:
LGTM

PR #74, opened 2015-09-30
add drag and drop to registry editor

joshm-3b6b58:
This should be working essentially. I think there are issues with the feature as I've implemented it. Right now, it has the ability to drag entire folders into entire folders. 

Adding the ability to copy keys is easy to add.

I'm a bit concerned that it feels like "move" but is actually "copy" and that will be confusing to users. 

It's lacking the feature to drop into a window and place it in the folder that is contained in that window. That feature has an edge case that it wouldn't make sense if the window was full of directories, there'd be no place to drop it. I've started to add a "Drop here" dialog to handle this, but it would be nice to get some feedback..

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -104,7 +104,12 @@
         <tbody is="selectable-table" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">
           <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc"> 
             <tr name="{{item.name}}">
-              <td class="dir" name="{{item.name}}" draggable="true">
+              <td class="dir" 
+                  name="{{item.name}}" 
+                  draggable="true" 
+                  ondragover="classList.add('over')" 
+                  ondragleave="classList.remove('over')" 
+                  on-drop="dirDrop">
                 <a is="app-link" path="{{item.url}}" href="{{item.url}}">
                   <dir-item kind="folder" >{{item.name}}</dir-item>
                 </a>
@@ -157,10 +162,7 @@
     </div>
 
     <!-- Dialogs for various registry operations -->
-    <paper-dialog id="dropTarget">
-      <h1>Drop here to place in this folder</h1>
-    </paper-dialog>
-    
+
     <paper-dialog id="newKeyDialog" modal>
       <h1>Create new key</h1>
       <form>
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -102,16 +102,16 @@
     <div id="reg-area">
       <table>
         <tbody is="selectable-table" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">
-          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc"> 
+          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc">
             <tr name="{{item.name}}">
               <td class="dir" 
                   name="{{item.name}}" 
                   draggable="true" 
-                  ondragover="classList.add('over')" 
-                  ondragleave="classList.remove('over')" 
+                  on-dragover="onDirDragOver" 
+                  on-dragleave="onDirDragLeave" 
                   on-drop="dirDrop">
                 <a is="app-link" path="{{item.url}}" href="{{item.url}}">
-                  <dir-item kind="folder" >{{item.name}}</dir-item>
+                  <dir-item kind="folder">{{item.name}}</dir-item>
                 </a>
               </td>
               <td class="label-wide"></td>
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -1,4 +1,5 @@
 <link rel="import" href="../bower_components/paper-input/paper-textarea.html">
+<link rel="import" href="../bower_components/paper-toast/paper-toast.html">
 <link rel="import" href="app-link.html">
 <link rel="import" href="dir-item.html">
 
@@ -41,6 +42,10 @@
 
     tbody {
       display:table-header-group;
+    }
+    
+    td.dir.over {
+      background-color: #93d4d4;
     }
 
     .table-header {
@@ -97,11 +102,11 @@
     <div id="reg-area">
       <table>
         <tbody is="selectable-table" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">
-          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc">
+          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc"> 
             <tr name="{{item.name}}">
-              <td>
+              <td class="dir" name="{{item.name}}" draggable="true">
                 <a is="app-link" path="{{item.url}}" href="{{item.url}}">
-                  <dir-item kind="folder">{{item.name}}</dir-item>
+                  <dir-item kind="folder" >{{item.name}}</dir-item>
                 </a>
               </td>
               <td class="label-wide"></td>
@@ -110,7 +115,7 @@
         </tbody>
         <tbody is="selectable-table" attr-for-selected="name"  selected="{{selKey}}" on-tap="selectKey">
           <template id="keyList" is="dom-repeat" items="{{keys}}" filter="filterFunc">
-            <tr name="{{item.name}}">
+            <tr class="key" name="{{item.name}}">
               <td>
                 <dir-item kind="key">{{item.name}}</dir-item>
               </td>
@@ -152,9 +157,12 @@
     </div>
 
     <!-- Dialogs for various registry operations -->
+    <paper-dialog id="dropTarget">
+      <h1>Drop here to place in this folder</h1>
+    </paper-dialog>
+    
     <paper-dialog id="newKeyDialog" modal>
       <h1>Create new key</h1>
-
       <form>
         <paper-input id="newKeyInput" label="Key"></paper-input>
         <paper-textarea id="newValueInput" label="Value"></paper-textarea>
@@ -183,6 +191,18 @@
       </form>
       <div class="buttons">
         <paper-button on-click="doNewFolder" dialog-confirm autofocus>OK</paper-button>
+        <paper-button dialog-dismiss>CANCEL</paper-button>
+      </div>
+    </paper-dialog>
+    
+     <paper-dialog id="dragCopyDialog" modal>
+      <h1>Copy "<span id="originName"></span>" from <span id="originPath"></span></h1>
+      <form>
+        <paper-input id="dragCopyNameInput" label="New Name" always-float-label></paper-input>
+        <paper-input id="dragCopyPathInput" label="New Path" always-float-label></paper-input>
+      </form>
+      <div class="buttons">
+        <paper-button on-click="doDragCopy" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
@@ -217,5 +237,7 @@
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
+    
+    <paper-toast id="toastText" text="tada"></paper-toast>
   </template>
 </dom-module>
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -129,18 +129,26 @@ export class LabradRegistry extends polymer.Base {
 
   //Drag and drop
   @listen('dragenter')
-  @listen('dragover')
-  suppressBehaviour(event) {
-    //change colors of cells to indicate target
+  overBehaviour(event) {
     event.preventDefault(); //I don't understand this.
-    event.target.closest('td').classList.add('over');
+  }
+
+  @listen('dragover')
+  changeBehaviour(event) {
+    //allows the ctrl key to be pressed to change cursor between copy/move
+    event.preventDefault(); //I don't understand this. But needs to be here
+
+    if (event.ctrlKey) {
+      event.dataTransfer.dropEffect = "copy";
+    }
+    else {
+      event.dataTransfer.dropEffect = "move";
+    }
   }
 
   @listen('dragleave')
   leaveDrag(event) {
-    //change colors of cells back to normal
-    event.preventDefault(); //I don't understand this.
-    event.target.closest('td').classList.remove('over');
+    event.preventDefault(); //I don't understand this. But it needs to be here
   }
 
   @listen('dragstart')
@@ -148,17 +156,10 @@ export class LabradRegistry extends polymer.Base {
     //detect start of drag event, grab info about target
     var data: any; //I tried enum, but kept getting errors...
     data = {path: this.path, name: event.target.name};
-    //console.log('drag started', this.path);
     event.dataTransfer.setData('text', JSON.stringify(data));
-    event.dataTransfer.effectAllowed = 'move';
-    if (event.ctrlKey) {
-      this.$.toastText.text = "copy item...";
-      this.$.toastText.show();
-    }
-    else {
-      this.$.toastText.text = "move... not implemented yet :/ ctrl+click to copy";
-      this.$.toastText.show();
-    }
+    event.dataTransfer.effectAllowed = 'copyMove';
+    this.$.toastText.text = "drag to move, hold ctrl to copy";
+    this.$.toastText.show();
   }
 
   @listen('dragend')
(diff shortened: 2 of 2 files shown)


client-js/app/elements/labrad-registry.ts
```diff
@@ -128,6 +128,69 @@ export class LabradRegistry extends polymer.Base {
...
+          copyOriginName = this.$.originName,
+          copyOriginPath = this.$.originPath,
+          copyNameElem = this.$.dragCopyNameInput,
+          copyPathElem = this.$.dragCopyPathInput;
+      copyNameElem.value = data.name;
+      copyOriginName.textContent = data.name;
+      copyOriginPath.textContent = JSON.stringify(data.path);
+      copyPathElem.value = this.pathToString( newPath );
+      dialog.open();
+      swindow.setTimeout(() => copyNameElem.$.input.focus(), 0);
```
maffoo: typo: `swindow`

client-js/app/elements/labrad-registry.ts
```diff
@@ -128,6 +128,69 @@ export class LabradRegistry extends polymer.Base {
...
+
+  @listen('drop')
+  handleDrop(event) {
+
+    event.preventDefault();
+    var data: any;
+    data = JSON.parse(event.dataTransfer.getData('text'));
+
+    console.log('dropped on', event.target.closest('tr').name, 'with data', data ); 
+    if ( event.target.closest('tr').className.includes('dir') ) {
```
maffoo: style guide says no extra spaces in parentheses. Same thing in several places below.

maffoo:
As we discussed, I would prefer to have most of the drop area serve as a target for copying into the current directory, instead of into subdirectories. Copying into subdirectory would happen only if dropping directly onto the directory name.

It would also be nice to allow copying or moving using modifier keys with the drag, as is common for desktop environments.

joshm-3b6b58:
I don't know if it bumps the thread when I push a new commit, but I pushed a new commit yesterday.

maffoo:
This doesn't work for me on mac. Ctrl-click is equivalent to right click and so brings up a context menu instead of starting a drag. The way I'm used to this sort of thing working, you can start the drag and then change modifier keys mid drag to determine whether it's going to be a copy or move operation (if the icon can change accordingly to give visual feedback, that would be great).

I also noticed something strange which is that the paper-drawer-panel seems to be capturing drag events at the far left of the window; if I click on a directory icon and start dragging the drawer panel slides in from the left, rather than starting a drag on the directory item itself. If I select the directory row and then click on the directory icon I can get it to capture the drag even instead. I wonder if there's a way to disable the drawer panel drag behavior; I think having the hamburger button to toggle it is sufficient.

maffoo:
It looks like adding the attribute `disable-edge-swipe` to the `paper-drawer-panel` will disable the drag-drop capture by the hidden drawer panel.

joshm-3b6b58:
I will try to fix the ctrl-drag behaviour when I get around to this tomorrow evening at the latest. 

I'll add the `disable-edge-swipe` as well.

joshm-3b6b58:
So I added the ctrl-modifier shows up in the cursor. There's a little hiccup that the curson won't update until it's moved slightly.

Does this solve the ctrl-click issue on mac?

I had tried adding `disable-edge-swipe` into the `paper-drawer-panel` tag but that didn't work for me for some reason..

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,98 @@ export class LabradRegistry extends polymer.Base {
     );
   }
 
+  //Drag and drop
+  @listen('dragenter')
+  @listen('dragover')
+  suppressBehaviour(event) {
+    //change colors of cells to indicate target
+    event.preventDefault(); //I don't understand this.
+    event.target.closest('td').classList.add('over');
```
maffoo: when I drag a folder, I get lots of errors from this line:

```
Uncaught TypeError: Cannot read property 'classList' of null
```

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,98 @@ export class LabradRegistry extends polymer.Base {
...
+      event.dataTransfer.dropEffect = "move";
+    }
+    
+  }
+
+  @listen('dragleave')
+  leaveDrag(event) {
+    //change colors of cells back to normal
+    event.preventDefault(); //I don't understand this.
+    event.target.closest('td').classList.remove('over');
```
maffoo: same error as above also thrown by this line

joshm-3b6b58:
@maffoo restructured code to fix errors. I think it's a little easier to understand how it works now, too. Let me know what you think.

joshm-3b6b58:
bump?

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
   }
 
+  //Drag and drop
+  @listen('dragenter')
+  overBehaviour(event) {
+    event.preventDefault(); //I don't understand this.
+  }
+
+  @listen('dragover')
+  changeBehaviour(event) {
```
maffoo: I would recommend naming all these methods after the events they listen to in a consistent way, e.g.

``` typescript
@listen('dragenter')
onDragEnter(event) { ... }

@listen('dragover')
onDragOver(event) { ... }

...
```

client-js/app/elements/labrad-registry.html
```diff
@@ -94,11 +99,16 @@
...
         <tbody is="selectable-table" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">
-          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc">
+          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc"> 
             <tr name="{{item.name}}">
-              <td>
+              <td class="dir" 
+                  name="{{item.name}}" 
+                  draggable="true" 
+                  ondragover="classList.add('over')" 
+                  ondragleave="classList.remove('over')"
```
maffoo: we should try to keep code separate from layout; can these be moved to the .ts file?

client-js/app/elements/labrad-registry.html
```diff
@@ -94,11 +99,16 @@
     <div id="reg-area">
       <table>
         <tbody is="selectable-table" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">
-          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc">
+          <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc">
```
maffoo: nit: remove trailing whitespace

client-js/app/elements/labrad-registry.html
```diff
@@ -94,11 +99,16 @@
...
-              <td>
+              <td class="dir" 
+                  name="{{item.name}}" 
+                  draggable="true" 
+                  ondragover="classList.add('over')" 
+                  ondragleave="classList.remove('over')" 
+                  on-drop="dirDrop">
                 <a is="app-link" path="{{item.url}}" href="{{item.url}}">
-                  <dir-item kind="folder">{{item.name}}</dir-item>
+                  <dir-item kind="folder" >{{item.name}}</dir-item>
```
maffoo: nit: no extra whitespace after tag attributes

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+  @listen('dragend')
+  endDrag(event) {
+    console.log('drag ended',event);
+  }
+
+  dirDrop(event) {
+    //behaviour for dropping on folders
+    console.log(event);
+    var data: any;
+    data = JSON.parse(event.dataTransfer.getData('text'));
```
maffoo: You can just say

```
var data = JSON.parse(event.dataTransfer.getData('text'));
```

The compiler will infer a type of any (which I presume is what `JSON.parse` returns).

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+  @listen('dragleave')
+  leaveDrag(event) {
+    event.preventDefault(); //I don't understand this. But it needs to be here
+  }
+
+  @listen('dragstart')
+  startDrag(event) {
+    //detect start of drag event, grab info about target
+    var data: any; //I tried enum, but kept getting errors...
+    data = {path: this.path, name: event.target.name};
```
maffoo, line 166:
```
var data = {path: this.path, name: event.target.name};
```

I don't understand the comment about 'enum'.

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+  }
+
+  @listen('dragleave')
+  leaveDrag(event) {
+    event.preventDefault(); //I don't understand this. But it needs to be here
+  }
+
+  @listen('dragstart')
+  startDrag(event) {
+    //detect start of drag event, grab info about target
```
maffoo, line 164: We haven't been very consistent about format of comments, but the styleguide recommends JSDoc format (https://google.github.io/styleguide/javascriptguide.xml?showone=Comments#Comments) which puts the comments before the method and uses `/*\* */ as delimiters:

``` typescript
/**
 * Detect start of drag event and grab info about target.
 */
@listen('dragstart')
startDrag(event) {
  ...
}
```

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+    console.log(event);
+    var data: any;
+    data = JSON.parse(event.dataTransfer.getData('text'));
+    console.log('dirDrop');
+    event.target.closest('td').classList.remove('over');
+    if (event.ctrlKey) {
+      console.log('dropped on', event.target.closest('td').name, 'with data', data ); 
+      console.log(event.target);
+      var newPath: string[] = this.path.splice(0);
+      var oldFullPath: string[] = data.path;
```
maffoo: I have been using `Array<string>` for array types, rather than `string[]`. That's probably a residual scalaism, but in any case we should be consistent. The reason I prefer the former is that it looks like any other generic type that we might create for custom classes, such as `Observable<T>`, even if it's a bit more verbose. But I could be persuaded.

Of course in this case explicit types are not needed because the compiler will infer them, but it can still sometimes be useful for documentation.

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+      window.setTimeout(() => copyNameElem.$.input.focus(), 0);
+    }
+  }
+
+  @listen('drop')
+  handleDrop(event) {
+    //handles folders dropped not on other folders
+    event.preventDefault();
+    var data: any;
+    data = JSON.parse(event.dataTransfer.getData('text'));
```
maffoo: same comment as above: `var data = JSON.parse(...);`

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+    console.log('dropped', event.dataTransfer);
+    if (event.ctrlKey) {
+      if (JSON.stringify(this.path) != JSON.stringify(data.path)){
```
maffoo: nit: whitespace before open brace

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+        //copying into window by dropping not on a folder, but 
+        //can't copy into the same path. Comparing string is lame, I know
+        console.log('copying into window', this.path, data.path);
+        var oldFullPath: string[] = data.path;
+        var dialog = this.$.dragCopyDialog,
```
maffoo: I would suggest either coalescing all var declarations, or none of them, e.g.

``` typescript
var oldFullPath = data.path,
    dialog = this.$.dragCopyDialog,
    copyOriginName = this.$.originName,
    ...
```

client-js/app/elements/labrad-registry.ts
```diff
@@ -268,6 +363,24 @@ export class LabradRegistry extends polymer.Base {
...
+  doDragCopy() {
+    var self = this;
+    var newName =  this.$.dragCopyNameInput.value;
+    var newPath = JSON.parse(this.$.dragCopyPathInput.value);
+    var oldPath = JSON.parse(this.$.originPath.textContent);
+    var oldName = this.$.originName.textContent;
+
+    this.socket.copyDir({path: oldPath, dir: oldName, newPath: newPath, newDir: newName}).then(
+      (resp) => {
+        self.repopulateList(resp);
```
maffoo: You'll want to change this to match the other api calls after rebasing, since `copyDir` no longer returns a directory listing.

client-js/app/elements/labrad-registry.ts
```diff
@@ -129,6 +129,101 @@ export class LabradRegistry extends polymer.Base {
...
+  }
+
+  @listen('dragstart')
+  startDrag(event) {
+    //detect start of drag event, grab info about target
+    var data: any; //I tried enum, but kept getting errors...
+    data = {path: this.path, name: event.target.name};
+    event.dataTransfer.setData('text', JSON.stringify(data));
+    event.dataTransfer.effectAllowed = 'copyMove';
+    this.$.toastText.text = "drag to move, hold ctrl to copy";
```
maffoo: might want to change this since move is not currently implemented.

maffoo:
On linux, it seems the source and destination paths are not getting properly set. For example, I tried dragging directory `['test', 'quux']` onto `['test', 'qux']` and the dialog came up wanting to copy `quux` from path `[]` to `['qux']`. Another time, the destination path came up as `[]` too. Haven't looked to try and figure out where the data is getting lost.

On mac, does not work at all due to modify keys; if I try to hold ctrl when initially clicking to start the drag, it triggers a right click and opens a context menu, but if I start dragging and then press ctrl, it doesn't seem to properly change the drop operation (no '+' appears on the cursor and the copy dialog doesn't trigger when I actually drop the directory).

joshm-3b6b58:
@maffoo trying to track down the behavior on mac. There may be a bug where you have to move the mouse slightly after pressing ctrl or the dragover doesn't update. Does drop work on mac if you:

1) start drag
2) press and hold ctrl
3) move mouse onto target
4) drop onto target

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Yup

client-js/app/elements/labrad-registry.ts

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: for `window.setTimeout( () => copyNameElem.$.input.focus(), 0);` ?

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Right

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: I don't feel strongly either way, so I can change this.

client-js/app/elements/labrad-registry.ts
joshm-3b6b58, line 164: OK

client-js/app/elements/labrad-registry.ts
joshm-3b6b58, line 166: I wanted to define a type for the object data, but I couldn't without errors. http://www.typescriptlang.org/Handbook say to use `enum` for this data type, but I was confused about the syntax

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: OK will do.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: Sure, I can figure that out.

client-js/app/elements/labrad-registry.ts
maffoo, line 166: It seems to me an enum is not the right thing here. In any case, if you just combine the var declaration and initialization then the compiler will infer a type of `{path: Array<string>; name: string}`. This is all a bit academic anyway, since the only thing that's happening is to immediately stringify this.

client-js/app/elements/labrad-registry.ts
maffoo: I meant this:

```
if (condition) { // yes

if (condition){ // no
```

maffoo:
I thought moving the mouse might be the issue as well and tried exactly the steps 1-4 you suggest, but still no luck on mac.

joshm-3b6b58:
OK, well the first error you found I fixed I was using `this.path.splice` where I should've been using `this.path.slice` for former clears the first array and returns values the latter does what I wanted. 

I'll try troubleshooting the mac issue when I'm near my laptop next. I can't begin to speculate what the issue is there.

joshm-3b6b58:
@maffoo if you get the chance, can you see if copy works on osx now? I changed `if( event.ctrlKey )` to `if( event.ctrlKey || event.button == 2 )` which is what stackoverflow told me to do. I can try it later when I get home if you can't get around to it right now.

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Ah line 207?

maffoo:
If I hold down ctrl+alt then it works, though `event.button == 0` in this case, so the extra check doesn't help. The way these events are generated is still mysterious to me.

client-js/app/elements/labrad-registry.ts
maffoo: 208

joshm-3b6b58:
wacky. OK I'll mess with it at home.

maffoo:
To be clear, I'd be ok with merging this and moving on; we can sort out the finer details of event handling later. When I do a drag/drop in _most_ systems I start dragging then fiddle with the modifier keys until the cursor indicates the operation I want, then drop. The only thing I would suggest here is get rid of the toast since the text there is misleading. Aside from my previous comments (keep code in .ts file, whitespace, etc), LGTM.

joshm-3b6b58:
OK cool. This is just bugging me, so I want to take a look at home. I haven't spent any time in front of a mac in writing this so maybe it'll be a quick fix.

Thanks for the comments.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: nice catch

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Oh, I didn't know that. Neat.

joshm-3b6b58:
@maffoo changed according to comments, I don't know if you want to take a quick look.

PR #76, opened 2015-10-01
Add login dialog to prompt user to enter labrad credentials

maffoo:
Also refactored main app so that UI lives in a custom element, rather than in an autobound template in index.html, so that we can treat it more like all the other custom elements we have.

When testing, be sure to restart both the scala and node dev servers, as some server APIs have changed.


joshm-3b6b58:
How do I log out? I closed my chrome browser completely (I think) but when I went to localhost:3000 I wasn't prompted to re-enter my password.

maffoo:
Good point; I should add that :) Right now the credentials are stored in localStorage, which essentially lasts forever. I could store them in sessionStorage instead (in which case they'd go away when you close the browser window), and/or have an option to explicitly log out.

joshm-3b6b58:
I think the lasts forever but with the option to log out is best for now.

maffoo:
@joshmutus, did you take a look at the latest changes? I could make the "Remember password" option default to true, or just remove it altogether. Let me know what you think.

joshm-3b6b58:
I never use remember password features since I forget my passwords if I
don't enter them constantly. I would just keep it as it is defaulted to
false.

The logout button is very easy to see :)

joshm-3b6b58:
Otherwise LGTM

PR #77, opened 2015-10-01
Use arrow functions with lexical this instead of self and bind.

maffoo:
I just learned (and verified with manual tests) that arrow functions in typescript (and es6) capture `this` from the surrounding lexical scope, so there's no need for the `var self = this` hackery when using arrow functions. Changed a number of places in labrad-registry to reflect this, and also used arrow functions in place of `bind` in a few spots where methods are being invoked as callbacks. I'm really glad to have found this out, as it fixes one of the big annoyances with javascript callbacks.


joshm-3b6b58:
Nice! Checked out branch, tested registry functions.
LGTM

PR #80, opened 2015-10-07
Fix hashbang config on login.

maffoo:
Had the wrong hashbang config for `page` when logging in with saved credentials. This causes the app to use old-school hashbang urls like `http://localhost/#!/nodes/` instead of nice clean urls like `http://localhost/nodes/`.


joshm-3b6b58:
LGTM

PR #82, opened 2015-10-09
Fix registry dir copy to work when copying dir into itself.

### LINKED ISSUE #73: prevent from copying directories into themselves
Being able to copy a directory itself makes a very deep recursive structure of directories containing itself. 

My head hurts even trying to describe the issue.

maffoo:
Previously the directory copy operation would create the destination directory before getting a listing of the source directory. As a result, copying a directory into itself led to an infinite recursion. If we instead get the source listing first, then the problem goes away and the copying a directory into itself is a perfectly well-behaved operation.

We also change the way the registry api methods work so that only the `dir` call returns a registry listing, and the others return nothing. The client must then request a new directory listing after it makes another call to manipulate the registry. There are a few reasons for doing this. First, in operations such as copy and move, the listing returned may not actually correspond to the directory the page is showing; sometimes we may be viewing the source directory, other times the destination directory (for example in a copy or move triggered by drag and drop) or some other directory entirely (e.g. in drag and drop from another window when we then edit the destination). It's best to just let the client ask for what it wants. Second, if in the future we decide to batch operations (e.g. to edit multiple keys or something like that), then we want to make sure we don't transfer lots of listing data unnecessarily. Finally, once we implement events so that the client gets notified of registry changes in a directory, we won't need these directory listings at all.

Fixes #73

### DIFF
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -77,17 +77,20 @@ export class LabradRegistry extends polymer.Base {
     //TODO increment selected key on tab
   }
 
-  repopulateList(resp: RegistryListing) {
-    this.selDir = null;
-    this.splice('dirs', 0, this.dirs.length);
-    this.splice('keys', 0, this.keys.length);
+  repopulateList(): Promise<void> {
+    return this.socket.dir({path: this.path}).then((resp) => {
+      this.selKey = null;
+      this.selDir = null;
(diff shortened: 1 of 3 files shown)


joshm-3b6b58:
@maffoo so I can still copy a directory into itself, but I don't get the recursion. This is working as intended?

maffoo:
Yeah, that's the idea.

joshm-3b6b58:
OK great. LGTM

PR #86, opened 2015-10-16
Flesh out grapher ui

maffoo:
This fixes up the grapher ui and adds basic plotting for 1D datasets (there's a plot element for 2D datasets as well, but it's not fully working yet). I've also started refactoring the way page navigation works, introducing an interface called `Activity` to contain most of the logic like communicating with the server and responding to server events, and allowing us to simplify the elements themselves, which are awkward to use compared to basic js object due to things like properties getting set after the element is constructed.

Sorry for the size and relative disorganization of this commit, but I figure it's better to have people try it out than spend too much time reorganizing the change. To test, you'll obviously want to run a local data vault server, and I've created a script to produce some demo data sets in `server/src/test/scripts/data_demo.py`. You can run that and then watch data coming in live. And we can add more test cases to that script to exercise other features of the grapher ui as we continue to build it out.


joshm-3b6b58:
@maffoo do I have to install some packages to get this to work? I'm getting and error:

```
err  Error: Error loading "github:mbostock/d3@3.5.6" at file:/usr/local/google/home/mutus/programming/scalabrad-web/client-js/jspm_packages/github/mbostock/d3@3.5.6.js
     Error loading "github:mbostock/d3@3.5.6" from "scripts/app" at file:/usr/local/google/home/mutus/programming/scalabrad-web/client-js/.tmp/scripts/app.js
     ENOENT, open '/usr/local/google/home/mutus/programming/scalabrad-web/client-js/jspm_packages/github/mbostock/d3@3.5.6.js'
         at Error (native)
```

maffoo:
`jspm install` should fix that error.

joshm-3b6b58:
Yup, that did the trick. Thanks.

joshm-3b6b58:
~~I'm a bit lost following the trail of breadcrumbs here. I can't figure out how the plot1D actually renders on the page.~~
~~AFAICT the DOM tree is something like:~~

~~index.html (it has to start here) > labrad-app.html (this has the paper drawer) > labrad-pages (but I can't explicitly see how these are loaded) > labrad-page-grapher > labrad-grapher .... and I'm lost. I can only see how the directory structure is loaded here but not the plots. I'm guessing this happens in app.ts?~~

OK I found `DatasetActivity`.

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -0,0 +1,255 @@
...
+  private chartBody: any;
+  private clip: any;
+  private xAxis: any;
+  private yAxis: any;
+  private xScale: any;
+  private yScale: any;
+  private line: any;
+  private zoom: any;
+  private limits = {xMin: 0, xMax: 1, yMin: 0, yMax: 1};
+  private dataLimits = {xMin: NaN, xMax: NaN, yMin: NaN, yMax: NaN};
```
joshm-3b6b58, line 32: Why NaN here?

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -0,0 +1,255 @@
...
+    p.line = d3.svg.line()
+            .x((d) => p.xScale(d[0]))
+            .y((d) => p.yScale(d[1]));
+
+    p.zoom = d3.behavior.zoom()
+            .x(p.xScale)
+            .y(p.yScale)
+            .on('zoom', p.handleZoom.bind(p));
+
+    // plot area
```
joshm-3b6b58, line 101: Should these comments have the 

```
/**
*
*/
```

I think these are vestigial.

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.dataLimits.xMax = this.safeMax(this.dataLimits.xMax, x);
+      for (let i = 0; i < this.numTraces; i++) {
+        let y = row[i+1];
+        this.dataLimits.yMin = this.safeMin(this.dataLimits.yMin, y);
+        this.dataLimits.yMax = this.safeMax(this.dataLimits.yMax, y);
+      }
+    }
+
+    // calculate view limits
+    this.limits.xMin = isNaN(this.dataLimits.xMin) ? 0 : this.dataLimits.xMin;
```
joshm-3b6b58, line 201: TIL `isNaN()` is a thing.

client-js/app/elements/labrad-plot2d.ts
```diff
@@ -0,0 +1,322 @@
...
+    p.zoom = d3.behavior.zoom()
+            .x(p.xScale)
+            .y(p.yScale)
+            .on('zoom', p.handleZoom.bind(p));
+
+    // plot area
+    p.svg = d3.select(this)
+            .append('svg:svg')
+            .attr('width', width + p.margin.left + p.margin.right)
+            .attr('height', height + p. margin.top + p.margin.bottom)
```
joshm-3b6b58: space in `p.margin.top`

client-js/app/elements/labrad-server.ts
```diff
@@ -0,0 +1,9 @@
+import {ServerInfo, SettingInfo} from '../scripts/manager';
```
joshm-3b6b58, line 1: What is `labrad-server.ts`??

client-js/app/scripts/app.ts
```diff
@@ -68,156 +103,378 @@ window.addEventListener('WebComponentsReady', () => {
...
+
+    private api: datavault.DataVaultService;
+    private lifetime = new Lifetime();
+
+    constructor(api: datavault.DataVaultService, path: Array<string>) {
+      this.api = api;
+      this.path = path;
+    }
+
+    start(): Promise<ActivityState> {
```
joshm-3b6b58, line 208: This is important and it'd be nice to have a small amount of documentation since this is how we add new interaction with the server. Just something little as a reminder.

server/src/test/scripts/data_demo.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import division
```
joshm-3b6b58, line 1: Omigosh I haven't looked at Python in forever. Weird not seeing semicolons

server/src/test/scripts/data_demo.py
```diff
@@ -0,0 +1,84 @@
...
+    for i in xrange(nx):
+        x = sum(dx[:i])
+        for j in xrange(ny):
+            y = sum(dy[:j])
+            z = 5 * math.cos(4*math.pi * (x**2 + y**2) / rMax2) + random.random()
+            dv.add([x, y, z])
+            time.sleep(0.01)
+
+
+
```
joshm-3b6b58: I probably should know this but what are the whitespace guidelines between functions? There are 3 spaces here, the rest have 2.

joshm-3b6b58:
Looked through. Some super small comments. LGTM

client-js/app/elements/labrad-plot1d.ts
maffoo, line 32: We need a sentinel to indicate that a limit has not been computed (because we have no data yet). Could alternately use `null`, but `NaN` is an in-band sentinel for numbers, so I went with that.

client-js/app/elements/labrad-plot1d.ts
maffoo, line 101: `/** */` is for jsdoc comments on visible parts of the api (e.g. classes, methods, or public attributes). Inline comments should use `//`.

client-js/app/elements/labrad-plot2d.ts
maffoo: fixed

client-js/app/elements/labrad-server.ts
maffoo, line 1: This element shows information about a server (it's doc strings and setting definitions). It's been there for a long time, I just refactored it to put the code in typescript, as with other elements.

client-js/app/scripts/app.ts
maffoo, line 208: I added documentation to the Activity class itself to describe what these methods do.

server/src/test/scripts/data_demo.py
maffoo: fixed to have 2 blank lines throughout.

Issue #84, opened 2015-10-14
Registry editor should receive updates when key values change


DanielSank:

joshm-3b6b58:
@maffoo so the plumbing looks like it's mostly in place for the RPC, we can set it to watch a path when the path is loaded in the browser. The thing I'm confused about is that when a subscribed event is received it just logs it to the console. Did you have any ideas about how to wire this up? We could fire an event, but the RPC isn't affiliated with any DOM element. Should we use the machinery in `observable.ts`?

maffoo:
Yeah, we should add `Observable`s in `registry.ts`, then add listeners in the registry element. If you want to take a crack at this, you can take a look at the way the node does it for an example. I've got some changes coming with the grapher ui that should clean things up quite a bit, but for now we could set it up like the node controller.

joshm-3b6b58:
Cool, I'd like to take a crack at it. That'll make me understand how the `Obervable`s actually work.

PR #88, opened 2015-10-27
Update registry page on change

joshm-3b6b58:
Fires a blanket repopulating of list on a change in the registry page being displayed. Kind of a shotgun approach, but it works. 

Also factored out populating list in `app.ts` as this already happens in `repopulateList()`

Fixes https://github.com/labrad/scalabrad-web/issues/84

Also, I'm appreciating moving this backend code to typescript so can use the type signatures and I can make sure evertyhing is kosher with the compiler to know if I'm doing the right thing :)

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -141,54 +141,55 @@ window.addEventListener('WebComponentsReady', () => {
 
   class RegistryActivity implements Activity {
     path: Array<string>;
+    private elem: LabradRegistry;
 
-    constructor(path: Array<string>) {
+    private api: registry.RegistryApi;
+    private lifetime = new Lifetime();
+
+    constructor(api: registry.RegistryApi, path: Array<string>) {
+      this.api = api;
       this.path = path;
     }
 
     start(): Promise<ActivityState> {
+      this.api.newItem.add(() => this.onNewItem(), this.lifetime);
       console.log('loading registry:', this.path);
-      return reg.dir({path: this.path}).then((listing) => {
-        var breadcrumbs = [];
-        for (var i = 0; i <= this.path.length; i++) {
-          breadcrumbs.push({
-            name: (i == 0) ? 'registry' : this.path[i-1],
-            isLink: i < this.path.length,
-            url: registryUrl(this.path.slice(0, i))
-          });
-        }
+      return this.api.watch({path: this.path}).then(() => {
+        return this.api.dir({path: this.path}).then((listing) => {
+          var breadcrumbs = [];
+          for (var i = 0; i <= this.path.length; i++) {
+            breadcrumbs.push({
+              name: (i == 0) ? 'registry' : this.path[i-1],
+              isLink: i < this.path.length,
+              url: registryUrl(this.path.slice(0, i))
+            });
+          }
 
-        var dirs = [];
-        for (var i = 0; i < listing.dirs.length; i++) {
-          var dir = listing.dirs[i];
-          dirs.push({
-            name: dir,
-            url: registryUrl(this.path, dir)
-          });
-        }
-        var keys = [];
-        for (var i = 0; i < listing.keys.length; i++) {
-          keys.push({
-            name: listing.keys[i],
-            value: listing.vals[i]
-          });
-        }
-        var elem = <LabradRegistry> LabradRegistry.create();
-        elem.path = this.path;
-        elem.dirs = dirs;
-        elem.keys = keys;
-        elem.socket = reg;
+          var elem = <LabradRegistry> LabradRegistry.create();
 
-        return {
-          elem: elem,
-          route: 'registry',
-          breadcrumbs: breadcrumbs
-        };
-      });
+          elem.path = this.path;
+          elem.dirs = [];
+          elem.keys = [];
+          elem.socket = this.api;
+          elem.repopulateList();
+
+          this.elem = elem;
+
+          return {
+            elem: elem,
+            route: 'registry',
+            breadcrumbs: breadcrumbs
+          };
+        });
+     });
+    }
+
+    onNewItem() {
+      this.elem.repopulateList();
     }
 
     stop(): Promise<void> {
-      return Promise.resolve(null);
+      return this.api.unwatch({path: this.path});
     }
   }
 
@@ -470,11 +471,11 @@ window.addEventListener('WebComponentsReady', () => {
   }
 
   page('/registry', () => {
-    activate(new RegistryActivity([]));
+    activate(new RegistryActivity(reg, []));
   });
   function mkRegRoute(n: number) {
     page('/registry/' + pathRoute(n), (ctx, next) => {
-      activate(new RegistryActivity(getPath(ctx)));
+      activate(new RegistryActivity(reg, getPath(ctx)));
     })
   }
   for (var i = 0; i <= 20; i++) {
--- a/client-js/app/scripts/registry.ts
+++ b/client-js/app/scripts/registry.ts
@@ -1,3 +1,4 @@
+import {Observable} from './observable';
 import * as rpc from "./rpc";
 
 export interface RegistryListing {
@@ -8,6 +9,8 @@ export interface RegistryListing {
 }
 
 export interface RegistryApi {
+  newItem: Observable<{name: string}>;
+
   dir(params: {path: Array<string>}): Promise<RegistryListing>;
   set(params: {path: Array<string>; key: string; value: string}): Promise<void>;
   del(params: {path: Array<string>; key: string}): Promise<void>;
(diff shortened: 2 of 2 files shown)


client-js/app/scripts/app.ts
```diff
@@ -141,13 +141,20 @@ window.addEventListener('WebComponentsReady', () => {
 
   class RegistryActivity implements Activity {
     path: Array<string>;
+    private globElem: LabradRegistry;
```
joshm-3b6b58: Using `globElem` is a hack. I tried making this `private elem: LabradRegistry;` and switching the `elem`s to `this.elem` like in `DatavaultActivity` but I get the following error

```
app/scripts/app.ts(158,14): error TS2322: Type 'Promise<{ elem: number; route: string; breadcrumbs: any[]; }>' is not assignable to type 'Promise<ActivityState>'.
  Type '{ elem: number; route: string; breadcrumbs: any[]; }' is not assignable to type 'ActivityState'.
    Types of property 'elem' are incompatible.
      Type 'number' is not assignable to type 'HTMLElement'.
        Property 'accessKey' is missing in type 'Number'.
app/scripts/app.ts(177,17): error TS2362: The left-hand side of an arithmetic operation must be of type 'any', 'number' or an enum type.
app/scripts/app.ts(177,22): error TS2304: Cannot find name 'elem'.
```

client-js/app/scripts/app.ts
```diff
@@ -141,13 +141,20 @@ window.addEventListener('WebComponentsReady', () => {
...
 
-    constructor(path: Array<string>) {
+    private api: registry.RegistryServiceJsonRpc;
```
maffoo: Use the interface type here (`registry.RegistryApi`) rather than an implementation type. This allows swapping out implementations, e.g. for tests.

client-js/app/scripts/app.ts
maffoo: I don't see how changing the name could lead to that error. Seems like you must have been doing something different in the `start` method to get that. Maybe it just wasn't a clean compile?

client-js/app/scripts/app.ts
```diff
@@ -141,13 +141,20 @@ window.addEventListener('WebComponentsReady', () => {
...
+    private lifetime = new Lifetime();
+
+    constructor(api: registry.RegistryServiceJsonRpc, path: Array<string>) {
```
maffoo: here too, use interface type for `api`

client-js/app/scripts/app.ts
```diff
@@ -141,13 +141,20 @@ window.addEventListener('WebComponentsReady', () => {
...
+      this.api = api;
       this.path = path;
     }
 
     start(): Promise<ActivityState> {
+      this.api.watch({path: this.path});
```
maffoo: `watch` is an api call that returns a promise. You want to make sure that the result of that promise gets dealt with, because it could be an error, and also because the activity is not fully ready to start until the call has completed. Something like:

``` typescript
this.api.newItem.add(...);
return this.api.watch({path: this.path}).then(() => {
  this.reg.dir({path: this.path})
}).then((listing) => {
  ...
});
```

maffoo: It's definitely a bit awkward to string together promises like this (arrow functions help, but don't alleviate all the pain). The next version of typescript, 1.7, will have async/await, which will make it possible to do inlineCallbacks-style coroutines instead of writing chains of callbacks like this; it can't come soon enough, IMO (The typescript roadmap is [here](https://github.com/Microsoft/TypeScript/wiki/Roadmap)). Until then, we have to just suffer through the pain.

client-js/app/scripts/app.ts
```diff
@@ -187,7 +183,13 @@ window.addEventListener('WebComponentsReady', () => {
       });
     }
 
+    onNewItem(x) {
```
maffoo: since the parameter is not used, I'd suggest removing it. I used the name `x` in the callbacks elsewhere only because the variable was being discarded (`observable.add(x => doSomethingWithoutX())`). A better name would have been something like "ignored", but in any case I'd suggest just omitting it. If you do want to pass it through, however, then maybe choose a more meaningful name and include the type here.

client-js/app/scripts/app.ts
```diff
@@ -187,7 +183,13 @@ window.addEventListener('WebComponentsReady', () => {
...
+      console.log('got a new item',x);
+      this.globElem.repopulateList();
+    }
+
     stop(): Promise<void> {
+      this.api.unwatch({path: this.path});
```
maffoo: As with watch, this is an api call that returns a promise, so you can just return that result from this method:

``` typescript
stop(): Promise<void> {
  return this.api.unwatch({path: this.path});
}
```

client-js/app/scripts/app.ts
joshm-3b6b58: Sorry, i didn't explain that correctly. It breaks if I changes lines 170-174 and 179 to `this.elem` as well

client-js/app/scripts/app.ts
joshm-3b6b58: So I'm getting a type signature error when I use `RegistryApi` instead. The code in `DatasetActivity` uses `DataVaultService` instead of `DataVaultApi`

client-js/app/scripts/app.ts
joshm-3b6b58: Alright, I soldiered through it and did this. Callbacks will be nice

client-js/app/scripts/app.ts
joshm-3b6b58: This was just leftover from debugging.

client-js/app/scripts/app.ts
joshm-3b6b58: So it's OK to get rid of `return Promise.resolve(null);` ?

client-js/app/scripts/app.ts
joshm-3b6b58: OK, now the error is gone and it seems to be working. 

compilers ¯_(ツ)_/¯

client-js/app/scripts/app.ts
joshm-3b6b58: Done

client-js/app/scripts/app.ts
maffoo: I don't know exactly what you did, but it'd be fine to just change line 176, that is, to still use a local variable `elem` in the `start` method and then assign the value of that variable to `this.elem` as well.

client-js/app/scripts/app.ts
maffoo: yes, just return the Promise created and returned by the call to `unwatch`

client-js/app/scripts/app.ts
joshm-3b6b58: done

joshm-3b6b58:
@maffoo committed changes. Let me know what you think when you have a chance

client-js/app/scripts/app.ts
```diff
@@ -141,54 +141,55 @@ window.addEventListener('WebComponentsReady', () => {
...
-        var breadcrumbs = [];
-        for (var i = 0; i <= this.path.length; i++) {
-          breadcrumbs.push({
-            name: (i == 0) ? 'registry' : this.path[i-1],
-            isLink: i < this.path.length,
-            url: registryUrl(this.path.slice(0, i))
-          });
-        }
+      return this.api.watch({path: this.path}).then(() => {
+        return reg.dir({path: this.path}).then((listing) => {
```
maffoo: use `this.api` instead of `reg`.

client-js/app/scripts/app.ts
```diff
@@ -141,54 +141,55 @@ window.addEventListener('WebComponentsReady', () => {
...
-        return {
-          elem: elem,
-          route: 'registry',
-          breadcrumbs: breadcrumbs
-        };
-      });
+          elem.path = this.path;
+          elem.dirs = [];
+          elem.keys = [];
+          elem.socket = reg;
```
maffoo: use `this.api` instead of 'reg`

client-js/app/scripts/app.ts
maffoo: One of the nice things about promises is that you can avoid levels of nesting and instead string `.then` callbacks together sequentially. Here, for example, you can do:

``` typescript
return this.api.watch({path: this.path})
  .then(() => this.api.dir({path: this.path}))
  .then((listing) => {
    var breadcrumbs = [];
    ...
  });
```

I think this will be even nicer when we can use async functions, as I mentioned previously, but as long as we're using explicit promises I think the sequential composition is clearer than nesting.

maffoo:
A couple of small comments. Otherwise LGTM.

Issue #89, opened 2015-10-29
Improve zooming behavior in grapher interface


maffoo:
A few improvements to the zooming behavior that we'd like to implement:
- zoom to box that user drags with mouse (like current grapher)
- when zoomed in or panned, don't autoscale as new data points come in
- after zooming, add a way to get back to "home view" that shows entire data set (and does autoscaling as new data points come in)

Assigning to @ccwa for her first contribution!

joshm-3b6b58:
Welcome @ccwa! Where are you joining us from?

ccwa:
Thanks, @joshmutus, I am based in MTV.

@maffoo, I have addressed all your comments, pls take a look when you have time. I probably can squeeze more time in the coming weeks, so feel free to assign more grapher related issue to me.

jwenner:
@maffoo, can we now close this with #101 merged in?

maffoo:
closed by #101

joshm-3b6b58:
This was still open, so I closed it

PR #101, opened 2015-11-16
#89 zooming behavior change on 1dplot

ccwa:
dragging a rectangle to zoom for 1d plot.

### DIFF
--- a/client-js/app/elements/labrad-plot1d.ts
+++ b/client-js/app/elements/labrad-plot1d.ts
@@ -22,12 +22,15 @@ export class Plot1D extends polymer.Base implements Plot {
   private svg: any;
   private chartBody: any;
   private clip: any;
+  private width: number;
+  private height: number;
   private xAxis: any;
   private yAxis: any;
   private xScale: any;
   private yScale: any;
   private line: any;
   private zoom: any;
+  private zoomToRect: boolean = true;
   private limits = {xMin: 0, xMax: 1, yMin: 0, yMax: 1};
   private dataLimits = {xMin: NaN, xMax: NaN, yMin: NaN, yMax: NaN};
   private margin: {top: number; right: number; bottom: number; left: number};
@@ -35,6 +38,7 @@ export class Plot1D extends polymer.Base implements Plot {
   private plot: any;
 
   attached() {
+    this.$.pan.style.color = '#AAAAAA';
     this.redraw();
     window.addEventListener('resize', (event) => this.redraw());
   }
@@ -54,21 +58,25 @@ export class Plot1D extends polymer.Base implements Plot {
   }
 
   private redraw() {
-    var rect = this.getBoundingClientRect();
-    while (this.firstChild) {
-      this.removeChild(this.firstChild);
+    var area = this.$.plot,
+        rect = area.getBoundingClientRect();
+    while (area.firstChild) {
+      area.removeChild(area.firstChild);
     }
-    this.createPlot(Math.max(rect.width, 400), Math.max(rect.height, 400));
+    this.createPlot(area, Math.max(rect.width, 400), Math.max(rect.height, 400));
     this.plotData(this.data);
   }
 
-  private createPlot(totWidth: number = 1000, totHeight: number = 500): void {
+  private createPlot(area: HTMLElement, totWidth: number = 1000, totHeight: number = 500): void {
     var p = this;
     p.limits = {xMin: 0, xMax: 100, yMin: 0, yMax: 100};
-    p.margin = {top: 50, right: 50, bottom: 50, left: 40};
+    p.margin = {top: 50, right: 10, bottom: 50, left: 40};
 
     var width = totWidth - p.margin.left - p.margin.right;
     var height = totHeight - p.margin.top - p.margin.bottom;
+
+    p.width = width;
+    p.height = height;
 
     p.xScale = d3.scale.linear()
             .domain([p.limits.xMin, p.limits.xMax])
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-plot1d.ts
```diff
@@ -105,13 +110,20 @@ export class Plot1D extends polymer.Base implements Plot {
             .attr('height', height + p. margin.top + p.margin.bottom)
             .append('svg:g')
             .attr('transform', `translate(${p.margin.left}, ${p.margin.top})`)
-            .call(p.zoom);
-
+            .call(p.zoom)
+	    .append('svg:g')
+	    .on('mousedown', p.zoomRectangle.bind(p));
```
maffoo: indentation looks off here and a few places below. Is there maybe a mix of tabs and spaces?

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -252,4 +258,46 @@ export class Plot1D extends polymer.Base implements Plot {
...
+    this.xAxis.scale(this.xScale);
+    this.yAxis.scale(this.yScale);
+    this.zoom.x(this.xScale);
+    this.zoom.y(this.yScale);
+    this.handleZoom();
+  }
+
+  private zoomRectangle() {
+    var e = this,
+        origin = d3.mouse(e),
```
maffoo: I would suggest using destructuring assignment to pull out the x and y components, instead of using indices throughout:

``` javascript
[originX, originY] = d3.mouse(e),
```

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -252,4 +258,46 @@ export class Plot1D extends polymer.Base implements Plot {
...
+        rect = e.svg.append('rect')
+	    .attr('stroke', 'red')
+            .attr('fill-opacity', 0.5);
+
+    e.svg
+        .on('mousemove.zoomRect', function() {
+          var m = d3.mouse(e);
```
maffoo: Here, too, I think it would be clearer to just pull out the separate components:

``` javascript
var [x, y] = d3.mouse(e);
```

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -252,4 +258,46 @@ export class Plot1D extends polymer.Base implements Plot {
...
+	  var m = d3.mouse(e);
+          m[0] = Math.max(0, Math.min(e.width, m[0]));
+          m[1] = Math.max(0, Math.min(e.height, m[1]));
+          if (m[0] !== origin[0] && m[1] !== origin[1]) {
+	    e.zoom.x(e.xScale.domain([origin[0], m[0]].map(e.xScale.invert).sort()))
+                .y(e.yScale.domain([origin[1], m[1]].map(e.yScale.invert).sort()));
+          }
+          rect.remove();
+          e.handleZoom();
+        }, true);
```
maffoo: what does `true` do here?

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -252,4 +258,46 @@ export class Plot1D extends polymer.Base implements Plot {
...
+
+  private zoomRectangle() {
+    var e = this,
+        origin = d3.mouse(e),
+        rect = e.svg.append('rect')
+	    .attr('stroke', 'red')
+            .attr('fill-opacity', 0.5);
+
+    e.svg
+        .on('mousemove.zoomRect', function() {
```
maffoo: In typescript, can use arrow functions instead:

``` javascript
.on('mousemove.zoomRect', () => {
 ...
});
```

One advantage is that arrow functions bind `this` with lexical scope, instead of dynamic scope as in plain javascript. This would obviate the need for the local alias `e` for `this` in the outer function. (This file generally has some bad examples of aliasing `this`, mostly for brevity. We should probably clean that up at some point...)

maffoo:
It would be nice to be able to change zooming modes to allow using the scrollwheel to zoom and drag to pan as it works currently, while also enabling the drag to zoom as implemented here. Maybe just having a simple button to toggle modes would suffice.

Also, I notice that on my machine the zoom rectangle is offset from the mouse cursor position by about 40 pixels or so. Perhaps a problem with the coordinate systems?

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -252,4 +258,46 @@ export class Plot1D extends polymer.Base implements Plot {
          .attr('d', this.line);
     }
   }
+  
+  private reset() {
```
maffoo: a docstring would be helpful

client-js/app/elements/labrad-plot1d.ts
```diff
@@ -252,4 +258,46 @@ export class Plot1D extends polymer.Base implements Plot {
...
+    this.xScale.domain([this.limits.xMin, this.limits.xMax]);
+    this.yScale.domain([this.limits.yMin, this.limits.yMax]);
+    this.xAxis.scale(this.xScale);
+    this.yAxis.scale(this.yScale);
+    this.zoom.x(this.xScale);
+    this.zoom.y(this.yScale);
+    this.handleZoom();
+  }
+
+  private zoomRectangle() {
```
maffoo: a docstring would be helpful

maffoo:
Two other quirks I've noticed: sometimes the zoom rectangles don't get removed and just remain in the DOM; haven't been able to track down exactly when this happens though. Also, when I drag to zoom the axis and tick label text gets highlighted like it's being selected; probably some event is being allowed to propagate where it shouldn't, or the default is not being cancelled.

client-js/app/elements/labrad-plot1d.ts
ccwa: Fixed. seems like there was a problem with my editor.

client-js/app/elements/labrad-plot1d.ts
ccwa: Done

client-js/app/elements/labrad-plot1d.ts
ccwa: Done

client-js/app/elements/labrad-plot1d.ts
ccwa: Done.

client-js/app/elements/labrad-plot1d.ts
ccwa: Done.

client-js/app/elements/labrad-plot1d.ts
ccwa: Thanks for the explanation.

DanielSank:
@ccwa thanks so much for working on this.

client-js/app/elements/labrad-plot1d.ts
ccwa: It ensures that the zoom is on the selected rectangular region.

maffoo:
LGTM

PR #106, opened 2015-11-22
Implement unit testing with Jasmine

joshm-3b6b58:
Sooooooo... I don't know how I elegant this is, but I have unit tests that are written in typescript that will compile and run in jasmine with gulp.

To use this branch, you have to install the dependencies

```
npm install jasmine
tsd install jasmine
npm install gulp-jasmine-browser
```

Then you'll have to do a 

```
gulp clean
```

since I had to re-structure the .tmp folder for everything to play nice. To run the tests type:

```
gulp jasmine
```

Which will start a server listening on localhost:8888 and you will need to point your browser there to see the results of your tests. The tests have to be run in a browser or else I get a [`ReferenceError: window is not defined` error](http://stackoverflow.com/questions/27224919/gulp-jasmine-window-is-not-defined-error)

First of all let me know how it runs. I kind of wanted to just get something out there for people to see so I didn't get too deep if there are any misgivings.

But I think it checks all the boxes, we can write all our tests in typescript (and use help from a compiler), it sorta integrates with our build, using the same ES6 modules, etc and it uses gulp without adding too much foreign machinery. 

I'm not sure if the test I wrote for `Observable` is total nonsense. I just messed with it untill I got something to compile and pass :0


client-js/lib/helloWorld.js
```diff
@@ -0,0 +1,9 @@
+function Agent() {}
```
joshm-3b6b58: None of this lib stuff should be here, I'll delete it now

client-js/app/test/index.html
```diff
@@ -1,22 +0,0 @@
-<!doctype html>
```
joshm-3b6b58, line 1: I don't know how this got into master, but I don't think we want to use web-component tester anyhow...

maffoo, line 1: This came in with the polymer example project. I'm glad you've deleted it; if we ever need this it's no problem to get it back.

maffoo:
We should add `jasmine` and `gulp-jasmine-browser` to `devDependencies` in `package.json`. That way, new users will get all the needed libraries simply by doing the standard `npm install`.

In addition, I've been checking the typescript `*.d.ts` files in to the repo instead of making users figure out how to use tsd; it's fine to use to fetch the right definition files, but then you can just commit the files that are new so other users will get them without having to get tsd working. (Ideally, we could just use tsd, but I've found that the definition files sometimes require manual tweaks, especially because we're using ES6 and most of the definition files are tailored for ES5 which can lead to the occasional incompatibility.)

client-js/gulpfile.js
```diff
@@ -18,6 +18,9 @@ var url = require('url');
 var glob = require('glob');
 var merge = require('merge2');
 var exec = require('child_process').exec;
+var gulp = require('gulp');
```
maffoo: this import already exists (line 4)

maffoo:
Can you say a few words about how you settled on jasmine? I haven't looked much at js testing libraries, so I'm just curious how you picked this one.

client-js/test/spec/tsTestSpec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  function testFunc(message: string) {
+    console.log("testFunc called");
+    return message;
+  }
+
+
+  it("Add Callback", function() {
+    console.log("In callback test");
+    expect(observable.add(testFunc("something"))).toEqual("something");
```
maffoo: This should not compile. The code is passing a string to `observable.add`, but it wants a callback of type `any => void`, that is, a callback function. If the compiler is accepting this then there might be a problem with the way the compiler is invoked that it is not seeing the type definitions.

joshm-3b6b58:
@maffoo here's the thought process for picking Jasmine. All the typescript testing suites are tied to Visual Studio or NuGet (windows package manager). There are infinity js testing frameworks and JSUnit was the most Google friendly, Jasmine seems like the most developped of the JSUnit derivatives.

client-js/test/spec/tsTestSpec.ts
joshm-3b6b58: I will look into this. I'm pretty sure the compiler is invoked generally, becuase I was able to do things wrong and get compiler errors... Could the problem be how observable is instanced?

client-js/gulpfile.js
joshm-3b6b58: Oh, duh, thanks.

client-js/test/spec/tsTestSpec.ts
joshm-3b6b58: @maffoo Actually `add()` is a thing in the [jasmine typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/jasmine/jasmine.d.ts#L264), could that be the problem?

maffoo: but `observable` is declared above as being of type `Observable<any>`, so this should be typechecked against that class's [add](https://github.com/labrad/scalabrad-web/blob/master/client-js/app/scripts/observable.ts#L23) method.

joshm-3b6b58: So, I'm not sure what's going on, but the compiler is doing some type checking, if I change: 

```
observable.add(testFunc("something"))
```

to 

```
observable.add(testFunc(4))
```

I get the compiler error: `test/spec/tsTestSpec.ts(39,36): error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.`

Now what I don't know if it's doing the type checking properly for the imports? Maybe something is mis-configured.

joshm-3b6b58: Even if I copy the code from `observable.ts` into `tsTestSpec.ts` and comment out the import, I still don't get a compiler error? I know the compiler is checking type definitions according to the comment above. I know it's seeing the right code, because I'm just pasting it in. What else am I missing?

maffoo: I think I see the problem. You need to actually declare the type of the `observable` variable. Right now it is implicitly typed as `any`. Instead, you should do:

``` javascript
var observable: Observable<any>;

beforeEach(() => {
  observable = new Observable<any>();
});

...
```

I don't know what type the compiler was inferring for this, but that should help.

Of course, then you'll have to the test itself. Calling `.add` itself does not return anything. You'll want to actually call the observable and check that the callback was invoked.

joshm-3b6b58: Right, I was pretty sure the actual test was bogus, but I wanted to get another set of eyes on the overall structure before diving too deep. I mainly wanted to see that I could compile and run jasmine functions over the existing codebase. I'll rewrite the test.

client-js/test/spec/tsTestSpec.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+});
+
+/*
+* Test to create new observable and add callback
+* This is mostly to see if the plumbing for the
+* tests interfaces with the scripts in /app
+*/
+
+describe("observable", function() {
```
joshm-3b6b58: @maffoo can you check out this test? I had to use some extra machinery to play nice with the async call, but it passes and I think it tests the correct thing.

joshm-3b6b58:
@maffoo does this all look OK? Do you want me to do anything else for the PR? We'd talked offline about getting the commandline tests runner to work with `window` in `Observable`. Adding automatically compiling tests on changes to the files. Do you want that in this PR?

client-js/test/spec/tsTestSpec.ts
```diff
@@ -0,0 +1,61 @@
...
+* it complained about asyncronous things.
+*/
+//  it("call Callback", function() {
+//    var p = new Promise((resolve, reject) => resolve("variable"));
+//    p.then((n) => observable.call(n))
+//    .then((n) => expect(bar).toEqual(n));//.then(expect(bar).toEqual('variable'));
+//    console.log("In callback test", bar);
+//    //expect(bar).toEqual('variable');  
+//    
+//  });
```
maffoo: Remove commented code

client-js/test/spec/tsTestSpec.ts
maffoo: looks ok; personally I prefer to have things close together, rather than doing setup in a function like `beforeEach`, which spreads out the code that you have to read to see what a test is doing. Maybe if you have lots of specs that share common setup this could make sense, but even then, I would prefer to have a function that you call explicitly within the specs instead of relying on side effects from `beforeEach`. For example:

``` javascript
function setup(): Observable<any> {
  var observable = new Observable<any>();
  // do a bunch of setup
  return observable;
}

it("does something", function(done) {
  var observable = setup();
  ...
}
```

For the particular case here, I would say it's better to do the work like calling the observable in the spec functions, not the setup function. The setup should be reserved for instantiating objects, etc., not performing the actual logic that you want to test.

``` javascript
it("calls added callbacks asynchronously", function(done) {
  var observable = new Observable<string>();
  observable.add((msg) => {
    expect(msg).toEqual("foo");
    done();
  });
  observable.call("foo");
});
```

Anyway, we'll have to see how our usage evolves. For now, what you have is fine.

maffoo:
Need to add `gulp-jasmine-browser` to `package.json` so users will get it when they run npm install. Probably want to add jasmine itself as well (I think it gets pulled in as a dependency of the gulp stuff, but it's better to be explicit).

I think fixing `Observable` to work outside the browser should be as simple as removing `window.` and just calling `setTimeout` as a global function. You might at least try that, and if it works then it would be great to make the tests work on the command line directly. Having browser tests as well might be nice, but we don't necessarily need them for now. I'd say don't work about automatically rerunning the tests on file changes; that can definitely come later.

One final thing I wanted to ask about is the file organization. I'd suggest moving `jasmine.json` to `test/jasmine.json` and getting rid of the top-level `spec/` folder. In addition, there are currently two folders `test/spec/` which has test code and `test/test/` which has library code used by the tests. I'm not sure the best way to organize this, but having these two folders is a bit confusing. Fine to leave as is for now I guess, since I don't have a good concrete suggestion. I would however suggest renaming the individual files since in the main app code we've been using `words-with-hyphens` instead of `camelCase` for file names.

client-js/test/spec/tsTestSpec.ts
joshm-3b6b58: So I can't put the `expect().toEqual()` in an actual callback or Jasmine complains. We'll have to play with this a bit...

joshm-3b6b58:
So removing `window` allows me to run the tests from the command line. However I don't know if this breaks anything else. The registry browser still seems to work as usual. 

I'll get the command line tests auto-running and fix the other comments.

client-js/test/spec/tsTestSpec.ts
maffoo: That test ran fine for me with expect in a callback; what is the error that you're seeing?

joshm-3b6b58: I gues it's working fine now. Maybe I typed something in wrong last time. Alrighty.

joshm-3b6b58:
So, if I move `jasmine.json` from it's default location anybody using this repo will have to manually run [`jasmine JASMINE_CONFIG_PATH=relative/path/to/your/jasmine.json`](http://jasmine.github.io/2.3/node.html#section-CLI_Options) to point to the new path, since this is hard coded in some strange way I don't understand. There's no higher-level configuration file I can just edit and store in the repo, I don't think?

joshm-3b6b58:
@maffoo when you get a chance can you see if I made all the necessary changes?

client-js/gulpfile.js
```diff
@@ -52,14 +54,53 @@ gulp.task('compile-ts', function () {
...
     tsResult.js.pipe(sourcemaps.write('.')).pipe(gulp.dest('.tmp/'))
   ]);
 });
 
-// create a single executable js file using systemjs-builder
+gulp.task('jasmine-browser', ['bundle-test'], function() {
+  return gulp.src(['.tmp/testing/specBundle.js'])
+    .pipe(jasmineBrowser.specRunner())
+    .pipe(jasmineBrowser.server({port: 8888}));
+});
```
maffoo, line 84: nit: blank line after this

client-js/gulpfile.js
```diff
@@ -255,6 +297,21 @@ gulp.task('serve', ['bundle', 'insert-dev-config', 'styles', 'elements', 'images
   gulp.watch(['app/{scripts,elements}/**/*.ts'], ['bundle', reload]);
   gulp.watch(['app/images/**/*'], reload);
 });
+//
```
maffoo: nit: blank line instead of comment

client-js/test/spec/ts-test-spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+* Test to create new observable and add callback
+* This is mostly to see if the plumbing for the
+* tests interfaces with the scripts in /app
+*/
+
+describe("observable test", function() {
+  var observable: Observable<any>;
+  var lifetime = new Lifetime();
```
maffoo: looks like `lifetime` is not used. Also, since `observable` is declared as a local inside the test function below, that's not needed here either.

client-js/test/spec/ts-test-spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  it("calls added callbacks asynchronously", function(done) {
+  var observable = new Observable<string>();
+  observable.add((msg) => {
+    expect(msg).toEqual("foo");
+    done();
+  });
+  observable.call("foo");
+});
```
maffoo: nit: fix indentation.

maffoo:
A couple of nits about formatting, otherwise LGTM.

PR #107, opened 2015-11-30
Allow ENTER in login dialog

### LINKED ISSUE #95: Pressing ENTER after typing the password doesn't work
You have to click the `login` button. Hitting enter sends you back to where you started.

maffoo:
Fixes #95

### DIFF
--- a/client-js/app/elements/labrad-app.html
+++ b/client-js/app/elements/labrad-app.html
@@ -116,14 +116,14 @@
 
     <paper-dialog id="loginDialog" modal>
       <h1>Login to Labrad</h1>
-      <form>
+      <form id="loginForm">
         <paper-input id="passwordInput" label="password" always-float-label type="password"></paper-input>
         <paper-checkbox id="rememberPassword">Remember password</paper-checkbox>
       </form>
       <div>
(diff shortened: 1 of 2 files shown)


DanielSank:
Hey @maffoo, I'm thinking how best to test stuff like this. Can we set up a server to run the site so that folks to connect and see if it works on their system (i.e. various browsers)?

joshm-3b6b58:
I think we should place a heavy emphasis one just using Chrome for this...

DanielSank:
Tested to work. Lemme look at the code real quick.

DanielSank:
Hmm, I don't think I can review this. What does `app.$` mean?

maffoo:
The `.$` attribute is an object whose attributes refer to DOM nodes with explicit id attributes inside a custom element. This is something that polymer creates on elements defined through polymer. In this case, `app` is a `labrad-app` element, so `app.$.loginForm` selects the form with that id defined in `labrad-app.html`, etc.

DanielSank:
Definitely punting to @joshmutus on this one. I just don't know enough.

joshm-3b6b58:
LGTM

PR #111, opened 2015-12-09
Add PromiseInput class which serves as input to Promise chain.

maffoo:
In several places, we were pulling out the resolve and reject callbacks from a newly-created promise and storing them to be called later when we have a value or error with which to complete the promise. This adds a `PromiseInput` class to formalize this pattern.

I'm not too sure about the name, so would be happy if people have other suggestions. I thought about `Completable` which conveys the notion that once we resolve or reject, the operation is complete and we can't resolve or reject again. But it's also nice to have the name reflect that this class is closely related to `Promise`. As always, names are hard.

### DIFF
--- a/client-js/app/scripts/obligation.ts
+++ b/client-js/app/scripts/obligation.ts
@@ -1,0 +1,38 @@
+/**
+ * Obligation is the input side complement of Promise; while Promise provides a
+ * way to create chains of asynchronous callbacks, Obligation is the input that
+ * lets us send a value or error in to the beginning of the callback chain.
+ *
+ * The name is meant to evoke a scenario in which Alice gives to Bob a promise
+ * that she will deliver a value. We then say that Alice has an obligation to
+ * either deliver a value as promised or else explain why she failed to do so.
+ * An Obligation contains two methods `resolve` and `reject` which correspond
+ * respectively to these two cases, delivering a value or failing with an error.
+ *
+ * In the standard ES6 Promise API, resolve and reject functions are only
+ * accessible as arguments passed in to the executor function which we pass to
+ * the Promise constructor. However, in many cases we'd like to call these
+ * functions from outside the constructor; Obligation makes this possible. This
+ * is especially useful in situations where we need to connect callback-based
+ * code with Promise-based code.
+ *
+ * Compare the scala.concurrent library, which properly separates the input and
+ * output sides of an asynchronous callback chain into two classes:
+ * scala.concurrent.Promise is the input side (equivalent to Obligation here)
+ * and scala.concurrent.Future is the output side (equivalent to Promise here).
+ */
+export interface Obligation<R> {
+  resolve: (value?: R | Promise<R>) => void;
+  reject: (error: any) => void
+}
+
+/**
+ * Create an Obligation/Promise pair with the given result type.
+ */
+export function obligate<R>(): { obligation: Obligation<R>, promise: Promise<R> } {
+  var obligation: Obligation<R>;
+  var promise = new Promise<R>((resolve, reject) => {
+    obligation = { resolve: resolve, reject: reject };
+  });
+  return { obligation: obligation, promise: promise };
+}
--- a/client-js/app/scripts/async-queue.ts
+++ b/client-js/app/scripts/async-queue.ts
@@ -1,3 +1,5 @@
+import {Obligation, obligate} from './obligation';
+
 /**
  * An asynchronous queue of items of some type. Taking from the queue returns
  * a Promise that will fire when an item is ready (this Promise may already
@@ -5,14 +7,14 @@
  */
 export class AsyncQueue<A> {
   private items: Array<A> = [];
-  private promises: Array<{resolve: (A) => void; reject: (any) => void}> = [];
+  private waiters: Array<Obligation<A>> = [];
 
   /**
    * Add an item to the queue.
    */
   offer(a: A): void {
-    if (this.promises.length > 0) {
-      this.promises.shift().resolve(a);
+    if (this.waiters.length > 0) {
+      this.waiters.shift().resolve(a);
     } else {
       this.items.push(a);
     }
@@ -26,10 +28,9 @@ export class AsyncQueue<A> {
     if (this.items.length > 0) {
       return Promise.resolve(this.items.shift());
     } else {
-      var promises = this.promises;
-      return new Promise((resolve, reject) => {
-        promises.push({resolve: resolve, reject: reject});
-      });
+      var { obligation, promise } = obligate<A>();
+      this.waiters.push(obligation);
+      return promise;
     }
   }
 
@@ -38,8 +39,8 @@ export class AsyncQueue<A> {
    * containing the given reason.
    */
   close(reason: string = 'queue closed'): void {
-    while (this.promises.length > 0) {
-      this.promises.shift().reject(new Error(reason));
+    while (this.waiters.length > 0) {
+      this.waiters.shift().reject(new Error(reason));
     }
   }
 }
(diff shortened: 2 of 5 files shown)


DanielSank:
What's the need for the `_*` methods?

joshm-3b6b58:
@maffoo it would be nice to start adding unit tests for every new piece. I don't know if that's necessary for this PR though since I merged the unit testing branch probably after you started this.

maffoo:
@DanielSank, `_resolve` and `_reject` are not methods, they are class members that happen to have function type. The methods `resolve` and `reject` just call the functions contained in those variables. I did it this way because public members of a class are mutable and I want users of this class to _only_ be able to call but not modify these functions, which methods do.

@joshmutus, good call; I'll add a couple of tests. If nothing else, they're helpful for documentation :)

maffoo:
After consultation with @DanielSank, renamed `PromiseInput` to `Obligation`. Also added basic tests for this class. @DanielSank and @joshmutus, please have another look when you can.

DanielSank:
@joshmutus if you want to review this together I'm game. It took me more than an hour (and more importantly a discussion with @maffoo) before I understood enough about the (terribly documented) ES6 Promise API such that I could even comprehend this code.

joshm-3b6b58:
@DanielSank I'm in the cleanroom tomorrow, we can look at it together if it can wait until Friday.

client-js/app/scripts/obligation.ts
```diff
@@ -0,0 +1,48 @@
...
+ * traditional callback-based code with Promise-based code.
+ *
+ * Compare the scala.concurrent library, which properly separates the input and
+ * output sides of an asynchronous callback chain into two classes:
+ * scala.concurrent.Promise is the input side (equivalent to Obligation here)
+ * and scala.concurrent.Future is the output side (equivalent to Promise here).
+ */
+export class Obligation<R> {
+
+  public promise: Promise<R>;
```
DanielSank: Why public? It seems to me the only reason to make this public is so that an asynchronous API can do something like this:

``` javascript
somethingAsync(): Promise<A>() {
  var obligation = new Obligation<A>();
  myObligations.push(obligation);
  return obligation.promise;
}
```

If the only point of making the `promise` field public is to return it to the piece of code to which we're making the promise, I'd rather just use a Promise/Obligation factory and keep the `promise` field of `Obligation` private.

client-js/app/scripts/obligation.ts
```diff
@@ -0,0 +1,48 @@
...
+      this._resolve = resolve;
+      this._reject = reject;
+    });
+  }
+
+  /**
+   * Resolve our Promise with the given value.
+   */
+  resolve(value?: R | Promise<R>): void {
+    this._resolve(value);
```
DanielSank: ~~Why not~~

``` javascript
this.promise.resolve(value);
```

~~which would also let us remove the mind-boggling constructor?~~

I see. Man the ES6 Promise API is really weird.

client-js/app/scripts/obligation.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Obligation is the input side complement of Promise; if one party is promised
+ * a value, then of necessity another party must be obliged to provide the
```
DanielSank: I think "by necessity" is more standard. It might be more compact and clear to reword as

"If one party is promised a value, then the party who made the promise is obliged to provide that value or explain why they failed to do so".

maffoo: "of necessity" = 26.5 M google results; "by necessity" = 18.8 M google results. Therefore "of necessity" wins :)

But yes, I like your wording better anyway.

client-js/app/scripts/obligation.ts
```diff
@@ -0,0 +1,48 @@
...
+ * value or explain why they failed to do so. Obligation has two methods
+ * `resolve` and `reject` which can be used to send a value or an error into a
+ * chain of asynchronous callbacks. Promise is the output side of the chain,
+ * providing access to the asynchronous values by adding callback functions.
+ *
+ * In the traditional ES6 Promise API, resolve and reject functions are only
+ * available as arguments passed in to the executor function we pass to the
+ * Promise constructor. But in many situations we need to store the callbacks
+ * for later use or access them outside the constructor, which Obligation makes
+ * possible. This is especially useful in situations where we need to interface
```
DanielSank: The `resolve` and `reject` methods of the `Promise`living inside each `Obligation` are private. That being the case, how does `Obligation` make them accessible?

client-js/app/scripts/obligation.ts
DanielSank: Eh, that's a 60% / 40% split. I'd say that's easily within the range where we can pick whichever we like better. Besides, most of those hits are probably middle-schoolers' Facebook posts :P

client-js/app/scripts/obligation.ts
maffoo: I don't understand what you are asking. `Promise` does not have methods called `resolve` and `reject`. The `Promise` _constructor_ takes a function which gets called with two functions to resolve or reject the promise, respectively. We store those functions in private members of `Obligation` but make them accessible by letting you call them through the public `resolve` and `reject` methods on `Obligation` itself.

DanielSank: Can you just change "access them" to "call them"?

DanielSank: Also, you say `Promise` has no `resolve` and `reject` methods, but then [Mozilla says otherwise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). Is there a reference you'd recommend that's not the completely unreadable official ES6 spec?

maffoo: Those are not methods on a Promise instance, those are static methods that construct pre-resolved or pre-failed Promise instances. (They "lift a constant value into the Promise monad", if you like, though I'm guessing you don't.)

DanielSank: > Those are not methods on a Promise instance, those are static methods that construct pre-resolved or pre-failed Promise instances.

Ah.

> They "lift a constant value into the Promise monad", if you like, though I'm guessing you don't.

Pretty sure last time I asked you what a monad was you said "everything".

DanielSank: Anyway, I still think changing "access them" to "call them" would be helpful. Saying "access" makes me think I can get references to the `resolve` and `reject` functions and do whatever I want with them, which is not true. All I can do is call them through the Obligation's methods.

client-js/app/scripts/obligation.ts
maffoo: I tried converting `Obligation` to be an interface and using a factory function to produce `Obligation/Promise` pairs instead of having an `Obligation` hold a reference to the `Promise`: https://github.com/labrad/scalabrad-web/commit/2b79ab5ddeabc3e9afe302f7d032a981eea21a86

It's very javascripty (using an interface and structural types instead of classes), I have to say I don't like it as well. For me, just creating one `Obligation` object and having it hold a reference to the `Promise` is simpler. But take a look at that commit and let me know what you think.

DanielSank: I have read and comprehended the code in https://github.com/labrad/scalabrad-web/commit/2b79ab5ddeabc3e9afe302f7d032a981eea21a86, and I like it much better. The original way of having the `Promise` live inside an `Obligation` is asymmetric for no reason. In the new style where we say

``` javascript
var {obligation, promise} = obligate<A>();
```

the symmetry makes sense; `obligate` and `promise` are on equal footing.

DanielSank:
Going with the "symmetric" version (i.e. the `obligate` function) LGTM.

client-js/app/scripts/obligation.ts
DanielSank: I should point out that I know almost nothing about js and so am totally unaware of any implications that come from using an interface and "structural types" (don't even know what that means).

PR #113, opened 2015-12-09
Add main test entry point module which can import other tests

maffoo:
This way we can define tests in multiple files and just add a line to the test main module to import each file.


joshm-3b6b58:
LGTM

PR #114, opened 2015-12-10
Update typescript to take advantage of async/await

maffoo:
First, upgrade typescript to [version 1.7.3](http://blogs.msdn.com/b/typescript/archive/2015/11/30/announcing-typescript-1-7.aspx), which among other things adds support for ES6 async/await functions, which are to ES6 Promise what inlineCallbacks and generators are to twisted Deferreds. Second, modify Promise-using code to use async/await where applicable. I think this simplifies the code, allowing us to use standard control flow and things like try/catch/finally rather than having to chain .then and .catch calls on Promises.

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -105,38 +105,28 @@ window.addEventListener('WebComponentsReady', () => {
 
   var activity: Activity = null;
 
-  function activate(newActivity: Activity): void {
-    var p: Promise<void> = Promise.resolve(null);
-    if (activity) {
-      p = activity.stop();
-    }
-    p.then(
-      (success) => {
-        newActivity.start().then(
-          (state) => {
-            app.route = state.route;
-            if (state.breadcrumbs) {
-              app.hasBreadcrumbs = true;
-              app.breadcrumbs = state.breadcrumbs;
-            } else {
-              app.hasBreadcrumbs = false;
-            }
-            var content = app.$.content;
-            while (content.firstChild) {
-              content.removeChild(content.firstChild);
-            }
-            content.appendChild(state.elem);
-            activity = newActivity;
-          },
-          (error) => {
-            console.log('error while starting activity', error);
-          }
-        )
-      },
-      (error) => {
-        console.log('error while shutting down activity', error);
+  async function activate(newActivity: Activity): Promise<void> {
+    try {
+      if (activity) {
+        await activity.stop();
       }
-    )
+      var state = await newActivity.start();
+      app.route = state.route;
+      if (state.breadcrumbs) {
+        app.hasBreadcrumbs = true;
+        app.breadcrumbs = state.breadcrumbs;
+      } else {
+        app.hasBreadcrumbs = false;
+      }
+      var content = app.$.content;
+      while (content.firstChild) {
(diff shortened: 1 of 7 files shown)


client-js/app/scripts/app.ts
```diff
@@ -105,38 +105,28 @@ window.addEventListener('WebComponentsReady', () => {
 
   var activity: Activity = null;
 
-  function activate(newActivity: Activity): void {
-    var p: Promise<void> = Promise.resolve(null);
-    if (activity) {
-      p = activity.stop();
-    }
```
DanielSank, line 112: I'm curious about how this original version works. We make a resolved `Promise`, and then if `activity` tests true we set that promise to the result of `activity.stop()`. Then we start chaining callbacks. I guess the idea is that if `activity` tests false then we want to start chaining callbacks onto a `Promise` which completes immediately (i.e. is already resolved), but if `activity` is true then we have to chain after the `stop()` completes.

Do I have it right?

client-js/app/scripts/app.ts
```diff
@@ -351,34 +334,30 @@ window.addEventListener('WebComponentsReady', () => {
       this.plot.addData(data);
     }
 
-    stop(): Promise<void> {
+    async stop(): Promise<void> {
```
DanielSank, line 337: Is it correct to just think of `async` as "Make this function return a resolved `Promise` with value equal to whatever this function returns"? Sort of like something I'd normally expect to be a decorator, but here it's a keyword instead.

DanielSank:
I read line by line and I think it's all good. However, I'm a pretty low-res js scanner just due to inexperience.

client-js/app/scripts/app.ts
maffoo, line 112: That's correct. This kind of conditional logic is tricky when chaining promises because you have to play tricks to ensure that all code branches get chained together properly. async/await methods just let you use all the conventional control structures and all the branches get tied together under the hood.

client-js/app/scripts/app.ts
maffoo, line 337: A slight tweak on what you wrote: "Make this function return a `Promise` _which will eventually resolve_ to whatever this function (eventually) returns, or will be rejected if the function raises an exception." And of course, `async` also plumbs things together so that you can use `await` to deal with other asynchronous operation inside the function without turning the code inside out to write callbacks.

`async`/`await` here is very similar to twisted's `@inlineCallbacks`/`yield` in python. In fact, under the hood async functions are implemented with ES6 generators, so the analogy is really very close. One difference is that `async` functions can use plain old `return` to return values, whereas `@inlineCallbacks` functions in python cannot (at least until python 3).

maffoo:
@DanielSank, thanks for looking this over; if I deploy it on matrix-reloaded so you can try it out, would you feel comfortable LGTMing?

DanielSank:
Ehh, human testing seems kinda lame. If you can mess around and it works let's just call it good enough, since we don't have any tests set up for the actual application.

DanielSank:
I mean, I'll click around the site if you want.

DanielSank:
Clicked on stuff. Things happened.

LGTM :-|

PR #115, opened 2015-12-10
Add live view for grapher directories

maffoo:
There are still some ui rough edges here, but this adds a basic live view that opens new plots in a given directory.

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -13,6 +13,7 @@ import * as rpc from "./rpc";
 
 import {LabradApp} from "../elements/labrad-app";
 import {LabradGrapher} from "../elements/labrad-grapher";
+import {LabradGrapherLive, LabeledPlot} from "../elements/labrad-grapher-live";
 import {LabradManager} from "../elements/labrad-manager";
 import {LabradNodes, LabradInstanceController, LabradNodeController} from "../elements/labrad-nodes";
 import {LabradRegistry} from "../elements/labrad-registry";
@@ -64,6 +65,7 @@ window.addEventListener('WebComponentsReady', () => {
   // register our custom elements with polymer
   LabradApp.register();
   LabradGrapher.register();
+  LabradGrapherLive.register();
   LabradManager.register();
   LabradRegistry.register();
   LabradNodes.register();
@@ -72,6 +74,7 @@ window.addEventListener('WebComponentsReady', () => {
   LabradServer.register();
   Plot1D.register();
   Plot2D.register();
+  LabeledPlot.register();
 
   var body = document.querySelector('body');
   body.removeAttribute('unresolved');
@@ -115,6 +118,7 @@ window.addEventListener('WebComponentsReady', () => {
       if (state.breadcrumbs) {
         app.hasBreadcrumbs = true;
         app.breadcrumbs = state.breadcrumbs;
+        app.breadcrumbExtras = state.breadcrumbExtras;
       } else {
         app.hasBreadcrumbs = false;
       }
@@ -207,6 +211,10 @@ window.addEventListener('WebComponentsReady', () => {
           url: grapherUrl(this.path.slice(0, i))
         });
       }
+      var breadcrumbExtras = [
+        { name: 'dir view', isLink: false, url: '' },
+        { name: 'live view', isLink: true, url: grapherUrl(this.path) + '?live' }
+      ];
 
       this.elem = <LabradGrapher> LabradGrapher.create();
       this.elem.path = this.path;
@@ -216,7 +224,8 @@ window.addEventListener('WebComponentsReady', () => {
       return {
         elem: this.elem,
         route: 'grapher',
-        breadcrumbs: breadcrumbs
+        breadcrumbs: breadcrumbs,
+        breadcrumbExtras: breadcrumbExtras
       };
     }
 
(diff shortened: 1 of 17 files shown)


joshm-3b6b58:
Sorry I haven't had a chance to look at this yet... I'll try tomorrow.

joshm-3b6b58:
Just changed to this branch, when I do `gulp compile-ts` I get a compiler error:

```
app/scripts/app.ts(277,21): error TS1109: Expression expected.
```

Looking at the code, it looks fine? Do I have to upgrade my typescript first?

upgraded typescript, but that didn't fix anything. I ran `gulp clean` first. Here's the total compiler output:

```
mutus@mutus:~/programming/scalabrad-web/client-js$ gulp serve
[11:45:45] Using gulpfile ~/programming/scalabrad-web/client-js/gulpfile.js
[11:45:45] Starting 'compile-ts'...
[11:45:45] Starting 'insert-dev-config'...
[11:45:45] Starting 'styles'...
[11:45:46] Starting 'elements'...
[11:45:46] Starting 'images'...
[11:45:46] 'elements' all files 0 B
[11:45:46] Finished 'elements' after 331 ms
[11:45:46] 'insert-dev-config' all files 1.91 kB
[11:45:46] Finished 'insert-dev-config' after 799 ms
[11:45:46] 'styles' all files 370 B
[11:45:46] Finished 'styles' after 843 ms
[11:45:46] 'images' all files 24.13 kB
[11:45:46] Finished 'images' after 596 ms
app/scripts/app.ts(277,21): error TS1109: Expression expected.
[11:45:48] TypeScript: 1 syntax error
[11:45:48] TypeScript: emit succeeded (with errors)
[11:45:49] Finished 'compile-ts' after 3.26 s
[11:45:49] Starting 'bundle'...
     Building the single-file sfx bundle for app/scripts/app...

err  Error loading "app/scripts/app" at file:/usr/local/google/home/mutus/programming/scalabrad-web/client-js/.tmp/app/scripts/app.js
     file:/usr/local/google/home/mutus/programming/scalabrad-web/client-js/.tmp/app/scripts/app.js:242:31: Unexpected token ;


[11:45:50] 'bundle' errored after 1.13 s
[11:45:50] Error: Command failed: /bin/sh -c node_modules/.bin/jspm bundle-sfx app/scripts/app .tmp/scripts/bundle.js --skip-source-maps

    at ChildProcess.exithandler (child_process.js:751:12)
    at ChildProcess.emit (events.js:110:17)
    at maybeClose (child_process.js:1015:16)
    at Process.ChildProcess._handle.onexit (child_process.js:1087:5)
```

maffoo:
Well, that's broken. Line 277 should say 'await' instead of 'return'. I'll push a fix.

maffoo:
fixed.

joshm-3b6b58:
Very cool. I'm able to run it now. Live update works, looks nice! 

I get a gob warnings about async from the compiler. That's expected, right?

I'll continue to work through this.

maffoo:
There shouldn't be any warnings about async. Do you have typescript 1.7? Might just want to `npm install` and `jspm install` to be sure. (Also, you can do `npm install -f typescript` to update the global typescript package, though I don't think that's strictly necessary.)

joshm-3b6b58:
```
$ tsc -v
message TS6029: Version 1.7.5
```

joshm-3b6b58:
Stray UI/UX nits: The ? delimeter between the current folder and 'live' is confusing. It looks kind of like a formatting error. I would suggest distinguishing the live button from the path text by aligning it to the right hand of the nav bar (next to the logout) and making it a distinct button with an [icon](https://design.google.com/icons/#ic_live_tv) or [icon](https://design.google.com/icons/#ic_play_circle_outline).

~~Also as it stands I can't navigate to the live view when I have a graph presently open.~~ 

Oh I get it, it just goes live in the selected directory, of course. OK now I understand you're trying to indicate this link is for the live view in the current folder. 

Perhaps locating the link in the place it currently is but making it more of a distinct button would be more intuitive.

client-js/app/scripts/app.ts
```diff
@@ -247,17 +257,102 @@ window.addEventListener('WebComponentsReady', () => {
     }
   }
 
+  class DatavaultLiveActivity implements Activity {
```
joshm-3b6b58, line 259: How come there's a DatavaultLiveActivity in addition to the DatavaultActivity? 

Plots of datasets opened by themselves, without navigating to `?live` update when new data is added live.  

What's the distinction for these separate activities?

Is it so live can automatically show the latest dataset?

maffoo, line 259: I'm thinking of activities being in 1-1 correspondence with URLs or locations in the app. The DatavaultActivity shows the contents of a particular directory; the DatavaultLiveActivity shows a live-updating view of the most recent datasets in a directory. Since these do substantially different things, I made them separate activity classes. I'd be happy to change the names if you think there's a way to more clearly indicate what is going on. Also, I'll add some docstrings since that also helps to communicate the intent.

maffoo:
@joshmutus, I cleaned up the dir view/live view links to try to make it more clear that these are alternate views of the current directory. Let me know what you think. It's a bit awkward because the underline means a link to the _other_ location, but it looks like an emphasis which might indicate the current location. There are various ways this could be addressed, but I think to really make this work nice would require restructuring element classes so that, for example, we could put toggle buttons in; right now I've just hacked the breadcrumbs class in an admittedly janky way.

server/src/main/scala/org/labrad/browser/VaultApi.scala
```diff
@@ -106,6 +107,10 @@ class VaultApi(cxn: LabradConnection, client: VaultClientApi)(implicit ec: Execu
 
   import VaultApi._
 
+  private val DATA_AVAILABLE = 4444
```
joshm-3b6b58: What's the deal with these numeric sequences? What does DATA_AVAILABLE have to do with the number 4444?

server/src/test/scripts/data_demo.py
```diff
@@ -64,17 +64,18 @@ def main(dv):
 
     raw_input('Starting demo {}. Press [Enter] to continue. '.format(path))
 
-    print 'demo_1d_simple'
-    demo_1d_simple(dv)
+    while True:
```
joshm-3b6b58, line 67: Thanks for adding these scripts it really make testing this a breeze.

client-js/app/scripts/app.ts
joshm-3b6b58, line 259: I understand now why there are two of these, but there seems to be a lot of code duplicated between those two functions. It's beyond this PR, but I think it would be nice to have a lower (higher? I forget the convention) level function with the common skeleton called `DataVaultActivity` that `DataVaultLiveActivity` and something like `DataVaultStaticActivity` inherit from it.

joshm-3b6b58:
A couple questions, but everything works. LGTM

client-js/app/scripts/app.ts
maffoo, line 259: I agree there's room here to refactor and clean this up. Happy to take a look at that in a subsequent commit.

server/src/main/scala/org/labrad/browser/VaultApi.scala
maffoo: These are message IDs that we are free to chose arbitrarily; we tell the server to send each message to a particular ID and then use that to know how to handle the messages we get. I will add some comments to make this a bit more clear.

PR #117, opened 2015-12-22
Log summary of JSON-RPC messages, rather than full message

maffoo:
When using grapher the logs can be very verbose, and the request body can contain sensitive information. Reduce logging to fix both of these problems.


DanielSank:
LGTM

PR #120, opened 2015-12-29
Always do numeric comparison when computing new zoom limits

### LINKED ISSUE #119: Grapher: 1D zooming can flip data upside down
When zooming into a 1D plot by drawing the zoom rectangle from the top going to the bottom, the resulting plot will be flipped upside down relative to the original plot. Hence, a dip will turn into a peak. Using version 1.0.0-M3.

maffoo:
Previously, we used Array.sort, which actually converts numbers to strings before sorting.

Fixes #119

@jwenner

### DIFF
--- a/client-js/app/elements/labrad-plot1d.ts
+++ b/client-js/app/elements/labrad-plot1d.ts
@@ -104,7 +104,7 @@ export class Plot1D extends polymer.Base implements Plot {
     p.zoom = d3.behavior.zoom()
             .x(p.xScale)
             .y(p.yScale)
-            .on('zoom', p.handleZoom.bind(p));
+            .on('zoom', () => this.handleZoom());
 
     // plot area
     p.svg = d3.select(area)
@@ -115,7 +115,7 @@ export class Plot1D extends polymer.Base implements Plot {
             .attr('transform', `translate(${p.margin.left}, ${p.margin.top})`)
             .call(p.zoom)
             .append('g')
-            .on('mousedown', p.zoomRectangle.bind(this));
+            .on('mousedown', () => this.zoomRectangle());
 
     // background rectangle
     p.svg.append('rect')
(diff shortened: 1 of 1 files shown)


client-js/app/elements/labrad-plot1d.ts
```diff
@@ -306,11 +307,17 @@ export class Plot1D extends polymer.Base implements Plot {
...
-              this.zoom.x(this.xScale.domain([originX, x].map(this.xScale.invert).sort()))
-                  .y(this.yScale.domain([originY, y].map(this.yScale.invert).sort()));
+              var xScale = this.xScale,
+                  yScale = this.yScale,
+                  xMin = Math.min(xScale.invert(originX), xScale.invert(x)),
+                  xMax = Math.max(xScale.invert(originX), xScale.invert(x)),
+                  yMin = Math.min(yScale.invert(originY), yScale.invert(y)),
+                  yMax = Math.max(yScale.invert(originY), yScale.invert(y));
+              this.zoom.x(xScale.domain([xMin, xMax]))
+                       .y(yScale.domain([yMin, yMax]));
```
jwenner: Would it be good to have a comment somewhere around here about making sure we don't flip the axes?

jwenner:
@maffoo, at least so far as this version (actually, the one on matrix-silicon) having fixed the issue, LGTM. As far as the code, well, it would help if I knew any Java at all... Other than my one comment above, ask someone who knows anything about Java if you want a code-quality check.

client-js/app/elements/labrad-plot1d.ts
maffoo: will add a comment

maffoo:
@jwenner, this is actually typescript, not java, but all these curly-bracket languages look pretty similar :)

PR #123, opened 2015-12-30
refactor registry into single list 

joshm-3b6b58:
This simplifies the registry list into a single list rather than separate lists for directories and keys. This will make adding features easier down the line.

This makes the code easier to read by getting rid of a bunch of silly flags denoting the type of registry entry, etc.

This is the first in a series of changes I will be making to this. Right now the entries in the list aren't un-selectable, but they will be when I enable the selection of multiple entries.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -101,9 +101,9 @@
   <template>
     <div id="reg-area">
       <table>
-        <tbody is="selectable-table" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">
+        <tbody is="selectable-table" id="combinedList" selected="{{selectedIdx}}">
           <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc">
-            <tr name="{{item.name}}">
+            <tr name="{{item.name}}" class="dir">
               <td class="dir" 
                   name="{{item.name}}" 
                   draggable="true" 
@@ -117,8 +117,6 @@
               <td class="label-wide"></td>
             </tr>
           </template>
-        </tbody>
-        <tbody is="selectable-table" attr-for-selected="name"  selected="{{selKey}}" on-tap="selectKey">
           <template id="keyList" is="dom-repeat" items="{{keys}}" filter="filterFunc">
             <tr class="key" name="{{item.name}}">
               <td>
@@ -196,7 +194,7 @@
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
-    
+
      <paper-dialog id="dragCopyDialog" modal>
       <h1>Copy "<span id="originName"></span>" from <span id="originPath"></span></h1>
       <form>
@@ -210,7 +208,7 @@
     </paper-dialog>
 
     <paper-dialog id="copyDialog" modal>
-      <h1>Copy <span>{{selectType}}</span> "<span>{{selDir}}</span><span>{{selKey}}</span>"</h1>
+      <h1>Copy <span>{{selectedType}}</span> "<span>{{selectedItem}}</span>"?</h1>
       <form>
         <paper-input id="copyNameInput" label="New Name" always-float-label></paper-input>
         <paper-input id="copyPathInput" label="New Path" always-float-label></paper-input>
@@ -222,7 +220,7 @@
     </paper-dialog>
 
     <paper-dialog id="renameDialog" modal>
-      <h1>Rename <span>{{selectType}}</span> "<span>{{selDir}}</span><span>{{selKey}}</span>"</h1>
+      <h1>Rename <span>{{selectedType}}</span> "<span>{{selectedItem}}</span>"?</h1>
       <form>
         <paper-input id="renameInput" label="New Name" always-float-label></paper-input>
       </form>
@@ -233,17 +231,19 @@
     </paper-dialog>
 
     <paper-dialog id="deleteDialog" modal>
-      <h1>Delete <span>{{selectType}}</span> "<span>{{selDir}}</span><span>{{selKey}}</span>"?</h1>
+      <h1>Delete <span>{{selectedType}}</span> "<span>{{selectedItem}}</span>"?</h1>
       <div class="buttons">
         <paper-button on-click="doDelete" dialog-confirm autofocus>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
 
+
     <paper-dialog id="pendingDialog" modal>
       <h1 id="pendingOp"></h1> 
     </paper-dialog>
 
     <paper-toast id="toastTextSuccess" text="Directory Copied Successfully!" duration=3000></paper-toast>
+
   </template>
 </dom-module>
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -11,21 +11,15 @@ export class LabradRegistry extends polymer.Base {
 
   @property({type: Array, notify: true})
   path: Array<string>;
+  
+  @property({type: Array, notify: true})
+  selectedIdx: number;
 
   @property({type: Object})
   socket: RegistryApi;
 
   @property({type: String, notify: true})
   notify: string;
-
-  @property({type: String, notify: true, value: null})
-  selDir: string;
-
-  @property({type: String, notify: true, value: null})
-  selKey: string;
-
-  @property({type: String, notify: true, value: null})
-  selectType: string;
 
   @property({type: String, notify: true, value: ''})
   filterText: string;
@@ -37,9 +31,7 @@ export class LabradRegistry extends polymer.Base {
    */
   @observe('path')
   pathChanged(newPath: string[], oldPath: string[]) {
-    this.selDir = null;
-    this.selKey = null;
-    this.selectType = null;
+    this.selectedIdx = null;
     this.filterText = '';
   }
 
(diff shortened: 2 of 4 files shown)


client-js/gulpfile.js
```diff
@@ -51,7 +51,8 @@ gulp.task('compile-ts', function () {
       declarationFiles: false,
       noExternalResolve: true,
       experimentalDecorators: true,
-      emitDecoratorMetadata: true
+      emitDecoratorMetadata: true,
+      experimentalAsyncFunctions: true
```
joshm-3b6b58, line 55: I had to add this on my end to silence a whole bunch of compiler warnings...

client-js/app/elements/labrad-registry.html
```diff
@@ -101,9 +101,9 @@
   <template>
     <div id="reg-area">
       <table>
-        <tbody is="selectable-table" attr-for-selected="name" selected="{{selDir}}" on-tap="selectDir">
+        <tbody is="selectable-table" id="combinedList" selected="{{idx}}">
```
maffoo: maybe rename `idx` to something like `selectedIdx` to be more specific

client-js/app/elements/labrad-registry.html
```diff
@@ -101,9 +101,9 @@
...
           <template id="dirList" is="dom-repeat" items="{{dirs}}" filter="filterFunc">
-            <tr name="{{item.name}}">
+            <tr name="{{item.name}}" class = "dir">
```
maffoo: nit: remove spaces around `=` for element attributes

client-js/app/elements/labrad-registry.html
```diff
@@ -142,15 +140,15 @@
           <iron-icon icon="folder-open"></iron-icon>
           new dir
         </paper-button>
-        <paper-button on-click="copyClicked" disabled="{{!selected}}">
+        <paper-button on-click="copyClicked" disabled="{{selected}}">
```
maffoo: Seems strange to change the sense of `selected` here. Shouldn't these buttons be disabled only if nothing is selected?

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,49 @@ export class LabradRegistry extends polymer.Base {
     return item.name.match(this.regex);
   }
 
-  selectKey() {
-    this.selDir = null;
-    this.selectType = 'key';
+  @listen('idx-changed')
+  changedSelected() {
```
maffoo: Is this just here for debugging? Could probably remove since things are working.

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,49 @@ export class LabradRegistry extends polymer.Base {
...
+    console.log('ChangeSelected', this.idx);
   }
 
-  selectDir() {
-    this.selKey = null;
-    this.selectType = 'dir';
+  @computed()
+  selected(idx: number): boolean {
+    console.log("computing selected is: ", idx);
```
maffoo: This log message doesn't make sense to me.

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,49 @@ export class LabradRegistry extends polymer.Base {
...
+    if (idx != null) {
+        return false;
+      } else {
+        console.log('doesnt exist');
```
maffoo: ???

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,49 @@ export class LabradRegistry extends polymer.Base {
...
+        return true;
```
maffoo: This seems backwards. `selected` should be true if idx is non-null:

``` javascript
return idx != null;
```

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,49 @@ export class LabradRegistry extends polymer.Base {
...
+      }
+  }
+  @property({computed: '_computeSelectedType(idx)'})
+  selectedType: string;
```
maffoo: These can be written with the `computed` decorator, just like `selected` above:

``` javascript
@computed()
selectedType(idx: number): string {
  ...
}
```

Same for `selectedItem`.

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,49 @@ export class LabradRegistry extends polymer.Base {
...
+  
+  @property({computed: '_computeSelectedItem(idx)'})
+  selectedItem: string;
+  
+  _computeSelectedType(idx: number): string {
+    if (typeof this.dirs !== 'undefined') { //I don't know if there's a better way to do this
+      if (idx < this.dirs.length) {
```
maffoo: You could combine the two conditions:

``` javascript
if (this.dirs && idx < this.dirs.length) {
  return 'dir';
} else {
  return 'key';
}
```

client-js/app/elements/labrad-registry.ts
```diff
@@ -95,6 +105,7 @@ export class LabradRegistry extends polymer.Base {
     for (var j in resp.keys) {
       this.push('keys', {name: resp.keys[j], value: resp.vals[j]});
     }
+    this.$.combinedList.selected = null;
```
maffoo: Can you do this by assigning to `this.idx` instead of touching the DOM directly?

joshm-3b6b58: I tried that, it didn't work for me.

client-js/app/elements/labrad-registry.ts
```diff
@@ -366,11 +377,11 @@ export class LabradRegistry extends polymer.Base {
     var newPath = JSON.parse(this.$.copyPathInput.value);
 
     try {
-      if (this.selectType === 'dir') {
-        await this.socket.copyDir({path: this.path, dir: this.selDir, newPath: newPath, newDir: newName});
+      if (this.selectedType == 'dir') {
```
maffoo: Still best to use triple equals here (and numerous places below). Double equals can do weird type coercions and it's best not to get in the habit of using it.

client-js/app/elements/selectable-table.html
```diff
@@ -15,7 +15,7 @@
       is: 'selectable-table',
       extends: 'tbody',
       behaviors: [
-        Polymer.IronSelectableBehavior
+        Polymer.IronMultiSelectableBehavior
```
maffoo: Probably should leave this multi-selection for later, since we currently only support single selection.

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: I tried that first, but the compiler doesn't like it. It seems selectedType as something odd like a function that returns a string, rather than just a string and it throws errors. This was the only way I could work it so the compiler would be happy.

client-js/gulpfile.js
maffoo, line 55: This is fine, but maybe should put it in a separate commit when merging.

client-js/app/elements/labrad-registry.ts
maffoo: I see, you mean when you refer to `this.selectedType` in the code it picks up the method, rather than the computed property value.

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,49 @@ export class LabradRegistry extends polymer.Base {
...
+  selected(idx: number): boolean {
+    console.log("computing selected is: ", idx);
+    if (idx != null) {
+        return false;
+      } else {
+        console.log('doesnt exist');
+        return true;
+      }
+  }
+  @property({computed: '_computeSelectedType(idx)'})
```
maffoo: nit: add blank line between this decorator and the previous method.

joshm-3b6b58:
Thanks for the comments. Since this is UI heavy I wasn't writing test until we implement Selenium or some kind of browser testing framework. Are there any tests you would add for this?

client-js/app/elements/selectable-table.html
joshm-3b6b58: Thanks for picking this up, I was going to add multi-selectable to this PR, but figured I do this one thing at a time.

client-js/app/elements/labrad-registry.html
joshm-3b6b58: OK

client-js/app/elements/labrad-registry.html
joshm-3b6b58: ah yes. OK

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Yup

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: It's a leftover from some old debugging. Removed it.

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Yes, if I put it in the `if` statements on `doCopy()` for example. Also, they're passed to the RPC calls on line 375, for example. Those generate compiler errors.

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Yeah, sorry I confused myself when I changed something. I've re-reversed the returns in these if statements now.

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Ah thank you, I didn't realize you could do that in js. I knew it worked in python.

client-js/app/elements/labrad-registry.ts

joshm-3b6b58:
@maffoo I think I made all the changes, can you take a look when you get a chance? Thanks.

client-js/gulpfile.js
joshm-3b6b58, line 55: I may need a hand doing that when I rebase and squash...

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,41 @@ export class LabradRegistry extends polymer.Base {
     return item.name.match(this.regex);
   }
 
-  selectKey() {
-    this.selDir = null;
-    this.selectType = 'key';
+  @computed()
+  selected(selectedIdx: number): boolean {
+    if (selectedIdx != null) {
```
maffoo: I would suggest just

``` javascript
return selectedIdx !== null;
```

(note use of `!==` instead of `!=`)

client-js/app/elements/labrad-registry.ts
```diff
@@ -61,31 +53,41 @@ export class LabradRegistry extends polymer.Base {
...
+      return 'key';
+    }
   }
 
-  @computed()
-  selected(selectType: string, selDir: string, selKey: string): boolean {
-    console.log('selectType', selectType, 'selDir', selDir, 'selKey', selKey);
-    return (selectType === 'dir' && selDir != null) || (selectType === 'key' && selKey != null);
+  _computeSelectedItem(selectedIdx: number): string {
+    if (typeof this.dirs !== 'undefined') {
```
maffoo: can do the same thing here as in `_computeSelectedType`:

``` javascript
if (this.dirs) {
  return ...
}
```

maffoo:
Two small comments, otherwise LGTM.

PR #125, opened 2015-12-31
Add drag move for folders

joshm-3b6b58:
implemented drag move for folders. Will implement for keys if this all looks good.

### DIFF
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -202,18 +202,19 @@ export class LabradRegistry extends polymer.Base {
       console.log('dropped on', event.target.closest('td').name, 'with data', data ); 
       var newPath: string[] = this.path.slice(),
           oldFullPath: string[] = data.path,
-          dialog = this.$.dragCopyDialog,
-          copyOriginName = this.$.originName,
-          copyOriginPath = this.$.originPath,
-          copyNameElem = this.$.dragCopyNameInput,
-          copyPathElem = this.$.dragCopyPathInput;
+          dialog = this.$.dragDialog,
+          dragOriginName = this.$.originName,
+          dragOriginPath = this.$.originPath,
+          dragNameElem = this.$.dragNameInput,
+          dragPathElem = this.$.dragPathInput,
+          dragOp = this.$.dragOp;
       newPath.push(event.target.closest('td').name);
-      copyNameElem.value = data.name;
-      copyOriginName.textContent = data.name;
-      copyOriginPath.textContent = JSON.stringify(data.path);
-      copyPathElem.value = this.pathToString(newPath);
+      dragNameElem.value = data.name;
+      dragOriginName.textContent = data.name;
+      dragOriginPath.textContent = JSON.stringify(data.path);
+      dragPathElem.value = this.pathToString(newPath);
       dialog.open();
-      window.setTimeout(() => copyNameElem.$.input.focus(), 0);
+      window.setTimeout(() => dragNameElem.$.input.focus(), 0);
     }
   }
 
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-registry.ts
```diff
@@ -380,22 +374,39 @@ export class LabradRegistry extends polymer.Base {
...
-      console.log(resp)//this.repopulateList();
-      this.$.pendingDialog.close()
-      this.$.toastTextSuccess.show();
-    } catch (error) {
-      this.handleError(error);
+    if (this.$.dragOp.innerText === 'Copy') {
+      try {
+        this.$.pendingDialog.open();
+        this.$.pendingOp.innerText = "Copying...";
+        console.log(this.$.dragOp.innerText)
```
maffoo: I would remove these log statements, now that things are working.

client-js/app/elements/labrad-registry.ts
```diff
@@ -380,22 +374,39 @@ export class LabradRegistry extends polymer.Base {
...
+      }
+    }
+    else if (this.$.dragOp.innerText === 'Move') {
+      try {
+        this.$.pendingDialog.open();
+        this.$.pendingOp.innerText = "Moving...";
+        console.log(this.$.dragOp.innerText)
+        var resp = await this.socket.moveDir({path: oldPath, dir: oldName, newPath: newPath, newDir: newName});
+        console.log(resp)
+        this.$.pendingDialog.close()
```
maffoo: Why no toast here?

client-js/app/elements/labrad-registry.ts
```diff
@@ -380,22 +374,39 @@ export class LabradRegistry extends polymer.Base {
...
-    } catch (error) {
-      this.handleError(error);
+    if (this.$.dragOp.innerText === 'Copy') {
+      try {
+        this.$.pendingDialog.open();
+        this.$.pendingOp.innerText = "Copying...";
+        console.log(this.$.dragOp.innerText)
+        var resp = await this.socket.copyDir({path: oldPath, dir: oldName, newPath: newPath, newDir: newName});
+        console.log(resp)
+        this.$.pendingDialog.close()
```
maffoo: I would put this close call in a finally block so it happens whether the copy succeeds or fails.

client-js/app/elements/labrad-registry.ts
```diff
@@ -220,27 +218,24 @@ export class LabradRegistry extends polymer.Base {
...
-            copyOriginPath = this.$.originPath,
-            copyNameElem = this.$.dragCopyNameInput,
-            copyPathElem = this.$.dragCopyPathInput;
-        copyNameElem.value = data.name;
-        copyOriginName.textContent = data.name;
-        copyOriginPath.textContent = JSON.stringify(data.path);
-        copyPathElem.value = this.pathToString(this.path);
-        this.$.dragCopyDialog.open();
-        window.setTimeout(() => copyNameElem.$.input.focus(), 0);
+    if (JSON.stringify(this.path) != JSON.stringify(data.path)) {
```
maffoo: Should check the operation here and only return early if trying to move; copying to the same path is perfectly reasonable.

maffoo:
A few small comments about the code, but the basic functionality seems fine. I'd like to make some tweaks to the drag-drop behavior, but I'll file separate issues. Other than the minor comments, LGTM.

joshm-3b6b58:
Thanks for the comments. Since this is UI heavy I wasn't writing test until we implement Selenium or some kind of browser testing framework. Are there any tests you would add for this?

client-js/app/elements/labrad-registry.ts

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Added another toast for successful move

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Done

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: How do we want to handle renaming the folder? adding  "Copy of" on the front?

maffoo: When we pop up a normal directory copy dialog we just use the plain directory name, not "Copy of <name>", so I'd say just do the same here too.

joshm-3b6b58: Right but we don't want users to try and make a copy of the directory with the same name, since they can't, to it has to at least default to a unique new folder name, or gracefully handle the error of copying without changing the name. 

I guess the latter is desirable, but introduces extra logic to the front-end. 

I also guess I've never checked what happens if I try to copy a folder with the same name. Ill see how that fails.

Edit: when I try to copy a folder without changing the name is basically just does nothing. It doesn't try to create a new dir, it doesn't send an error. That could be a confusing behavior, since the user might expect the folder to have been copied.

maffoo:
I can't think of any convenient way to test this; even with something like selenium I'm not sure that it's possible to test drag and drop behavior. This might be one where we just have to do manual tests for a while, though I'd be glad to be proven wrong on that.

PR #126, opened 2016-01-04
Clean up the login logic

maffoo:
There was a subtle concurrency bug with the login logic which sometimes resulted in a login dialog popping up, even when the credentials had already been saved in local or session storage. This fixes that problem and also cleans up the logic overall to use promises and async functions more consistently.

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -10,6 +10,7 @@ import * as registry from "./registry";
 import * as datavault from "./datavault";
 import * as nodeApi from "./node";
 import * as rpc from "./rpc";
+import {obligate} from "./obligation";
 
 import {LabradApp} from "../elements/labrad-app";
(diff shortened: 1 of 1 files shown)


client-js/app/scripts/app.ts
```diff
@@ -636,7 +644,7 @@ window.addEventListener('WebComponentsReady', () => {
       throw new Error('no credentials');
     } else {
       try {
-        return mgr.login({
+        await mgr.login({
```
maffoo, line 651: This was the bug; we were returning instead of waiting for the login to complete or fail.

joshm-3b6b58:
Devil's advocate; if there are subtle concurrency bugs, should we have a unit test for this?

maffoo:
In an ideal world, yes. But it's not really possible to unit test this as it's tied to polymer UI, etc.

joshm-3b6b58:
LGTM

PR #127, opened 2016-01-04
Fix y-axis label on 2D plots

maffoo:
We have to choose the y-axis label differently for 2D versus 1D plots. This was broken when we consolidated the 1- and 2-D plotting code.


joshm-3b6b58:
LGTM

Issue #124, opened 2015-12-31
reg.copyDir returns to early when copying a folder of folders


joshm-3b6b58:
When copying a large directory of keys, the `await registry.copyDir` will wait until the whole folder is copied to return. However when copying a folder of folders, it will return after the first folder is copied. This causes some confusion because the folder will appear to be copied, when the folders contained therein aren't copied yet.

maffoo:
I'm not able to reproduce this issue exactly, but I've fixed up the directory copy logic in #128 and that may help here as well.

joshm-3b6b58:
It only happens to me when I have a folder that contains a subfolder with a
lot of keys that takes a while to copy. I'll see if the latest PR fixes
this.

PR #128, opened 2016-01-04
Clean up registry copy logic to not interleave reading and writing.

maffoo:
We previously changed the dir copying code to avoid a potential infinite recursion when copying a folder directly inside itself, however the fix did not handle a case when a folder is copied into a subfolder of itself. This change fixes these sorts of problems by first loading the entire directory contents, including all subdirectories, and then writing the whole thing to the new location in one shot. This is more memory intensive on the server, but for typical copies that we do it should be fine.

### DIFF
--- a/server/src/main/scala/org/labrad/browser/RegistryApi.scala
+++ b/server/src/main/scala/org/labrad/browser/RegistryApi.scala
@@ -22,7 +22,19 @@ object RegistryApi {
   }
 }
 
+/**
+ * The contents of a registry directory, potentially including recursive subdirectory contents
+ */
+case class RegistryContents(dirs: Seq[(String, RegistryContents)], keys: Seq[(String, Data)])
+object RegistryContents {
+  val Empty = RegistryContents(dirs = Nil, keys = Nil)
+}
 
+/**
+ * Contents of a single registry directory, not including any subdirectory contents.
+ *
+ * This is in a form suitable for encoding in JSON.
+ */
 case class RegistryListing(path: Seq[String], dirs: Seq[String], keys: Seq[String], vals: Seq[String])
 object RegistryListing { implicit val format = Json.format[RegistryListing] }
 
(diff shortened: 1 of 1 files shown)


joshm-3b6b58:
Tested this and it works. Fixes bug #124 The scala is still greek to me...
@DanielSank feel like applying your newly learned scala-fu to this PR?

server/src/main/scala/org/labrad/browser/RegistryApi.scala
```diff
@@ -22,7 +22,19 @@ object RegistryApi {
   }
 }
 
+/**
+ * The contents of a registry directory, potentially including recursive subdirectory contents
+ */
+case class RegistryContents(dirs: Seq[(String, RegistryContents)], keys: Seq[(String, Data)])
+object RegistryContents {
+  val Empty = RegistryContents(dirs = Nil, keys = Nil)
+}
```
DanielSank, line 31: @maffoo I see this pattern over and over but I don't get it. What's the point of making a case class and an object of the same name?

DanielSank:
Hold the phone... async and await exist in Scala?!

server/src/main/scala/org/labrad/browser/RegistryApi.scala
maffoo, line 31: An object with the same name as a trait or class is called a "companion object" in scala. This is a useful place to put things that are associated with the class as a whole, rather than associated with a specific instance of the class. In java, one would use static members for this instead, but scala doesn't have statics.

For case classes, the compiler automatically generates a companion object with various convenience features, like an apply method for creating instances (so you can say `MyCaseClass(x)` to construct an instance, which actually calls a method on the companion object `MyCaseClass.apply(x)`, instead of calling the constructor like `new MyCaseClass(x)`).

In the code here, I've explicitly defined a companion object to hold a special empty instance of `RegistryContents`. But in addition, the compiler adds the autogenerated members as usual for case class companion objects; for example you can see that I'm using the companion object apply method to create the Empty instance.

maffoo:
@DanielSank, yes, these come from the scala-async project (https://github.com/scala/async). They are macros that do the necessary code rewrites, but there are a couple of caveats that rather severely limit their usefulness: error handling is kind of crappy, because if a Future that you're awaiting fails, that exception does not get raised by the await for your code to try to handle, rather the whole async block fails immediately (contrast with typescript where you can use try/catch like normal to handle async errors); second, because scala for-loops are implemented with calls to foreach, map and flatMap methods, you can't await inside a loop, which is really annoying. I still think async/await is useful, but not nearly as useful as it _could_ be.

PR #129, opened 2016-01-05
Move activities and places into their own modules.

maffoo:
Pull the Activity classes out into a separate module instead of putting them inside an event callback in the main app module. Among other things, this forces us to declare all the dependencies for each activity in its constructor, rather than close over variables from the outer scope when creating activities. We also put the functions for generating urls into a single location so they can be more self-contained.

### DIFF
--- a/client-js/app/scripts/activities.ts
+++ b/client-js/app/scripts/activities.ts
@@ -1,0 +1,368 @@
+import {Activity, ActivityState} from "./activity";
+import {AsyncQueue} from "./async-queue";
+import {Lifetime} from "./lifetime";
+import * as manager from "./manager";
+import * as places from "./places";
+import * as registry from "./registry";
+import * as datavault from "./datavault";
+import * as node from "./node";
+
+import {LabradGrapher} from "../elements/labrad-grapher";
+import {LabradGrapherLive, LabeledPlot} from "../elements/labrad-grapher-live";
+import {LabradManager} from "../elements/labrad-manager";
+import {LabradNodes} from "../elements/labrad-nodes";
+import {LabradRegistry} from "../elements/labrad-registry";
+import {LabradServer} from "../elements/labrad-server";
+import {Plot} from "../elements/labrad-plot";
+
+export class RegistryActivity implements Activity {
+  private elem: LabradRegistry;
+  private lifetime = new Lifetime();
+
+  constructor(private api: registry.RegistryApi, public path: Array<string>) {}
+
+  async start(): Promise<ActivityState> {
+    this.api.newItem.add(() => this.onNewItem(), this.lifetime);
+    console.log('loading registry:', this.path);
+    await this.api.watch({path: this.path});
+    var listing = await this.api.dir({path: this.path});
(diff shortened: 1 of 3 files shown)


joshm-3b6b58:
Sorry, I've been in the cleanroom most of the week, I'll take a look at this soon :/

joshm-3b6b58:
Some comments about the overall structure of this would be very helpful for people trying to navigate this to add functionality.

client-js/app/scripts/activities.ts
```diff
@@ -0,0 +1,363 @@
...
+
+    var breadcrumbs = [];
+    for (var i = 0; i <= this.path.length; i++) {
+      breadcrumbs.push({
+        name: (i == 0) ? 'grapher' : this.path[i-1],
+        isLink: i < this.path.length,
+        url: places.grapherUrl(this.path.slice(0, i))
+      });
+    }
+    var breadcrumbExtras = [
```
joshm-3b6b58, line 84: This can be broken out to be more readable as on line 87

client-js/app/scripts/app.ts
```diff
@@ -2,10 +2,10 @@
 
 import page from "page";
 
-import {Activity, ActivityState} from "./activity";
-import {AsyncQueue} from './async-queue';
-import {Lifetime} from "./lifetime";
+import {Activity} from "./activity";
+import * as activities from "./activities";
 import * as manager from "./manager";
+import * as places from "./places";
```
joshm-3b6b58: Nit, this import isn't used.

client-js/app/scripts/places.ts
```diff
@@ -0,0 +1,29 @@
+// Function for creating URLs of all the places in the application.
+
```
joshm-3b6b58, line 2: So this just creates URLs, it doesn't handle routing at all?

joshm-3b6b58:
Other than I think it needs some commentary it LGTM. For example, if I want to add a widget to the web app, when do I create an activity? Why do I need a place? What goes is app.ts?

client-js/app/scripts/activities.ts
maffoo, line 84: done

client-js/app/scripts/app.ts
maffoo: removed

client-js/app/scripts/places.ts
maffoo, line 2: That's correct. The "routing" happens in `app.ts` which uses the page.js library to set up routes that launch various activities. We should probably refactor this to keep the info about creating and parsing urls in this one module, but that will be for another PR.

maffoo:
I added a few comments in `app.ts` to explain how the routes and activities are used, and how the routes relate to the URL functions in the places module. As I mentioned above, there's a bit more refactoring I want to do to keep all the URL parsing/generating code in one place instead of split between `app.ts` and `places.ts`. I think once that is in place it'll be more clear what needs to be done to add new pages to the application.

joshm-3b6b58:
LGTM

PR #136, opened 2016-01-19
add drag move+copy for keys

### LINKED ISSUE #91: make keys copy + move draggable
Reminder to me to do this

joshm-3b6b58:
Please (manually) test all the different cases here since it can be hard to keep track of.

AFAICT we have:
drag copy folder on folder
drag move folder on folder
drag copy folder into window
drag copy key on folder
drag copy key on window
drag move key on folder
drag move key on window

There are also some lame bugs that I didn't notice before since there are so many cases, that I ended up fixing in this PR.

fixes #91

### DIFF
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -188,7 +188,7 @@ export class LabradRegistry extends polymer.Base {
   startDrag(event) {
     //detect start of drag event, grab info about target
     var data: any; //I tried enum, but kept getting errors...
-    data = {path: this.path, name: event.target.name};
+    data = {path: this.path, name: event.target.name, kind: event.target.className.split(' ')[0]};
     event.dataTransfer.setData('text', JSON.stringify(data));
     event.dataTransfer.effectAllowed = 'copyMove';
   }
@@ -210,12 +210,14 @@ export class LabradRegistry extends polymer.Base {
    * behaviour for dropping on folders
    */
   dirDrop(event) {
+    event.stopPropagation();
     var data = JSON.parse(event.dataTransfer.getData('text'));
+    this.$.dragDialog.dragData = data;
     var newPath: string[] = this.path.slice();
     newPath.push(event.target.closest('td').name);
     event.target.closest('td').classList.remove('over');
-
     this.$.dragNameInput.value = data.name;
+    this.$.dragClass.textContent = data.kind;
     this.$.originName.textContent = data.name;
     this.$.originPath.textContent = JSON.stringify(data.path);
     this.$.dragPathInput.value = this.pathToString(newPath);
@@ -223,11 +225,14 @@ export class LabradRegistry extends polymer.Base {
     if (event.ctrlKey || event.button == 2 ) {
       //should I use switch and case here instead of ifs?
       this.$.dragOp.innerText = 'Copy';
+      this.$.dragDialog.open();
+      setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
     } else {
       this.$.dragOp.innerText = 'Move';
+      this.$.dragDialog.open();
+      setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
     }
-    this.$.dragDialog.open();
-    window.setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
+
   }
 
   /**
@@ -237,20 +242,23 @@ export class LabradRegistry extends polymer.Base {
   handleDrop(event) {
     event.preventDefault();
     var data = JSON.parse(event.dataTransfer.getData('text'));
-    var newPath: string[] = this.path.slice();
-    newPath.push(event.target.closest('td').name);
+    this.$.dragDialog.dragData = data;
     event.target.closest('td').classList.remove('over');
     this.$.dragNameInput.value = data.name;
+    this.$.dragClass.textContent = data.kind;
     this.$.originName.textContent = data.name;
     this.$.originPath.textContent = JSON.stringify(data.path);
     this.$.dragPathInput.value = this.pathToString(this.path);
-      if (event.ctrlKey) {
-        this.$.dragOp.innerText = 'Copy';
-      } else if(JSON.stringify(this.path) != JSON.stringify(data.path)) {
-        this.$.dragOp.innerText = 'Move';
-      }
+
+    if (event.ctrlKey) {
+      this.$.dragOp.innerText = 'Copy';
       this.$.dragDialog.open();
-      window.setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
+      setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
+    } else if(JSON.stringify(this.path) != JSON.stringify(data.path)) {
+      this.$.dragOp.innerText = 'Move';
+      this.$.dragDialog.open();
+      setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


client-js/app/elements/labrad-registry.ts
```diff
@@ -193,47 +193,60 @@ export class LabradRegistry extends polymer.Base {
    * behaviour for dropping on folders
    */
   dirDrop(event) {
+    event.stopPropagation();
+    console.log('dirDrop');
```
joshm-3b6b58: I'll delete this.

client-js/app/elements/labrad-registry.ts
```diff
@@ -206,34 +207,41 @@ export class LabradRegistry extends polymer.Base {
     if (event.ctrlKey || event.button == 2 ) {
       //should I use switch and case here instead of ifs?
       this.$.dragOp.innerText = 'Copy';
+      this.$.dragDialog.open();
+      window.setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
```
maffoo: I think you can just use `setTimeout` instead of `window.setTimeout`. We should get in this habit because the former will work no matter what the global object is called, e.g. in unit tests and in the browser, whereas the latter only works in browsers.

client-js/app/elements/labrad-registry.ts
```diff
@@ -206,34 +207,41 @@ export class LabradRegistry extends polymer.Base {
...
     var data = JSON.parse(event.dataTransfer.getData('text'));
     var newPath: string[] = this.path.slice();
     newPath.push(event.target.closest('td').name);
     event.target.closest('td').classList.remove('over');
     this.$.dragNameInput.value = data.name;
+    this.$.dragClass.textContent = data.kind;
     this.$.originName.textContent = data.name;
     this.$.originPath.textContent = JSON.stringify(data.path);
     this.$.dragPathInput.value = this.pathToString(this.path);
       if (event.ctrlKey) {
```
maffoo: There's an extra level of indentation here that's not needed.

client-js/app/elements/labrad-registry.ts
```diff
@@ -206,34 +207,41 @@ export class LabradRegistry extends polymer.Base {
...
+        this.$.dragDialog.open();
+        window.setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
       } else if(JSON.stringify(this.path) != JSON.stringify(data.path)) {
         this.$.dragOp.innerText = 'Move';
+        this.$.dragDialog.open();
+        window.setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
       }
-      this.$.dragDialog.open();
-      window.setTimeout(() => this.$.dragPathInput.$.input.focus(), 0);
+
```
maffoo: remove blank line

client-js/app/elements/labrad-registry.ts
```diff
@@ -387,25 +395,40 @@ export class LabradRegistry extends polymer.Base {
       try {
         this.$.pendingDialog.open();
         this.$.pendingOp.innerText = "Copying...";
-        var resp = await this.socket.copyDir({path: oldPath, dir: oldName, newPath: newPath, newDir: newName});
-        this.$.toastCopySuccess.show();
+        switch (this.$.dragClass.innerText) {
```
maffoo: I'm not a fan of passing data through the html text representation of the dialog element. Instead, I would suggest storing the data from the data transfer event in an attribute of the dialog, and then looking it up, e.g.:

``` javascript
// in handleDrop
var data = JSON.parse(...);
this.$.dragDialog.dragData = data;

// here
switch (this.$.dragDialog.dragData['kind']) {
  ...
}
```

Of course, even better than this would be to have an alternate API for showing dialogs that returns a Promise which fires when the dialog is dismissed. Then we could just put the whole dialog flow inside a single async method and we wouldn't have to pass data through the DOM like this at all. But that's a change for another time.

client-js/app/elements/labrad-registry.ts
```diff
@@ -206,34 +207,41 @@ export class LabradRegistry extends polymer.Base {
     if (event.ctrlKey || event.button == 2 ) {
```
maffoo, line 225: I was reading a bit about the DOM drag-n-drop support and it seems like we should be using `event.dataTransfer.dropEffect` here to decide what to do. This should get assigned values by the browser and so should work better cross-platform. For example, on mac, copy is normally triggered using the Alt key, rather than Ctrl. Have you tried checking `dropEffect` here instead?

maffoo, line 225: nvm; after digging a bit more it seems that support for `dropEffect` is woefully inconsistent across different browsers, with chrome being a particular offender. So looks like that's not really something we can rely on.

client-js/app/elements/labrad-registry.ts
```diff
@@ -206,34 +207,41 @@ export class LabradRegistry extends polymer.Base {
...
   /**
    * handles folders dropped not into folders
    */
   @listen('drop')
-  handleDrop(event) {{"path":["two"]}
+  handleDrop(event) {
     event.preventDefault();
     var data = JSON.parse(event.dataTransfer.getData('text'));
     var newPath: string[] = this.path.slice();
     newPath.push(event.target.closest('td').name);
```
maffoo: I get errors on this line:

```
Uncaught TypeError: Cannot read property 'name' of null
```

If the drop happens somewhere else in the registry list (not on a subdirectory) we should just use the current path and not push an additional segment onto it.

maffoo:
For some reason, copying is not working for me when trying this on mac. If I drag and then hold down Ctrl, the drop even does not even fire. As I mentioned above, other mac apps seem to want to use Alt to toggle between move/copy behavior, so that may be part of the problem.

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: d'oh. OK done

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: done

client-js/app/elements/labrad-registry.ts
joshm-3b6b58: Agreed on both fronts. Adding dialogs involves a lot more boilerplate than necesary. I'll fix this using an attribute.

joshm-3b6b58:
Is it not working at all? For keys and folders?

maffoo:
Moving a key or dir onto a directory works. Moving onto a key (to move to the current folder instead of a subfolder) does not work. Copying by holding down Ctrl does not work for keys or dirs.

joshm-3b6b58:
@maffoo so I'm using this on my mac and moving or copying onto a key to move into the current folder both work for me. 
Copying is strange since I have to hold down ctrl+alt. I'm having trouble understanding best practices for detecting differences between mac and not mac for the ctrl key vs the meta key, but the standard in finder seems to be holding down alt is drag copy. Do these functions at least work for you?

maffoo:
Moving and copying keys and folders works for me; copying requires Ctrl+Alt, as you said.

This change LGTM for now, though of course you'll want to rebase on the latest master and make sure things still work there. We can sort out the cross-platform modifier keys issue later.

joshm-3b6b58:
@maffoo I did the rebase, do you mind quickly double checking this for me? The drag actions seemed a bit flakey for me at first, I disabled the edge-swiping on the drawer panel to make this easier. Can you test it when you get the chance before I push to master?

maffoo:
LGTM

PR #137, opened 2016-01-19
add roboto mono for registry keys and values

### LINKED ISSUE #100: Use monospace font in registry editor

joshm-3b6b58:
@DanielSank can you take a look? This also fixes a brainfart I made while rebase :/

fixes #100

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -97,6 +97,10 @@
     dir-item {
       white-space: nowrap;
     }
+    .mono {
+      font-family: 'roboto mono', monospace;
+    }
+
   </style>
   <template>
     <div id="reg-area">
@@ -107,8 +111,8 @@
               <td class="dir" 
                   name="{{item.name}}" 
                   draggable="true" 
-                  on-dragover="onDirDragOver" 
-                  on-dragleave="onDirDragLeave" 
+                  on-dragover="onDirDragOver"
+                  on-dragleave="onDirDragLeave"
                   on-drop="dirDrop">
                 <a is="app-link" path="{{item.url}}" href="{{item.url}}">
                   <dir-item kind="folder">{{item.name}}</dir-item>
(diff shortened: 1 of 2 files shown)


DanielSank:
In my relative ignorance I can say it look good. Can you put it up on a test server somewhere to look at, or even post a screenshot?

joshm-3b6b58:
Good idea

screen shot 2016-01-19 at 11 44 34 am

joshm-3b6b58:
@DanielSank does this look OK?

client-js/app/elements/labrad-registry.html
```diff
@@ -97,6 +98,10 @@
     dir-item {
       white-space: nowrap;
     }
+    .mono {
+      font-family: 'roboto mono', monospace;
```
maffoo, line 101: I presume having the monospace fallback here means this will still use a monospace font (like courier) even if offline and can't connect to the net to download roboto mono. Have you tested that? I want to make sure that the app still works in an offline scenario, e.g. on lab computers that might not have access to the wider internet.

joshm-3b6b58, line 101: That's the idea. I'm actually not sure how to best test this, since my machines probably already have the fonts installed.

client-js/app/elements/labrad-registry.html
```diff
@@ -2,6 +2,7 @@
 <link rel="import" href="../bower_components/paper-toast/paper-toast.html">
 <link rel="import" href="app-link.html">
 <link rel="import" href="dir-item.html">
+<link href='https://fonts.googleapis.com/css?family=Roboto+Mono' rel='stylesheet' type='text/css'>
```
maffoo: just curious, if this is included in another element as well, will it get deduped and downloaded only once?

client-js/app/elements/labrad-registry.html
maffoo, line 101: maybe open in an incognito window after disabling your internet connection

joshm-3b6b58, line 101: Yup, it works (icognito, offline on the right):
screenshot from 2016-01-25 14 54 02

client-js/app/elements/labrad-registry.html
joshm-3b6b58: Good question, I have no idea. I could probably put this in elements.html

joshm-3b6b58: I can move the import to elements.html and it still works, this way the font can be called in the local css without including the import. I guess that wouldn't make the custom elements standalone, but at this point they are pretty interdependent anyhow.

maffoo:
LGTM

PR #138, opened 2016-01-25
Fix bad rebase in labrad-registry.ts

maffoo:
@joshmutus, I thought you had fixed this as part of another change, but I don't see the fix in master.


joshm-3b6b58:
It's part of the mono-type PR but hasn't gotten an LGTM.

That being said, LGTM

PR #139, opened 2016-01-26
Update jsonrpc macro to allow any number of params and default values

maffoo:
Previously, we had separate cases spelled out explicitly for each different number of parameters expected by the scala method being called. But, by taking advantage of the unhygienic nature of quasiquote macros, we can simplify this. In particular, the ASTs to extract each argument refer to a term called "params" that is not defined until later when we put these ASTs inside a macros that defines the function invoked with the jsonrpc params.

Also fix the handling of default parameters, so that we can correctly dispatch to methods with default parameter values.


maffoo:
@DanielSank, since I know you've looked at this recently.

DanielSank:
Looks like I'm gunna have to learn something to review this. Might take a bit.

DanielSank:
LGTM.

PR #140, opened 2016-01-26
Display error message and reconnect after connection loss

maffoo:
If we lose the websocket connection to the web server, or the server loses its connection to the labrad manager, pop up a dialog box informing the user and attempt to reconnect after some delay.


maffoo:
bump @joshmutus

joshm-3b6b58:
Checked out tested, killed the manager, got the dialog. LGTM! (sorry it took so long)

PR #149, opened 2016-02-17
Make it optional to include params in dataset info

maffoo:


joshm-3b6b58:
I don't really understand what's happenening here. Is this about displaying the dataset parameters? When I click a dataset, the parameters will show up on the right hand side for a split second before page plot renders on top of everything.

maffoo:
@joshmutus, the goal here is to make it possible to fetch dataset info with or without params. I've now fixed the event handling issue so that when we go render a dataset we don't needlessly fetch the params anyway.

joshm-3b6b58:
LGTM

PR #150, opened 2016-02-17
Display cursor coords when mousing over plots.

### LINKED ISSUE #144: Grapher: Possible to add coordinates at cursor?
The old Grapher displays the coordinates at the cursor and updates such coordinates as the cursor is moved. When trying to manually select an operation point, this can be useful. Is this even possible?

maffoo:
Fixes #144


jwenner:
This functionality works well on matrix-reloaded. This, as far as the functionality, LGTM assuming that this PR is the same as that on matrix-reloaded vis-a-vis this feature.

What documentation is considered appropriate for HTML and typescript? Since you don't have any code documentation here. So unless this is typical and considered appropriate, "looks bad to me" relative to docstrings.

client-js/app/elements/labrad-plot.ts
```diff
@@ -95,6 +95,9 @@ export class Plot extends polymer.Base {
   @property({type: Number, value: 0})
   numIndeps: number;
 
+  @property({type: String, value: ''})
+  currPos: string;
```
joshm-3b6b58, line 99: Why make this a string?

client-js/app/elements/labrad-plot.html
```diff
@@ -26,6 +33,7 @@
   </style>
   <template>
     <div id="plot"></div>
+    <div id="pos">{{currPos}}</div>
```
joshm-3b6b58: It's a bit strange to me for `currPos` to also include the text ``current position:`

Probably splitting hairs here but the way that seems natural to me is to write this as:

```
<div id="pos">The current position is: <span>{{currPos}}</span></div>
```

and have currPos just be numbers. I guess there aren't tuples in ts so you might have to do currX and currY, which is even more verbose.

joshm-3b6b58:
Also, I'm getting something like 15 decimal places on the coordinate display. It would be easier for me to read in scientific notation.

joshm-3b6b58:
Nit: there's a layout bug in live-view
screenshot from 2016-02-19 10 04 56

client-js/app/elements/labrad-plot.ts
maffoo, line 99: I don't want to put string formatting code in the template, so the string has to be assembled here and passed to the template as a string. You could pass numbers, but then you'd get the same 15-digit default formatting that you've pointed out is kind of annoying.

joshm-3b6b58, line 99: I see. I would say keep it as a string, but I would move the expository text to the html template.

maffoo, line 99: done

maffoo:
@joshmutus, I cannot reproduce the layout issue in chrome on mac. Are you seeing that on linux?

joshm-3b6b58:
@maffoo Yes, Linux

joshm-3b6b58:
@maffoo so the `prettyNumber()` works for the usual test cases, but when I tried to make the x value significantly larger it behaves in a way that I find undexpected:
screenshot from 2016-02-22 13 50 43

The x value doesn't appear in exponential form and the y value is truncated to 1 digit.

Here are the first 10 array values:

```
[ 0.          0.29340001]
[  1.00000000e+07   2.02666955e+00]
[  2.00000000e+07  -4.01688210e+00]
[  3.00000000e+07   5.14093998e+00]
[  4.00000000e+07  -5.02656140e+00]
[  5.00000000e+07   4.51872983e+00]
[  6.00000000e+07  -2.32641769e+00]
[  7.00000000e+07   6.34071053e-01]
[  8.00000000e+07   2.06855246e+00]
[  9.00000000e+07  -3.31805019e+00]
```

maffoo:
@joshmutus, should be fixed now. In addition to the heuristic measure of how long the fixed point vs exponential strings are, I've set 1 million as a cutoff beyond which we always use exponential format (but still with the minimal number of significant digits based on the pixel resolution). That number can easily change, of course. Let me know what you think.

joshm-3b6b58:
LGTM

joshm-3b6b58:
Sorry this took so long :(

PR #151, opened 2016-02-17
Remove play

### LINKED ISSUE #64: Remove playframework
This is related to #58 (and possibly #59). The playframework currently determines how the server code is packaged and run, and the way it does this is different from our other scala binaries like the manager and various servers. If we use a different web server library then we can have more control over how this works.

Currently we do not use much in the way of play-specific functionality. The server just serves some static files (produced by the node.js-based build in `client-js`) and provides api connections over websockets. The websocket layer is play-agnostic with just a small shim in `BrowserController.scala` to adapt it to the way play handles websockets, so it should be fairly easy to swap in a different web server library.

The logical choice here would be [netty](http://netty.io/), which we already are using in scalabrad and which is also used in gRPC, among many other projects.

maffoo:
Fixes #64

Get rid of play framework for the web server, and instead use straight netty. This lets us control the server config more tightly, by parsing command line params, for example. Also modify the server to allow connecting to and controlling multiple labrad managers.


joshm-3b6b58:
Play is a Scala framework, right? Why does this PR touch so much client-side code?

maffoo:
@joshmutus, I lumped the play->netty change in with some changes to allow the interface to connect to multiple managers on the backend. That's where all the client-side changes are coming from. I can split this up if you prefer; I know these big combined changes can be a pain to review.

joshm-3b6b58:
You don't have to split it up, knowing what the motivation for the client-side changes was helps.

joshm-3b6b58:
@maffoo something's not working for me. What am I missing?
1. I checked out the `remove-play` branch
2.  restarted the scala server (`ctrl-d` then `$ server/run`)
3.  `gulp clean, gulp serve`-d
4. restarted my manager, redid steps 2,3
   But I can't seem to get a connection through the client: 
   screenshot from 2016-02-23 16 26 23

maffoo:
@joshmutus, did you restart sbt? Does the server start and indicate that it is listening on a port? Also, can you check in the dev tools what port the client is trying to connect to with the websocket? It should be 7667. You can alternately just browse to localhost:7667 and talk to the server directly (you'll need to run just `gulp` to build the dist version of the client).

joshm-3b6b58:
@maffoo restarted sbt, ran server/run errored with following traceback:

```
[error] Exception in thread "main" java.util.NoSuchElementException: None.get
[error]     at scala.None$.get(Option.scala:347)
[error]     at scala.None$.get(Option.scala:345)
[error]     at org.labrad.browser.WebServer.<init>(WebServer.scala:98)
[error]     at org.labrad.browser.WebServer$.main(WebServer.scala:156)
[error]     at org.labrad.browser.WebServer.main(WebServer.scala)
java.lang.RuntimeException: Nonzero exit code returned from runner: 1
    at scala.sys.package$.error(package.scala:27)
[trace] Stack trace suppressed: run last server/compile:run for the full output.
[error] (server/compile:run) Nonzero exit code returned from runner: 1
[error] Total time: 27 s, completed Feb 23, 2016 4:47:09 PM
```

maffoo:
Run `gulp` first. It looks like it's trying to load the build `index.html` file up front and failing because there's no dist version. I'll look into fixing that, but for now that should let you try things out.

joshm-3b6b58:
Run which `gulp` command?

joshm-3b6b58:
So I just ran `gulp` and everything works, but the server's not running in a separate task now, that that what's supposed to happen?

maffoo:
Just `gulp`, nothing else. That runs the default command, which builds the dist version.

joshm-3b6b58:
So what's the status of 'dev mode' then? Is that gone or is that being added back.

maffoo:
Once you've run `gulp` once, you should be able to go back to running a separate node dev server with `gulp serve`, the scala server is just expecting an `index.html` file so you had to do the build once to get everything in place. As I said, I'll fix that so it works after a `gulp clean`, for example.

joshm-3b6b58:
Ah, OK. weird. Thanks.

maffoo:
@joshmutus, I've now fixed the issue with having to run `gulp` before `gulp serve`. If you want to try again you should be able to run `gulp clean && gulp serve`, then restart the scala server and still run everything in dev mode. If you browse to `localhost:7667` directly you'll see a placeholder message.

joshm-3b6b58:
Yup, that worked pretty seamlessly

README.md
```diff
@@ -16,9 +16,9 @@ $ sbt server/run
 ```
 
 After downloading a bunch of packages (these are cached locally, so this will only be slow the first time),
-you'll have a server running on `localhost:9000`. It will try to connect to a labrad manager running on
+you'll have a server running on `localhost:7667`. It will try to connect to a labrad manager running on
 the same machine, so you should start up a labrad manager if you want to see anything interesting.
-Note that the server uses the [play framework](https://www.playframework.com/) which
+Note that the server uses a json library from the [play framework](https://www.playframework.com/) which
```
joshm-3b6b58, line 21: Is it still true that you need java 8 even though we're just importing one library?

client-js/app/scripts/rpc.ts
```diff
@@ -92,6 +92,7 @@ export class JsonRpcSocket {
    * Close the rpc socket by closing the underlying WebSocket.
    */
   close(code?: number, reason?: string) {
+    code = code || 1000;
```
joshm-3b6b58, line 95: This is basically saying: if no code is provided, set code to 1000. Are there other codes? Are you just planning for the future?

client-js/app/elements/labrad-app.ts
```diff
@@ -32,8 +60,12 @@ export class LabradApp extends polymer.Base {
...
 
   doLogout(event) {
-    window.localStorage.removeItem('labrad-credentials');
-    window.sessionStorage.removeItem('labrad-credentials');
+    var key = 'labrad-credentials';
+    if (this.host) {
+      key += '.' + this.host;
+    }
+    window.localStorage.removeItem(key);
+    window.sessionStorage.removeItem(key);
```
joshm-3b6b58, line 68: What's going on here? Are you storing separate session keys for being connected to multiple hosts?

client-js/app/scripts/places.ts
```diff
@@ -1,19 +1,35 @@
 // Function for creating URLs of all the places in the application.
 
-export function grapherUrl(path: Array<string>, dir?: string): string {
-  return '/grapher/' + pathStr(path, dir);
-}
+export class Places {
+  fullPrefix: string;
```
joshm-3b6b58, line 4: So I understand the functions in here for creating URLs but I don't know what the `Places` class is or how it helps? I can see it invokes as a property in all the sub-applications but I don't undestand what's it doing. What's required beyond just taking a path to a resource and generating a URL?

README.md
maffoo, line 21: Unfortunately, yes. The play json library references some classes that were added in java8, so it locks us in to that version. There are probably other json libraries we could swap in, but I haven't looked into it too much.

client-js/app/elements/labrad-app.ts
maffoo, line 68: Yes, the credentials are stored separately for each host, since they could have different logins. Also, we use local and session storage depending on whether you say to remember the password between browser restarts.

client-js/app/scripts/places.ts
maffoo, line 4: The idea is that we create an instance of the `Places` class and then pass that around to any parts of the app that need to generate URLs to link to other places within the app. In order to generate these URLs correctly, you need to know which manager you're talking to, and possibly what url prefix the app is being served under. For example, we might be on a shared host where scalabrad-web is not at the root like `https://shared-host.com/` but rather at `https://shared-host.com/labrad/`. Then all intra-app links need to have that `/labrad` prefix, so we set that when creating the `Places` instance and then the other parts of the app don't need to change to create the correct URLs.

client-js/app/scripts/rpc.ts
maffoo, line 95: Passing a code of 1000 to the [close method](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#close%28%29) just means we're closing the socket normally because we're done with it, not because there was an error. There are lots of other [status codes](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent#Status_codes), but typically we won't need them.

client-js/app/scripts/activities.ts
```diff
@@ -368,13 +375,15 @@ export class ServerActivity implements Activity {
...
-  constructor(private mgrApi: manager.ManagerApi,
+  constructor(private places: Places,
+              private mgrApi: manager.ManagerApi,
               private nodeApi: node.NodeApi) {}
 
   async start(): Promise<ActivityState> {
     var nodesInfo = await this.nodeApi.allNodes();
     var elem = <LabradNodes> LabradNodes.create();
     elem.info = nodesInfo;
+    elem.places = this.places
```
joshm-3b6b58: nit: missing semi colon

client-js/app/scripts/app.ts
```diff
@@ -30,6 +31,51 @@ interface Creds {
...
+ * Make a route with n path segments labeled p<i>.
+ *
+ * On a page change, the path can be extracted using the getPath function.
+ * TODO: figure out a way to use wildcard routes instead of multiple routes
+ * each with a fixed number of segments.
+ */
+function pathRoute(n: number): string {
+  var route = '';
+  for (var i = 0; i < n; i++) {
+    route += ':p' + i + '/';
```
joshm-3b6b58: This looks like you're appending emoji :0

client-js/app/scripts/app.ts
```diff
@@ -67,44 +124,35 @@ window.addEventListener('WebComponentsReady', () => {
...
   // Stop the current activity and start a new one. This allows activities to
   // do cleanup work before their UI is replaced, such as removing event
   // listeners.
-  async function activate(newActivity: Activity): Promise<void> {
+  async function activate(manager: string, newActivityFunc: (Services) => Activity): Promise<void> {
     try {
       if (activity) {
         await activity.stop();
       }
+      if (loggedInTo !== manager) {
```
joshm-3b6b58, line 145: I still haven't gotten used to the triple equals, but should this be `!===`?

joshm-3b6b58:
Everything on the client LGTM beside the one missing semicolon. I've tested the branch so the scala is probably fine.

client-js/app/scripts/activities.ts
maffoo: fixed

client-js/app/scripts/app.ts
maffoo: LOL; I changed this to use string interpolation, which is a bit less whimsical but more readable.

client-js/app/scripts/app.ts
maffoo, line 145: No, just `!==`. Think about the overall number of characters, not the number of equal signs. This will help to remember that `==` and `!=` correspond, as do `===` and `!==`.

PR #157, opened 2016-02-23
Change selKey to selectedItem so edit value dialog renders correctly

joshm-3b6b58:
Super small PR to fix a rendering error I just noticed


maffoo:
LGTM

Issue #148, opened 2016-02-09
Going from grapher to menu - Star/trash above menu


jwenner:
When I am on the grapher page and switch to the menu, the star and trash icons stay on top of the menu. This partially hides the "Manager" option.

Note: Refers to branch operating on matrix-reloaded

joshm-3b6b58:
@maffoo @jwenner 
I think I've fixed this here https://github.com/labrad/scalabrad-web/commit/69154baa5d2150a4eec3e69db7e809199a94cb52 but it's branch off of the maffoo/dev PR. I have insufficient gitfu to know where to go from here. I can continue to review the dev PR and then submit a PR once that's merged?

jwenner:
I think you can do a PR onto the `u/maffoo/dev` branch (e.g., martinisgroup/pyle#869). Right, @maffoo?

maffoo:
Fixed with #161 and deployed

PR #161, opened 2016-03-04
Style improvements and fixes from dev branch

maffoo:
These changes are deployed and in use internally. I'm putting them up in one big PR for now, but please let me know if you'd prefer to have this broken out.

@joshmutus


joshm-3b6b58:
Start/Trash floats above menu 
image

maffoo:
Yeah, my css-fu needs a lot of work. @jwenner also reported this: #148.

joshm-3b6b58:
Ah, I hadn't seen the issue.

joshmutus:
My understanding of ```position: fixed``` is that the thing that is rendered last is rendered on top. Since the drawer panel is rendered before everything else, the buttons appear on top of it...

client-js/app/elements/labrad-grapher.html
```diff
@@ -42,6 +53,27 @@
...
+    }
+    iron-icon.folder {
+      color: #CFBA78;
+    }
+    iron-icon.dataset {
+      color: #209131;
+    }
   </style>
   <template>
     <div id="outer">
```
joshm-3b6b58, line 97: No indentation for new divs here

client-js/app/elements/labrad-app.html
```diff
@@ -52,6 +52,14 @@
...
       height: 100%;
     }
+
+    #location {
+      white-space: nowrap;
+      text-overflow: ellipsis;
+      overflow: hidden;
+      direction: rtl;
+    }
+
```
joshm-3b6b58, line 62: Nit: We have inconsistent whitespacing between css selectors. Most of the other files have no newlines between selectors.

client-js/app/elements/labrad-registry.ts
```diff
@@ -58,6 +60,27 @@ export class LabradRegistry extends polymer.Base {
...
+  listItems(
+    path: Array<string>,
+    dirs: Array<{name: string; url: string}>,
+    keys: Array<{name: string; value: string}>,
+    kick: number
+  ): Array<{name: string; isParent: boolean; isDir: boolean; isKey: boolean; url?: string; value?: string}> {
+    var items = [];
+    if (path.length > 0 && this.places) {
+      var url = this.places.registryUrl(path.slice(0, -1));
+      items.push({name: '..', isParent: true, isDir: false, isKey: false, url: url});
```
joshm-3b6b58, line 72: Thanks for adding this. I like this solution.

client-js/app/elements/labrad-plot.ts
```diff
@@ -181,6 +188,32 @@ export class Plot extends polymer.Base {
     }
   }
 
+  @observe('drawMode2D')
+  drawMode2DChanged(newMode: string, oldMode: string) {
+    switch (newMode) {
+      case 'dots':
+        this.$.dots.style.color = 'black';
+        this.$.rectfill.style.color = '#AAAAAA';
```
joshm-3b6b58, line 196: Can we make the chart background white? I think the grey erodes a lot of the contrast. (I realize this may be leftover from my original D3 thing, but that was just copied from some example)

I think white background with grey gridlines, like what shows up on a google stock chart is pretty nice.

client-js/app/elements/labrad-plot.ts
```diff
@@ -181,6 +188,32 @@ export class Plot extends polymer.Base {
...
+        this.$.vargrid.style.color = '#AAAAAA';
+        break;
+
+      case 'vargrid':
+        this.$.dots.style.color = '#AAAAAA';
+        this.$.rectfill.style.color = '#AAAAAA';
+        this.$.vargrid.style.color = 'black';
+        break;
+    }
+    if (oldMode && newMode !== oldMode) {
```
joshm-3b6b58, line 212: What is happening here?

client-js/app/elements/labrad-plot.ts
```diff
@@ -6,7 +6,7 @@ import {viridisData} from '../scripts/colormaps';
 /**
  * Colors for traces in 1D plots
  */
-var COLOR_LIST = ['red', 'blue', 'yellow', 'green', 'magenta'];
+var COLOR_LIST = ['blue', 'red', '#00CC00', '#DDDD00', '#DD00DD', '#0088DD'];
```
joshm-3b6b58, line 9: Yellow is probably not a great color since it doesn't show up very well. Although perhaps we can chose a complete color scheme later.

client-js/app/elements/labrad-plot.ts
```diff
@@ -511,12 +602,37 @@ export class Plot extends polymer.Base {
         p = this;
 
     // adjust size, position, and color of each data rect
-    this.svg.selectAll('rect.data')
-        .attr('x', (d) => p.xScale(d[0]))
-        .attr('y', (d) => p.yScale(p.yNext[d[1]]))
-        .attr('width', (d) => p.xScale(p.xNext[d[0]]) - p.xScale(d[0]))
-        .attr('height', (d) => Math.abs(p.yScale(p.yNext[d[1]]) - p.yScale(d[1])))
-        .style('fill', (d) => getColor(d[2], zMin, zMax));
+    switch (this.drawMode2D) {
```
joshm-3b6b58, line 608: I haven't looked at this in a long time so I'm not sure where it is, but the rightmost border doesn't show up as black on the plots.

client-js/app/elements/labrad-plot.html
```diff
@@ -38,6 +41,10 @@
       <paper-icon-button icon="tab-unselected" id="rect" on-click="mouseModeZoomRect" title="drag to zoom"></paper-icon-button>
       <paper-icon-button icon="open-with" id="pan" on-click="mouseModePan" title="drag to pan"></paper-icon-button>
       <paper-icon-button icon="home" id="home" on-click="resetZoom" title="zoom out"></paper-icon-button>
+
+      <paper-icon-button icon="av:stop" id="dots" on-click="drawMode2DDots" title="draw with dots"></paper-icon-button>
```
joshm-3b6b58: It would be cool to toggle these disabled if it's just a line plot.

client-js/app/elements/labrad-plot.ts
joshm-3b6b58, line 9: not part of this PR but I just noticed whitespace missing around operands on line 20

client-js/app/elements/labrad-registry.html
```diff
@@ -25,8 +24,8 @@
       height: 100%;
     }
 
-    .iron-selected {
-      background-color: #ddd;
+    tbody tr:nth-child(1n).iron-selected {
+      background-color: #CCCCCC;
```
joshm-3b6b58, line 28: Again the newlines between css selectors. This is probably my fault from way back.

client-js/app/elements/labrad-registry.ts
```diff
@@ -73,7 +96,8 @@ export class LabradRegistry extends polymer.Base {
   selectedItem: string;
 
   _computeSelectedType(selectedIdx: number): string {
-    if (this.dirs && selectedIdx < this.dirs.length) {
+    var ofs = this.path.length > 0 ? 1 : 0;
```
joshm-3b6b58: Is this like the offset for the `...` at the beginning of the list? If so, can we call this `offset`?

client-js/app/elements/labrad-registry.ts
```diff
@@ -104,7 +129,7 @@ export class LabradRegistry extends polymer.Base {
 
     this.$.pendingDialog.close()
     this.$.combinedList.selected = null;
-
+    this.kick++;
```
joshm-3b6b58, line 132: What's getting `kick`ed and why?

client-js/app/index.html
```diff
@@ -19,7 +19,7 @@
   <!-- Add to homescreen for Chrome on Android -->
   <meta name="mobile-web-app-capable" content="yes">
   <meta name="application-name" content="LabRAD">
-  <link rel="icon" sizes="192x192" href="/images/touch/chrome-touch-icon-192x192.png">
+  <link rel="icon" href="/images/labrad.png">
```
joshm-3b6b58, line 22: Yay a logo! :+1:

client-js/app/scripts/app.ts
```diff
@@ -368,9 +400,23 @@ window.addEventListener('WebComponentsReady', () => {
...
-        await loginWithDialog(mgr, host);
+        if (topLevel) {
+          await loginWithDialog(mgr, host);
+        } else {
+          throw e;
+        }
       }
     }
+
+    var id = setInterval(() => {
```
joshm-3b6b58, line 411: I just noticed this pinging being logged on the server side, what's the purpose of this? Is this to see if the manager is present? If so, doesn't the websocket keep track if it's open by itself?

client-js/app/elements/labrad-grapher.html
maffoo, line 97: fixed

client-js/app/scripts/app.ts
maffoo, line 411: I had problems specifically on blacksand where the websocket connections would get closed after 30 seconds or so if there was nothing happening. I presume there is some http proxy on the corp network that causes this. These pings just go to the web server, they don't actually get sent along to labrad, so this just ensures that proxies don't think the websocket connection is idle and try to close it.

client-js/app/index.html
jwenner, line 22: While I like the general idea, I find that being just black and white with the outlines of the L and the R makes this look indistinct and almost like a Chinese character (particularly when viewed at a vertical angle and against the gray background of a tab or Bookmarks bar). I would instead propose separating the L and R and coloring them differently. Perhaps something like that below (feel free to use this if you wish):
labrad2

client-js/app/elements/labrad-registry.ts
maffoo, line 132: This is a hack to get polymer to recompute a computed property. You'll see that `listItems` depends on kick so that when we increment here it gets recomputed. The value of kick doesn't mater, it just needs to change. (We play the same trick in the node interface.) The problem is that if  a computed property depends on array-valued properties, it doesn't seem to get recomputed even if you update those properties with the polymer-provided `splice` function.

Did I mention I don't like databinding? :)

client-js/app/index.html
joshm-3b6b58, line 22: I was going to say before it looks like the new LA clippers logo.
image
 Now With Jim's color scheme it looks like it even more

joshm-3b6b58, line 22: I always though some kind of sine-wave/pipulse looking think would make a nice logo, but I feel this is beyond the scope of this PR

client-js/app/elements/labrad-registry.ts
joshm-3b6b58, line 132: Isn't that what the observe decorator is for? Does that not apply to this case?

client-js/app/scripts/app.ts
joshm-3b6b58, line 411: Oh. Weird.

client-js/app/elements/labrad-registry.ts
maffoo, line 132: Computed properties do observe the properties they depend on; I think the underlying problem is a bug in how polymer handles observing array-valued properties.

client-js/app/index.html
jwenner, line 22: Maybe something like the following (stolen from the Grapher):
labrad3

maffoo, line 22: That icon is actually from the [famfamfam silk icon set](http://www.famfamfam.com/lab/icons/silk/). I'd prefer to stick with polymer icons or stuff we make ourselves.

joshm-3b6b58, line 22: Woah, somebody made and distributed that on purpose? I had just assumed some grad student had haphazardly scrawled some lines using some generic icon editor.

client-js/app/elements/labrad-plot.ts
maffoo, line 196: This style on this line is for disabled toggle buttons, not the plot area. But I will make the plot area white; I think that's a good idea.

client-js/app/elements/labrad-plot.ts
maffoo, line 212: This triggers a redraw when the mode changes. The check on whether `oldMode` itself is non-null is because this observer gets fired when the property is first set, but in that case we don't need to redraw (the mode wasn't changed _from_ a previous mode).

client-js/app/elements/labrad-plot.ts
maffoo, line 9: Fixed whitespace around operands. Agreed on colorschemes. For now I modified this to match what the grapher does, but of course we can improve on that.

client-js/app/elements/labrad-plot.ts
maffoo, line 608: fixed.

joshm-3b6b58:
@maffoo rendering problem on dataset titles

image

There's slight difference between the arrangement divs on 69154ba and what you just pushed.

This is what 69154ba renders like:

image

joshm-3b6b58:
@maffoo Once title rendering bug on dataset parameters is fixed, LGTM.

client-js/app/elements/labrad-plot.html
maffoo: Done.

client-js/app/elements/labrad-registry.html
maffoo, line 28: I prefer to have the newlines to separate things; I find it's easier to read with some whitespace. Of course one can omit whitespace to group closely-related selectors tightly together.

I agree that we're rather inconsistent with formatting here, but I think that's something we can address in subsequent changes.

client-js/app/elements/labrad-registry.ts
maffoo: fixed.

PR #168, opened 2016-03-24
Update scalabrad to 0.6.2

maffoo:
This should fix the "Unable to create child event loop" problems that we've seen on occasion.


server/src/main/scala/org/labrad/browser/StaticResourceHandler.scala
```diff
@@ -73,7 +73,7 @@ class StaticResourceHandler(
 
     // create http response and set headers
     val response = new DefaultFullHttpResponse(HTTP_1_1, OK, Unpooled.wrappedBuffer(bytes))
-    HttpHeaderUtil.setContentLength(response, bytes.length)
+    HttpUtil.setContentLength(response, bytes.length)
```
maffoo, line 76: The version of netty included with scalabrad changed, and this class name changed with it.

joshm-3b6b58:
Checked out branch and tested with binary of latest scalabrad off bintray. LGTM

PR #171, opened 2016-04-06
Fix bower.json to have compatible polymer packages

joshm-3b6b58:
This way we have the last polymer elements and everything works form a fresh clone

fixes #170 and #158

PR #172, opened 2016-04-06
156 long keys

### LINKED ISSUE #156: Registry: Impossible to edit long keys
When a registry key is so long that it overflows the page, there is no way to get to the button to save the key. There should probably be a scroll bar in the edit window in such a case instead of just overflowing the edit window.

As an example, see the key couplingPts at registry/Jim/AliceBobRes/Cooldown151230/160222/Bc/:
image

joshm-3b6b58:
This is based on the other pull request. Requires a `bower install` to get the `<paper-dialog-scrollable>` if you don't already have it.

Fixes #156

### DIFF
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -14,6 +14,7 @@
 <link rel="import" href="../bower_components/paper-tabs/paper-tabs.html">
 <link rel="import" href="../bower_components/paper-drawer-panel/paper-drawer-panel.html">
 <link rel="import" href="../bower_components/paper-dialog/paper-dialog.html">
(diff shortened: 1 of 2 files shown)


joshm-3b6b58:
@jwenner I don't know if you're set up to test this.

jwenner:
@joshmutus, I don't believe that I can test this, given that all the computers I use run Windows. What I'm doing is to use the version on matrix-reloaded.

joshm-3b6b58:
Well, here's a picture for posterity:
image

maffoo:
LGTM; thanks for fixing this! Let's get the version stuff in #171 and/or #175 fixed and then rebase this before merging.

maffoo:
#175 is merged, so should be good to rebase this on master.

PR #175, opened 2016-04-14
Update versions in bower.json

maffoo:
Based on #171, but uses `1.x` notation for package versions instead of `~` and `^` (for reference, the version range syntax is defined [here](https://github.com/npm/node-semver#advanced-range-syntax)). Also, drop "resolutions" from bower.json, get rid of the unused platinum elements, and update typescript.


maffoo:
@joshmutus

joshm-3b6b58:
I ran `npm install` after checking out this branch. Now I have this strange error in the browser console:

``````

:3000/sw-import.js?baseURI=http%3A%2F%2Flocalhost%3A3000%2Fbower_components…2Fbootstrap%2Fsw-toolbox-setup.js&precache=&skipWaiting=true&version=1.0:1 GET http://localhost:3000/sw-import.js?baseURI=http%3A%2F%2Flocalhost%3A3000%2F…w%2Fbootstrap%2Fsw-toolbox-setup.js&precache=&skipWaiting=true&version=1.0 net::ERR_FILE_EXISTS```

But the app seems to largely work...
``````

joshm-3b6b58:
I opened this in an incognito tab and the error dissapears so this is probably something funny wiht my browser cache

joshm-3b6b58:
LGTM

Issue #116, opened 2015-12-21
Grapher: Add ability to plot one or some of dependent variables


jwenner:
Right now, there is no ability to plot only selected dependent variables. It would be nice if one could select to plot only one or only a selected set of dependent variables (as was possible with the Delphi Grapher).

joshm-3b6b58:
Fixed https://github.com/labrad/scalabrad-web/pull/176

jwenner:
@joshmutus, this doesn't seem to be fixed with the version on matrix-reloaded. Does this mean we should bump the version number and deploy?

PR #176, opened 2016-05-09
116 dependent

joshm-3b6b58:
Adds the ability to select which plots will be shown

Click the button on the middle right, below the reset zoom (home button)
image
Select which traces you want to see:
image
Press OK:
image
Will implement  ALL or NONE for asssiting in selecting on large datasets. For now buttons are disabled

@DanielSank @jwenner @maffoo

### DIFF
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -26,6 +26,8 @@
 <link rel="import" href="../bower_components/paper-material/paper-material.html">
 <link rel="import" href="../bower_components/paper-menu/paper-menu.html">
 <link rel="import" href="../bower_components/paper-styles/paper-styles-classes.html">
+<link rel="import" href="../bower_components/paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../bower_components/paper-radio-button/paper-radio-button.html">
 <link rel="import" href="../bower_components/paper-tabs/paper-tabs.html">
 <link rel="import" href="../bower_components/paper-toolbar/paper-toolbar.html">
 <link rel="import" href="../bower_components/paper-toast/paper-toast.html">
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -13,6 +13,7 @@
 <link rel="import" href="../bower_components/paper-button/paper-button.html">
 <link rel="import" href="../bower_components/paper-tabs/paper-tabs.html">
 <link rel="import" href="../bower_components/paper-drawer-panel/paper-drawer-panel.html">
+<link rel="import" href="../bower_components/paper-checkbox/paper-checkbox.html">
 <link rel="import" href="../bower_components/paper-dialog/paper-dialog.html">
 <link rel="import" href="../bower_components/paper-dialog-scrollable/paper-dialog-scrollable.html">
 <link rel="import" href="../bower_components/paper-header-panel/paper-header-panel.html">
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -13,7 +13,6 @@
 <link rel="import" href="../bower_components/paper-button/paper-button.html">
 <link rel="import" href="../bower_components/paper-tabs/paper-tabs.html">
 <link rel="import" href="../bower_components/paper-drawer-panel/paper-drawer-panel.html">
-<link rel="import" href="../bower_components/paper-checkbox/paper-checkbox.html">
 <link rel="import" href="../bower_components/paper-dialog/paper-dialog.html">
 <link rel="import" href="../bower_components/paper-dialog-scrollable/paper-dialog-scrollable.html">
 <link rel="import" href="../bower_components/paper-header-panel/paper-header-panel.html">
@@ -26,8 +25,6 @@
 <link rel="import" href="../bower_components/paper-material/paper-material.html">
 <link rel="import" href="../bower_components/paper-menu/paper-menu.html">
 <link rel="import" href="../bower_components/paper-styles/paper-styles-classes.html">
-<link rel="import" href="../bower_components/paper-radio-group/paper-radio-group.html">
-<link rel="import" href="../bower_components/paper-radio-button/paper-radio-button.html">
 <link rel="import" href="../bower_components/paper-tabs/paper-tabs.html">
 <link rel="import" href="../bower_components/paper-toolbar/paper-toolbar.html">
 <link rel="import" href="../bower_components/paper-toast/paper-toast.html">
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -54,7 +54,11 @@
       <form is="iron-form" id="formy">
       <template is="dom-repeat" items="{{deps}}">
         <div>
-          <paper-checkbox checked name="foo">{{index}}</paper-checkbox> <span>{{item.label}}</span>
+          <!-- The 0 in front on {{index}} is a hack around a stupid javascript thing 
+          where value="0" makes the checkbox not exist, value="00" counts as a label :/ -->
+          <paper-checkbox checked name="traces" value="0{{index}}">{{index}}</paper-checkbox>
+          <span> - </span>
+          <span>{{item.label}}</span>
         </div>
       </template>
       <paper-button on-click="submitTraces">OK</paper-button>
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -56,7 +56,7 @@
           <template is="dom-if" if="{{oneIndep}}">
             <template is="dom-repeat" items="{{deps}}">
               <div>
-                <!-- The 2 in front on {{index}} is a hack around a stupid javascript thing 
+                <!-- The 's' in front on {{index}} is a hack around a stupid javascript thing 
                 where value="0" makes the checkbox not exist, value="s0" counts as a label :/ -->
                 <paper-checkbox checked name="traces" value="s{{index}}">{{index}}</paper-checkbox>
                 <span> - </span><span>{{item.label}}</span>
@@ -70,6 +70,8 @@
             </div>
           </template>
           <template is="dom-if" if="{{!oneIndep}}">
+            <!-- The 's' in front on {{index}} is a hack around a stupid javascript thing 
+            where value="0" makes the checkbox not exist, value="s0" counts as a label :/ -->
             <paper-radio-group selected="s0">
               <template is="dom-repeat" items="{{deps}}">
                 <paper-radio-button name="s{{index}}">
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -51,7 +51,8 @@
     
     <paper-dialog id="traceSelector">
       <h1>Select Traces</h1>
-      <form is="iron-form" id="selectForm">
+      <paper-dialog-scrollable>
+        <form is="iron-form" id="selectForm">
           <template is="dom-repeat" items="{{deps}}">
             <div>
               <!-- The 2 in front on {{index}} is a hack around a stupid javascript thing 
@@ -61,16 +62,18 @@
               <span>{{item.label}}</span>
             </div>
           </template>
+        </form>
+      </paper-dialog-scrollable>
+        <div id="submitButtons" style="display:inline">
+          <paper-button raised on-click="submitTraces">OK</paper-button>
+          <paper-button raised dialog-dismiss>cancel</paper-button>
+        </div>
         <div id="selectButtons">
         <!-- TODO implement these  -->
           <paper-button disabled>ALL</paper-button>
           <paper-button disabled>NONE</paper-button>
         </div>
-        <div id="submitButtons" style="display:inline">
-          <paper-button raised on-click="submitTraces">OK</paper-button>
-          <paper-button raised dialog-dismiss>cancel</paper-button>
-        </div>
-      </form>
+      
     </paper-dialog>
   </template>
 </dom-module>
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -58,7 +58,7 @@
               <div>
                 <!-- The 's' in front on {{index}} is a hack around a stupid javascript thing 
                 where value="0" makes the checkbox not exist, value="s0" counts as a label :/ -->
-                <paper-checkbox checked name="traces" value="s{{index}}">{{index}}</paper-checkbox>
+                <paper-checkbox checked name="traces" data-index="{{index}}">{{index}}</paper-checkbox>
                 <span> - </span><span>{{item.label}}</span>
                 <span> - </span>
                 <span>{{item.legend}}</span> [<span>{{item.unit}}</span>]
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -52,19 +52,19 @@
     <paper-dialog id="traceSelector">
       <h1>Select Traces</h1>
       <form is="iron-form" id="selectForm">
-        <template is="dom-repeat" items="{{deps}}">
-          <div>
-            <!-- The 2 in front on {{index}} is a hack around a stupid javascript thing 
-            where value="0" makes the checkbox not exist, value="s0" counts as a label :/ -->
-            <paper-checkbox checked name="traces" value="s{{index}}">{{index}}</paper-checkbox>
-            <span> - </span>
-            <span>{{item.label}}</span>
-          </div>
-        </template>
+          <template is="dom-repeat" items="{{deps}}">
+            <div>
+              <!-- The 2 in front on {{index}} is a hack around a stupid javascript thing 
+              where value="0" makes the checkbox not exist, value="s0" counts as a label :/ -->
+              <paper-checkbox checked name="traces" value="s{{index}}">{{index}}</paper-checkbox>
+              <span> - </span>
+              <span>{{item.label}}</span>
+            </div>
+          </template>
         <div id="selectButtons">
-        <!--  -->
-          <paper-button disabled label="TODO">ALL</paper-button>
-          <paper-button disabled label="TODO">NONE</paper-button>
+        <!-- TODO implement these  -->
+          <paper-button disabled>ALL</paper-button>
+          <paper-button disabled>NONE</paper-button>
         </div>
         <div id="submitButtons" style="display:inline">
           <paper-button raised on-click="submitTraces">OK</paper-button>
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -58,7 +58,7 @@
           <template is="dom-if" if="{{is1D}}">
             <template is="dom-repeat" items="{{deps}}">
               <div>
-                <paper-checkbox checked name="traces" data-index="{{index}}">{{index}}</paper-checkbox>
+                <paper-checkbox checked name="traces" trace-index="{{index}}">{{index}}</paper-checkbox>
                 <span> - </span><span>{{item.label}}</span>
                 (<span>{{item.legend}}</span>) [<span>{{item.unit}}</span>]
               </div>
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -36,7 +36,7 @@
   </style>
   <template>
     <div id="plot"></div>
-    <div id="pos">Current position: <span>{{currPos}}</span></div>
+    <div id="pos">current position: <span>{{currPos}}</span></div>
     <div id="buttons">
       <paper-icon-button icon="tab-unselected" id="rect" on-click="mouseModeZoomRect" title="drag to zoom"></paper-icon-button>
       <paper-icon-button icon="open-with" id="pan" on-click="mouseModePan" title="drag to pan"></paper-icon-button>
@@ -71,8 +71,8 @@
       </div>
       <div id="selectButtons">
       <!-- TODO implement these  -->
-        <paper-button on-click="selectAll">ALL</paper-button>
-        <paper-button on-click="diselectAll">NONE</paper-button>
+        <paper-button on-click="selectAll">select all</paper-button>
+        <paper-button on-click="deselectAll">select none</paper-button>
       </div>
     </paper-dialog>
 
--- a/client-js/app/elements/labrad-plot.html
+++ b/client-js/app/elements/labrad-plot.html
@@ -53,15 +53,32 @@
       <h1>Select Traces</h1>
       <paper-dialog-scrollable>
         <form is="iron-form" id="selectForm">
-          <template is="dom-repeat" items="{{deps}}">
-            <div>
-              <!-- The 2 in front on {{index}} is a hack around a stupid javascript thing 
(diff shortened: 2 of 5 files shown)


jwenner:
While I'd like to be able to test this, all the computers I have access to are Windows computers. Could you please do a setup on a Linux computer? (Either that or I can't test this). Note I'd like to see how the selection menu handles long lists of dependent variables (see any of the DAC ScanPhases datasets for instance - example either at ['','Test','GHzFPGA','Ivan DAC 59','151030','0_0',3] or see Austin for a dataset at Google). I'd also like to see what you do in the case of a 2D dataset, where the menu icon is already used for "draw with vargrid"

But overall, I like the idea as seen in the photos you provided. Don't know enough to comment on the code, though.

joshm-3b6b58:
Thanks for taking a look Jim. I don't know how to deploy this branch on Matrix-reloaded for testing so I'll wait until this gets through review. If you have any suggestions from what you can see of the GUI in the screenshots, let me know

jwenner:
@joshmutus, could you please post screenshots of the following:
1. What happens to the menu when there's a lot of dependent variables. See the GHzDAC ScanPhases I mentioned above for examples.
2. What happens to the buttons in the case of a 2D dataset? Example: ['','Jim','AliceBobRes','Cooldown160406','160503','7-T1 Stability'].
3. What is the text which appears when you hover the mouse over the button?
4. What happens when the label of the traces does not match that of the first trace? For the example you showed, what happens to the y-axis label if you selected traces 1 and 3 (i.e., the two phases)?

joshm-3b6b58:
> 1. What happens to the menu when there's a lot of dependent variables. See the GHzDAC ScanPhases >I mentioned above for examples.

I hadn't thought of this. I'll make a it a scrollable dialog so they'll all show up.

> 1. What happens to the buttons in the case of a 2D dataset? Example: 
>    ['','Jim','AliceBobRes','Cooldown160406','160503','7-T1 Stability'].

Nothing. Not implemented yet. What do we want the behavior to be? I forget can we have multiple deps for 2 indeps? It's been a LONG time since I've taken data :/

> 1. What is the text which appears when you hover the mouse over the button?

"select trace"

> 1. What happens when the label of the traces does not match that of the first trace? For the example you > showed, what happens to the y-axis label if you selected traces 1 and 3 (i.e., the two phases)?

Good catch. Not the correct thing. I'll fix that.

joshm-3b6b58:
@jwenner So I've changed the behavior for the axis label for selecting traces.

If you have multiple traces with multiple labels and units, the axis label will take on the value of the lowest order of the selected trace. 

For example if I have
1 - magnitude
2 - phase
3 - magnitude
4 - phase

and I select traces 2 and 4, the label will show phase (with the legend and units fields filled in)
If I select traces 1 and 3 it will label magnitude. 
If I select traces 3 and 4 it will label magnitude
If I select traces 2 and 3 it will label phase

I worry displaying lots of axis labels will get cluttered, so I'm just displaying one right now. Is that behavior good for now?

Screenshot for posterity (Yes I know plot isn't re-scaling, I'm following the breadcrumbs on that now) :
image

jwenner:
Seems reasonable to me. I think this is akin to what the old Delphi grapher does (but I don't know the particular ordering that it uses). But you should definitely leave a comment in the code to this effect.

Long term, I don't know what a good solution is. One conceivable idea is to only plot traces with the same label as the first selected trace? Or to show all labels but color-coded? Or something else? Maybe @DanielSank or @maffoo has an idea. 

But as for now, I think that what you propose for the axis label should work.

joshm-3b6b58:
Also, just fixed the dialog so it will be scrollable for lots of dependents. I don't have a dataset locally with very many dependents, so I can't post a nice screenshot, but it works if I scrunch the screen up to be super duper small :)

jwenner:
Since at least Austin has done the DAC ScanPhases script at Google, you all probably have a dataset with lots of dependents there. Typical location is something like `['','Test','GHzDAC', boardName,...]`.

joshm-3b6b58:
I actually have no idea how to deploy on that machine either :(

joshm-3b6b58:
Plot limits now change with selected traces:

All Traces
image
Phase:
image

Mag:
image

joshm-3b6b58:
Thanks for the UI feedback @jwenner ! Very handy peer review even without knowing the code.

jwenner:
From what you've said and the screenshots you've posted, a provisional LGTM (provisional based on not having tried it myself and not knowing the code). Although, we should either not close #116 or raise a new issue until this is made available in the 2D scan case as well.

jwenner:
Thinking about it, one additional idea. On the "Select Traces" menu, would it be good to show both the label and legend (so "Magnitude S11" instead of "Magnitude")? That would at least provide a way to distinguish between two magnitudes, probabilities, etc.

Also, do the traces show up in the same order between the "Select Traces" menu and the list of dependent variables on the page which shows datasets/dependents/parameters?

joshm-3b6b58:
Sure that's easy to add

joshm-3b6b58:
added the thing. 

I don't really feel comfortable merging before someone takes a look at the code. Also, even if it gets merged I don't know how to deploy to matrix-reloaded. 

@DanielSank @ejeffrey do you know how to build and deploy on matrix reloaded?

client-js/app/elements/labrad-plot.html
```diff
@@ -70,8 +71,8 @@
       </div>
       <div id="selectButtons">
       <!-- TODO implement these  -->
-        <paper-button disabled>ALL</paper-button>
-        <paper-button disabled>NONE</paper-button>
+        <paper-button on-click="selectAll">ALL</paper-button>
```
jwenner: What do these buttons do? Do they plot all traces or just select all traces in the menu? If the latter, then they should be "Select ALL" and "Select NONE".

client-js/app/elements/labrad-plot.html
```diff
@@ -70,8 +71,8 @@
...
+        <paper-button on-click="diselectAll">NONE</paper-button>
```
jwenner: Do you mean deselectAll instead?

jwenner:
See my comments above about the ALL and NONE buttons. Don't know why they show up as outdated diffs...

DanielSank:
@joshmutus off the top of my head, I do not.

client-js/app/elements/labrad-plot.html
joshm-3b6b58: Sure, I guess it's more clear that way.

client-js/app/elements/labrad-plot.html
joshm-3b6b58: Aha, good catch

joshm-3b6b58:
2D datasets are now selectable using some radio buttons:
image
image
There's a small but in the rendering of that dialog, but once this bit gets reviews I think it's safe to close that issue.

A nice feature to add would be if selecting traces updated the URL and if traces could be selected via URL. This can be done in principle by setting `this.displayTraces` array from the URL

jwenner:
Still the same provisional LGTM (provisional based on not having tried it myself and not knowing the code).

joshm-3b6b58:
Thanks Jim. I want to get @maffoo's take on this. This waited 6mo, it can wait another week, I think.

client-js/app/elements/elements.html
```diff
@@ -25,6 +26,8 @@
 <link rel="import" href="../bower_components/paper-material/paper-material.html">
 <link rel="import" href="../bower_components/paper-menu/paper-menu.html">
 <link rel="import" href="../bower_components/paper-styles/paper-styles-classes.html">
+<link rel="import" href="../bower_components/paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../bower_components/paper-radio-button/paper-radio-button.html">
```
maffoo: I'd prefer to import these where they are used, instead of in this global file. That is, put these imports in `labrad-plot.html` directly. It's unfortunate that we inherited this bad habit from the polymer example project...

client-js/app/elements/labrad-plot.ts
```diff
@@ -103,6 +104,9 @@ export class Plot extends polymer.Base {
 
   @property({type: String, value: 'zoomRect'})
   mouseMode: string;
+  
+  @property({type: Object, value: ''})
```
maffoo: Remove `value`. Also, isn't this actually set with an Array of axis descriptors, not an Object?

client-js/app/elements/labrad-plot.ts
```diff
@@ -103,6 +104,9 @@ export class Plot extends polymer.Base {
...
+  deps: any;
```
maffoo: We should have a real typescript type here.

client-js/app/elements/labrad-plot.ts
```diff
@@ -498,11 +514,10 @@ export class Plot extends polymer.Base {
         }
       }
 
-      for (let i = 2; i < row.length; i++) {
-        let z = row[i];
+        let z = row[this.displayTraces[0]+2];
```
maffoo: pull out `this.displayTraces[0] + 2` into a named variable and use it here and below for more DRYness.

client-js/app/elements/labrad-plot.ts
```diff
@@ -498,11 +514,10 @@ export class Plot extends polymer.Base {
...
         this.dataLimits.zMin = safeMin(this.dataLimits.zMin, z);
         this.dataLimits.zMax = safeMax(this.dataLimits.zMax, z);
-      }
+      //}
```
maffoo: ?

client-js/app/elements/labrad-plot.ts
```diff
@@ -562,11 +578,54 @@ export class Plot extends polymer.Base {
...
       break;
     }
   }
+  
+  private selectTraces() {
+    this.$.traceSelector.open();
+  }
+
+  private selectAll() {
+    var checkboxes: any = document.getElementsByName('traces');
```
maffoo: Touching `document` breaks the encapsulation provided by polymer. You should use polymer's [dom api](https://www.polymer-project.org/1.0/docs/devguide/local-dom.html) instead, something like:

``` typescript
var checkboxes = Polymer.dom(this.$.traceSelector).querySelectorAll('traces')
```

client-js/app/elements/labrad-plot.ts
```diff
@@ -562,11 +578,54 @@ export class Plot extends polymer.Base {
...
+      checkboxes[i].checked = false;
+    }
+  }
+
+  private submitTraces() {
+    var selected: Array<string> = [];
+    var serialized: Array<string>;
+    switch (this.numIndeps) {
+      case 1:
+        serialized = this.$.selectForm.serialize().traces;
```
maffoo: Instead of serializing the form, I would suggest getting the checkbox list as in the `selectAll` method above and then reading the `checked` property.

client-js/app/elements/labrad-plot.html
```diff
@@ -41,12 +41,55 @@
...
+    <paper-dialog id="traceSelector">
+      <h1>Select Traces</h1>
+      <paper-dialog-scrollable>
+        <form is="iron-form" id="selectForm">
+          <template is="dom-if" if="{{oneIndep}}">
+            <template is="dom-repeat" items="{{deps}}">
+              <div>
+                <!-- The 's' in front on {{index}} is a hack around a stupid javascript thing 
+                where value="0" makes the checkbox not exist, value="s0" counts as a label :/ -->
+                <paper-checkbox checked name="traces" value="s{{index}}">{{index}}</paper-checkbox>
```
maffoo: maybe use a different attribute, like `data-index` to avoid this quirk of the `value` property.

client-js/app/elements/labrad-plot.html
```diff
@@ -41,12 +41,55 @@
...
+          </template>
+          <template is="dom-if" if="{{!oneIndep}}">
+            <!-- The 's' in front on {{index}} is a hack around a stupid javascript thing 
+            where value="0" makes the checkbox not exist, value="s0" counts as a label :/ -->
+            <paper-radio-group selected="s0">
+              <template is="dom-repeat" items="{{deps}}">
+                <paper-radio-button name="s{{index}}">
+                  <span>{{item.label}}</span>
+                  <span> - </span>
+                  <span>{{item.legend}}</span> [<span>{{item.unit}}</span>]
```
maffoo: The data vault formats these like `label (legend) [unit]`. I think we should follow that here as well.

server/src/test/scripts/data_demo.py
```diff
@@ -66,6 +66,20 @@ def demo_2d_simple(dv):
...
+def demo_2d_multi(dv):
+    dv.new('demo_2d_multi', ['x [GHz]', 'y [V]'],
+	   ['z1 [a.u.]', 'z2 [a.u]'])
+    add_params(dv)
+    for i in xrange(50):
+        x = i / 10
+        for j in xrange(50):
+            y = j / 10
+            z1 = 5 * math.cos((x**2 + y**2) / 10) + random.random() - 0.5
+	    z2 = 5 * 0
```
maffoo: there be tabs here!

client-js/app/elements/labrad-plot.ts
```diff
@@ -133,6 +138,9 @@ export class Plot extends polymer.Base {
   private y0: number = null;
   private dx0: number = -1;
   private dy0: number = -1;
+  private displayTraces: Array<number>;
+  private allOrNone: boolean = true;
+  private oneIndep: boolean;
```
maffoo: I'd prefer to call this `is1D` or something like that. Can also have an `is2D` property for 2D data sets. There are other things we may need to change in the UI based on the dimensionality, so these are useful properties to have.

client-js/app/elements/labrad-plot.ts
maffoo: It requires a bit of casting since the return type is a collection of generic nodes, but this seems to work:

``` typescript
var checkboxes = Polymer.dom(this.$.traceSelector).querySelectorAll('[name=traces]');
for (let checkbox of checkboxes) {
  (<HTMLInputElement>checkbox).checked = true;
}
```

maffoo:
I have a few comments about the code but mostly this looks good. However, it doesn't actually work for me. After I select some traces the `displayTraces` array is a bunch of `NaN`s, so the plotting fails. Haven't tracked down the problem, but this will all work differently after changing the `submitTraces` to not use form serialization anyway, so I didn't look too hard. I agree it would be nice to make it possible to reflect these kind of changes in the url so that it can be shared, but I think that could happen in a separate PR.

joshm-3b6b58:
Thanks for the comments. I'll get to those right away. 

I'll change the serialization and push that and see if your problem persists. Did you try plotting different 2D datasets?

server/src/test/scripts/data_demo.py
joshm-3b6b58: That's what I get for being too lazy to import the server stuff into my eclipse workspace :/

maffoo:
I only tried with 1D datasets so far, then ran into the problem.

client-js/app/elements/labrad-plot.ts
joshm-3b6b58: Oh yeah... there's a chronology to how I figured out all of this, heh.

joshm-3b6b58:
@maffoo can you take a look when you get a chance. I just fixed the way the checkboxes are read. It should log what traces are selected and which are added to `this.displayTraces` let me know if it works now.

maffoo:
@joshmutus, works, though I get a typescript compiler error on lines 611 and 612. I think you need to cast to the appropriate element type:

``` typescript
if ((<HTMLInputElement>checkbox).checked) {
  selected.push((<HTMLInputElement>checkbox).dataIndex);
}
```

maffoo:
BTW, I wish there were a better way to do that...

joshm-3b6b58:
Cool. I noticed the errors, just wanted to make sure it's functional.

client-js/app/elements/elements.html
joshm-3b6b58: Ah yes, I remember now. Fixed.

client-js/app/elements/labrad-plot.ts
joshm-3b6b58: True. I get confused as to how javascript (and typescript) call "objects" when I think of them as dictionaries. I would think that deps is an array of dictionaries, but Array seems fine here.

client-js/app/elements/labrad-plot.ts
joshm-3b6b58: You mean `deps: Array;`  ?

client-js/app/elements/labrad-plot.ts
joshm-3b6b58: oops

client-js/app/elements/labrad-plot.ts
joshm-3b6b58: done

client-js/app/elements/labrad-plot.ts
joshm-3b6b58:
```
  @property({type: Array})
  deps: Array<{label: string, legend: string, unit: string}>;
```

?

client-js/app/elements/labrad-plot.html
joshm-3b6b58: yup

client-js/app/elements/labrad-plot.ts
joshm-3b6b58: done

joshm-3b6b58:
Pushed to the branch addressing most of the comments. I'm having trouble casting the radio button types:

```
app/elements/labrad-plot.ts(617,56): error TS2339: Property 'dataIndex' does not exist on type 'HTMLElement'.
app/elements/labrad-plot.ts(623,58): error TS2339: Property 'selected' does not exist on type 'HTMLElement'.
```

AFAICT there are no HTMLElements with the properties selected or dataIndex, those come from IronSelectableBehavior, which isn't included in the polymer-ts typings. Should I just make that typing manually? Where do I put it?

joshm-3b6b58:
So I just renamed  `tab-index` because it exists on [`<HTMLInputElement>`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) and is of the correct type. Is that reasonable?

maffoo:
I don't think you want to use `tab-index` to just store some data, since that is meant for controlling tabbing in the UI. It worked fine for me to use `dataIndex` and `checked` with a cast to `<HTMLInputElement>` as in the snippet I posted earlier.

joshm-3b6b58:
@maffoo 
Data Index doesn't work for me on `<HTMLInputElement>``

```
app/elements/labrad-plot.ts(617,56): error TS2339: Property 'dataIndex' does not exist on type 'HTMLInputElement'.
```

joshm-3b6b58:
So, this quiets the compiler and allows me to create my own custom properties:

I made a new class: 

```
class CustomSelector extends HTMLElement {
  //Is this a hack?
  selected: string;
  dataIndex: number;
}
```

And I call 
`(<CustomSelector>radio).selected` and `(<CustomSelector>checkbox).dataIndex`

Is that a total hack?

client-js/app/elements/labrad-plot.ts
joshm-3b6b58: done

maffoo:
Technically, DOM attributes that start with `data-` are supposed to be accessed in javascript through the [`dataset` field](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes), so this should be

``` javascript
selected.push((<HTMLInputElement>checkbox).dataset.index);
```

Does that work for you? I tried that and while the typescript compiler accepted it, it didn't work at run time. I'm just curious if that works for you. I'm guessing polymer is doing something non-standard with data attributes. Also, it would appear that you and I have different typescript and/or other library versions and maybe that is the source of the different behavior we're seeing here.

Rather than create a class just to cast to it, I would suggest just casting to `<any>` and adding a comment with a TODO to come back and investigate this further.

joshm-3b6b58:
That neither works for me at runtime or compiles cleanly:

```
app/elements/labrad-plot.ts(617,64): error TS2339: Property 'index' does not exist on type 'DOMStringMap'
```

maffoo:
Ok, I figured out how to make this work with `data-` attributes.

First, you have to use "explicit attribute binding" in labrad-plot.html, as described [here](https://github.com/Polymer/polymer/issues/1737):

``` html
<paper-checkbox checked name="traces" data-index$="{{index}}">{{index}}</paper-checkbox>
```

Then, you need to use square-bracket indexing to get at items inside the `dataset` object in labrad-plot.ts (the type definiton for `DOMStringMap` is [here](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L6257)):

``` typescript
selected.push(Number(checkbox.dataset['index']));
```

All that being said, I'm not sure using a `data-` attribute like this is really worth it given that polymer already has mechanisms to pass things through the DOM to javascript.

How about this: use a new attribute name, say, `trace-index` and then just cast to `<any>` to make typescript happy:

in labrad-plot.html:

``` html
<paper-checkbox checked name="traces" trace-index="{{index}}">{{index}}</paper-checkbox>
```

in labrad-plot.ts

``` typescript
selected.push((<any>checkbox).traceIndex);
```

Sorry for all the back and forth on this. I figure if we work out a reasonable pattern here we can just use it in the future. Does this latter proposal seem reasonable to you?

joshm-3b6b58:
Thanks for figuring this out. Totally cool with the back and forth, I'm glad we can find a way to do this properly as I think we'll be having to expand on this feature in the near future for dataset specific plotting. I'll add all this as soon as I can.

One question, that DOMStringMap typing... should that already be in typescript? Do I have to upgrade it or do I have to add this typing manually?

maffoo:
That DOMStringMap definition is in `lib.d.ts` which is bundled with typescript. Basically, it provides typescript definitions for all of the built-in browser APIs. You can find it inside `node_modules` if you're curious, but there's nothing you need to do to use these definitions.

joshm-3b6b58:
Alright. I think I made all the changes you commented on. Please take a quick look when you get the chance.

Thanks for the input!

maffoo:
Choosing a trace on a 2D plot does not work for me. If I choose trace 0, the console log shows `[NaN]` for selected traces and the plotting does not work. If I choose trace 1, the console log shows `[1]` and it does seem to plot the new trace, although the coloring is wrong; it seems the z data limits need to be recomputed in that case.

joshm-3b6b58:
Hm.

You're on the latest commit of the branch? One of the earlier commits broke
2D trace selection.

At least the color scale should recompute each time...

maffoo:
Yeah, I'm on the latest commit. The radio buttons display strange behavior as well; when the dialog opens neither is selected, and I can select both or neither options, which I wouldn't expect with a radio button.

joshm-3b6b58:
I've seen all those problems in the past, but I'm not seeing them on my end. You sure you're on     68c0177?? 

Maybe polymer components are out of date? Do a bower install?

maffoo:
After fiddling with bower packages, it works. LGTM

joshm-3b6b58:
Cool, thanks for figuring that out. It's a bit troubling this is fragile in regards to the bower packages... 

How do we feel about squashing and merging through the web interface on this repo?

maffoo:
Merging through the web interface is fine by me. I just changed the repo settings to only allow squash merges.

joshm-3b6b58:
Shoot, forgot to say this fixes 116

PR #186, opened 2016-06-14
Use base tag in index.html to allow for relative urls

maffoo:
We send index.html to the browser to bootstrap the app no matter what route within the app the user loads. Hence, to use relative urls we must update the base against which urls in the index page are resolved to use the root path where the app is hosted, not the route from which the page was served.

On the server, we modify the base href appropriately before serving the page by manipulating the page DOM (previously this was done by injecting javascript, but the DOM manipulation is much cleaner). We also fix up the way this is handled on the client-side to be consistent with how the browser itself resolves relative URLs.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,6 +27,7 @@ lazy val server = project.in(file("server"))
 
     libraryDependencies ++= Seq(
       "org.scala-lang.modules" %% "scala-async" % "0.9.2",
+      "org.jsoup" % "jsoup" % "1.9.2",
       "org.labrad" %% "scalabrad" % "0.6.2"
     ),
 
--- a/client-js/app/index.html
+++ b/client-js/app/index.html
@@ -7,11 +7,14 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>LabRAD</title>
 
+  <!-- Resolve relative URLs against the root path, not the current route path -->
+  <base href="/">
+
   <!-- Chrome for Android theme color -->
   <meta name="theme-color" content="#303F9F">
 
   <!-- Web Application Manifest -->
-  <link rel="manifest" href="/manifest.json">
+  <link rel="manifest" href="manifest.json">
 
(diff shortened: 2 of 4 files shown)


maffoo:
@joshmutus

joshm-3b6b58:
Hey @maffoo can you provide me an example of how this worked before and how this works now? I'm having a little trouble understanding this from the PR description. 

For example when I request `https://SERVER/registry/Jim/AliceBobRes/Cooldown151230/160222/Bc/` what is happening different than before?

maffoo:
With the URL you provided, nothing is different. The idea here is I might have a single webserver where I want to host two copies of scalabrad-web (say, one running the latest release and one running a dev version for testing). I want to host the release version at the root url path so that people will see it by default when going to, say `https://my.server/registry/...`. But I also want to host the dev version at another path like `/dev/` so people can see it by going to `https://my.server/dev/registry/...`. The change here makes this possible. You would just have to run the dev server with the option `--url-prefix=/dev/`. (You would also have to run the dev server on a different port and set up a proxy server to forward the paths appropriately, but that is a different story).

The reason this didn't work before is that the `index.html` page had absolute URLs for things like javascript and html imports, so even if you loaded the initial page from the dev server on `https://my.server/dev/registry/...`, the browser would then load additional resources like javascript from the absolute URLs and get the versions from the server running on the root path.

Does that help?

joshm-3b6b58:
@maffoo OK, I understand that, how do I test it? Say I have two builds I'm running off the same server what uh, do I do with gulp to run them both at the same time?

joshm-3b6b58:
I'm assuming there are commits to follow that deal with handling and labelling the versioning of the dev servers? It looks good. I'm able to bounce back and forth between the two versions of the software. I'm curious how I'll go about deploying my own dev-servers etc, it would be great to go through some together and create some instructions.

joshm-3b6b58:
LGTM

PR #188, opened 2016-06-16
Show server and client versions in client-side UI

maffoo:
The server and client versions are pulled from the state of the git tree when the client and server are built, using the sbt-git plugin on the server side and the git-describe library in gulp for the client. Both of these give a string describing the version that is based on git tags and the state of the tree. If you build a version tagged as `1.0.1`, the version will just be `1.0.1`. Otherwise the version will show the number of commits beyond a given tag, so if we build master several commits after that tagged version, the version string will be, say, `1.0.1-7-g0c80392` to indicate that we are 7 commits past the `1.0.1` tag at revision `0c80392`. In addition, if the local tree has uncommitted changes when things are built, then we'd see that tacked on the end as well: `1.0.1-7-g0c80392-dirty`. We also change the sbt build to use this when producing archives, so we don't need to update version numbers in the code, just use git tags.

Note that I've included two somewhat unrelated changes to add some options for specifying a manager host suffix and optionally redirecting the grapher and registry routes when using a single grapher or registry with multiple managers as we tend to do. I included them because they were developed in that order and so it was easier to put them all up rather than disentangle. I can pull them apart if desired, but it shouldn't be too bad to review the individual commits.

@joshmutus

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,8 +1,11 @@
 lazy val commonSettings = Seq(
-  version := "1.0.1",
+  useJGit,
+  git.useGitDescribe := true,
   scalaVersion := "2.11.7",
   resolvers += "bintray" at "http://jcenter.bintray.com"
 )
+
+enablePlugins(GitVersioning)
 
 lazy val jsonrpc = project.in(file("jsonrpc"))
   .settings(commonSettings)
@@ -36,6 +39,17 @@ lazy val server = project.in(file("server"))
     fork in run := true,
     connectInput in run := true,
     javaOptions += "-Dorg.labrad.stopOnEOF=true",
+
+    // add a resource generator that saves the git version in a file
+    resourceGenerators in Compile += Def.task {
+      val versionStr = version.value
+      val s = streams.value
+      s.log.info(s"git version: $versionStr")
+      val dstDir = (resourceManaged in Compile).value
+      val dstFile = dstDir / "org" / "labrad" / "browser" / "version.txt"
+      IO.write(dstFile, versionStr)
+      Seq(dstFile)
+    }.taskValue,
 
     // add a resource generator that copies compiled files from the client project
     resourceGenerators in Compile += Def.task {
--- a/client-js/app/elements/labrad-app.html
+++ b/client-js/app/elements/labrad-app.html
@@ -60,6 +60,17 @@
       direction: rtl;
     }
 
+    #tech-info {
+      margin: 20px;
+    }
+    #tech-info h1 {
+      font-size: 16px;
+    }
+    #tech-info p {
+      font-size: 12px;
+      font-family: monospace;
+    }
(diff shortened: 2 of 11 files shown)


maffoo:
@joshmutus, I have this exact version running on /dev/maffoo/ now; you can check the revision in the UI there :-)

maffoo:
@jwenner, this version is now running on matrix at the usual host and port but under the /dev/maffoo/ path. If you'd like to check it out that would be great. You'll see the server and client versions in the UI when you open the main menu. I think this will be really helpful when we're trying to track down issues.

jwenner:
The git versions indeed show up.  Does the `g` in the version number need to be there? (I was initially expecting the g to be part of the git SHA-1 until I realized that it stood for git. And yes, I do realize that g=10 in hexadecimal land.)

As for the other changes (besides showing the versions), I can't figure out what's going on, so I can't comment on them.

joshm-3b6b58:
So the SHA tagging is really nice, I have a perhaps related question: when are we bumping versions?

jwenner:
> Add options to redirect grapher and registry routes to default manager. 

I'm not so sure I like this. After all, you want to make sure that the registry you are viewing and editing is the same as the one that the data-taking code uses. If for some reason they get out of sync, then this change would mask this and lead to lots of confusion. Similarly with the grapher and the data_vault.

maffoo:
@joshmutus, we've been rather inconsistent about bumping versions. Basically, we do it whenever we want to release a version for deployment. We may want to get more systematic about this.

@jwenner, the idea is that the person deploying the web interface would only enable the registry and grapher redirect options when you have a setup where the registry and data vault are shared among all managers, as we do. The scenario you propose where the views of these "get out of sync" on different managers should not happen, since they share an underlying backing store. My perspective is that this would help _avoid_ confusion by making it clear that there are not separate registries and data vaults on the different managers. But we can see how other people feel about this and it's easy to turn these options off.

jwenner:
What I was thinking of was back when we had registry caching (do we still have this?), when you had to update the registry on the data-taking computer (or was it through the particular manager?). Otherwise, I vaguely recall that the computer would not recognize the registry update and would use the old value. Note: this was back in the Delphi manager days.

maffoo:
Registry caching no longer exists. In addition, the latest version of the manager allows to share the registry using a multi-headed approach like the data vault where one registry process logs in to multiple managers. This makes notifications work properly across all managers when changes are made anywhere, which you don't get by just pointing all the registries at the same directory on the filesystem. I'd like to fix up our manager configurations to take advantage of this.

jwenner:
As per our discussion (where you said that the common registry/grapher is easy to turn off and on and that the default is off), LGTM based on functionality. Note that I have not looked over the code itself (as I don't know scala/typescript).

PR #189, opened 2016-06-16
Fix d3 package organization in package.json

maffoo:
d3 has apparently changed github organizations. jspm changes this when running a clean `jspm install` so just want to commit the change.


joshm-3b6b58:
LGTM

PR #191, opened 2016-06-22
Reduce the amount of output written to the console log.

adamcw:
Writting to the console harms performance, and excessive logging makes it very
difficult to follow. Removed all calls to console.log or promoted them to a
more appropriate console logging type: warn, info or error. Console log should
be disabled before merging with master. Chrome Development tools have a lot of
tools, such as to inspect websocket traffic that should make console logging in
production (especially inside loops) unnecessary.

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -109,7 +109,6 @@ window.addEventListener('WebComponentsReady', () => {
     // Strip trailing slash, since routes have leading slash already.
     prefix = href.substring(0, href.length - 1);
   }
-  console.log('urlPrefix', prefix);
 
   var body = document.querySelector('body');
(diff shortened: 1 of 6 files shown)


client-js/app/scripts/app.ts
```diff
@@ -363,11 +362,11 @@ window.addEventListener('WebComponentsReady', () => {
         await promises.sleep(5000);
         var services: Services = null;
         try {
-          console.log('attempting to connect');
+          console.log('Attempting to connect');
```
maffoo: `console.info`?

maffoo:
One tiny comment, otherwise LGTM.

adamcw:
Was on the edge about removing that one entirely and missed upgrading it to info, good catch.

PR #192, opened 2016-06-22
Registry UI fixes

maffoo:
- Limit registry keys to a single line in the main list view with ellipsis when a key is truncated.
- Allow double clicking anywhere on a registry key to edit, including the icon and text.
- Improve polymer dialog style when resizing page so edit box does not go off screen.

@DanielSank, @adamcw


adamcw:
Restructured the event listeners to better delegate the responsibility and avoid duplication. The CSS and remaining changes LGTM.

PR #193, opened 2016-06-22
Modified gulp to allow a --host flag to determine websocket host.

adamcw:
Previously the websocket gulp inserted into the web client was hardcoded to be
localhost:7667. This change allows any host to be provided via the --host
command line flag, eg `gulp serve --host adamcw.example.com:7667`.


maffoo:
Maybe call the option `--api-host` instead. Otherwise LGTM.

PR #194, opened 2016-06-22
Updated Polymer to v1.5.X in bower.json

adamcw:
Due to package element dependencies requiring Polymer 1.5.0, we need to resolve
the conflict between required versions. It is preferred to update to the latest
version of Polymer, rather than freeze it at an older version.


maffoo:
LGTM

DanielSank:
I just ran into this too (see #195). LGTM.

Issue #195, opened 2016-06-22
Avoid globally installing any dependencies


DanielSank:
We'd like to be able to install packages like typescript locally.

npm users often globally install things like gulp and typescript because it puts binaries associated with those packages on the system PATH so that e.g. `$ gulp` works from the shell. This is convenient but it's not great that we have to make our entire system point at one particular version of gulp; what if different projects want different versions?

If you _locally_ install gulp in a particular npm package at `.../myproject` then in fact you get a symlink at `.../myproject/node_modules/.bin/gulp` which points to `.../myproject/node_modules/gulp/bin/gulp.js`. So you see we actually do have a binary we can use. There are two ways to get at this binary.
1. Put in on your PATH, e.g. via `bashrc`. We're already doing this to get at our `node` binary so this isn't the worst idea ever. However, we would like the repo to be self-sufficient, so if we were to do this I'd provide a `setup-my-shell.sh` script which does the appropriate PATH modifications.
2. Use `npm run`. This is a special program which looks for scripts defined in `package.json`'s `scripts` field. So for example, if `package.json` contains
   
   ```
   "scripts": {"gulp": ./node_modules/.bin/gulp}
   ```
   
   then `npm run gulp` does what we want. Even better though, `npm run` knows about the `.bin` directory, so you can just do this:
   
   ```
     "scripts": {"gulp": gulp}
   ```
   
   and it works the same. If we go this route, I think the only change we have to make (aside from the  documentation) is that we have to put e.g. `npm run typescript` in the gulp file.

Here's a somewhat more verbose version of what's written here:
https://firstdoit.com/no-need-for-globals-using-npm-dependencies-in-npm-scripts-3dfb478908#.wdwywwrfo

maffoo:
I vote option 2, for sure.

I think we can avoid shelling out to typescript from inside the gulpfile by using the appropriate gulp plugin (gulp-typescript?) or maybe even shelling out to `npm run typescript` instead. The plugin is probably less hacky, but it means yet another dependency and as I recall when I tried it in the past the gulp-typescript plugin depended on a specific typescript version which was old, and in general you can't upgrade typescript until the plugin updates. Not a big deal now because the typescript version we need has been around for a while so the plugin probably works with it, but this is something to consider.

DanielSank:
[...]

PR #196, opened 2016-06-22
Eliminate global npm installs.

DanielSank:
See #195. This is an implementation of the second proposed method therein.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,6 @@ the node package manager `npm`. Then, from inside the `client-js` directory you 
 global and project-level dependencies:
 
 ```
-$ npm install gulp typescript jspm bower
 $ npm install
 $ npm run jspm install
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -31,10 +31,10 @@ the node package manager `npm`. Then, from inside the `client-js` directory you
 global and project-level dependencies:
 
 ```
-$ npm install -g gulp typescript jspm bower
+$ npm install gulp typescript jspm bower
```
maffoo: Should not have to do this anymore. All packages we need should be included in `package.json` and will get installed when you do `npm install`.

maffoo: Looks like we need to add `bower` to `devDependencies` in `package.json`, but gulp and typescript are already there.

maffoo:
LGTM

PR #198, opened 2016-06-24
Fixed cursor positioning on grapher.

adamcw:
Cursor positioning was messed up in two ways. Firstly, it wasn't correctly
accounting for the offset of the graph from the side and top of the page.
Secondly, the use of `offsetX` was relative to the item that you moused over
rather than the graph, so mousing directly over a line would cause the
positioning to jump to non-sense values as the offsets were incorrect for the
intended purpose. Fixed by using currentTarget and getBoundingClientRect to
manually calculate the offset relative to the plot.


maffoo:
LGTM

PR #201, opened 2016-06-27
Created an IronIconBakery and ListBuilder class to allow for more performant UI.

adamcw:
Applied these classes to the Grapher directory/dataset listing.

Polymer does a lot of fancy data-binding and event listening, which makes things
very slow at scale. Rendering a list of 4000 datasets would freeze the entire
application for 6-7 seconds, sometimes more. Even traversing simple lists took
a noticeable amount of time which just leads to a frustrating user experience,
especially if navigating multiple levels of a directory structure.

To fix this, I've moved the rendering of the grapher list view into raw
JavaScript rather than using the Polymer templating language. As a result,
small lists now render instantly, and large lists, while still take a few
seconds to entirely load (the DOM just doesn't like this many elements being
added) the page and the top of the list renders instantly giving the user the
impression of speed.

This implementation could be further optimized, but should already be overkill for the average use case, with filtering really the solution to lists greater than a few hundred items rather than trying to speed up huge DOM insertions.

This PR is a first step towards overhauling and unifying the list views across
the application to allow for easier fixes and improvements to the UI.

### DIFF
--- a/client-js/app/scripts/listbuilder.ts
+++ b/client-js/app/scripts/listbuilder.ts
@@ -1,0 +1,106 @@
+/**
+ * Allows for the generation of a large list of DOM nodes by chunking the
+ * rendering.
+ */
+export class ListBuilder {
+  /**
+   * Whether or not there is current a list rendering.
+   */
+  public isRendering: boolean = false;
+
+
+  /**
+   * The current renderId of a new render.
+   */
+  private renderId_: number = 0;
+
+
+  /**
+   * All chunks in the queue with a renderId less than or equal to this number
+   * will not fire, and will consider themselves interrupted.
+   */
+  private renderInterrupt_: number = -1;
+
+
+  /**
+   * The current chunks that are yet to be rendered.
+   */
+  private chunkQueue_: Array = [];
+
+
+  /**
+   * Renders a list of items, appending them to an element in chunks.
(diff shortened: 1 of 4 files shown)


joshm-3b6b58:
List of plots in grapher renders funny:

image

joshm-3b6b58:
Would it be simpler to just get a copy of the SVGs from [here](https://design.google.com/icons/) and dispense with the iron-icons altogether?

client-js/app/scripts/listbuilder.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * Allows for the generation of a large list of DOM nodes by chunking the
+ * rendering.
+ */
+export class ListBuilder {
```
joshm-3b6b58, line 5: Does [`<iron-list>`](https://elements.polymer-project.org/elements/iron-list) have this functionality built in already?

adamcw:
I have decided to abandon this pull request and use <iron-list>. I originally didn't use it for a few reasons: 1) It said it only supported fixed height lists 2) It didn't want to work 3) It didn't support the selectable table 4) I had no idea how filtering would work on such an item. Your comment lead me to review my decision and I do believe iron-list should be suitable, at least in the near term.

PR #202, opened 2016-06-27
Update the Registry menu UI styling

adamcw:
A few simple changes to the Register menu UI
- Items are now always the full width of the screen.
- The icons are now clickable.
- The back arrow is now always a black color instead of being link colored (you've generally visited the parent to get to the child, so the back link being 'visited' is meaningless).


maffoo:
LGTM

PR #203, opened 2016-06-27
Update d3 to v3.5.17

adamcw:
Updated to the latest d3 since it's preferred to be on the latest version if it doesn't break anything. v4.0.0 is going to bring a lot of changes, so we should probably try and keep up to date to avoid accidentally using old features where possible.

### DIFF
--- a/client-js/config.js
+++ b/client-js/config.js
@@ -10,7 +10,7 @@ System.config({
 
 System.config({
   "map": {
-    "d3": "github:mbostock/d3@3.5.6",
+    "d3": "github:mbostock/d3@3.5.17",
     "page": "npm:page@1.6.3",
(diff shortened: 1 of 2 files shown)


maffoo:
@adamcw, could you comment briefly on why this is needed? I have no particular objection, just curious if there's some specific bugfix or performance fix that we need.

maffoo:
LGTM. One nit for when you commit, we try to follow the git [recommendations](https://git-scm.com/book/ch5-2.html) to use the active present tense in commit messages, so "Update d3 to c3.5.17" instead of "Updated...". It's a minor point, but helps when reading the commit history if the tenses are consistent.

adamcw:
I updated this to try and fix a problem but it didn't help, mostly figured it was worth updating because it didn't break anything and is likely faster with better features.

PR #204, opened 2016-06-27
Add THREE.js to be used for improved plotting performance.

adamcw:
THREE.js will be made to transition plotting away from SVG into WebGL/Canvas in order to get better graphing performance.

### DIFF
--- a/client-js/config.js
+++ b/client-js/config.js
@@ -13,7 +13,7 @@ System.config({
     "d3": "github:mbostock/d3@3.5.17",
     "three": "github:mrdoob/three.js@master",
     "page": "npm:page@1.6.3",
-    "three": "github:mrdoob/three.js@master",
+    "three": "github:mrdoob/three.js@r78",
     "traceur": "github:jmcriffey/bower-traceur@0.0.88",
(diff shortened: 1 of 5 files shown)


client-js/config.js
```diff
@@ -12,6 +12,7 @@ System.config({
   "map": {
     "d3": "github:mbostock/d3@3.5.6",
     "page": "npm:page@1.6.3",
+    "three": "github:mrdoob/three.js@master",
```
maffoo: Can we use a tagged version instead of master? This could pull down a slightly different version for different people depending on when they run jspm install.

adamcw: Ahh, was meaning to do this. Done.

client-js/config.js
```diff
@@ -11,7 +11,9 @@ System.config({
 System.config({
   "map": {
     "d3": "github:mbostock/d3@3.5.17",
+    "three": "github:mrdoob/three.js@master",
```
maffoo: This looks like a dupe; can remove.

maffoo:
Remove redundant entry in `config.js`. Otherwise LGTM.

PR #205, opened 2016-06-27
Cleanup labrad-plot.

adamcw:
In the process of updating the plotter to use WebGL, I have reorganized labrad-plot.ts to be more in line with Google Style standards.
- Made the maximum line width 80 characters.
- Updated comments to start with capitals and end with full stops.
- Converted all uses of `var` to `let` or `const`.
- Made all private methods end in a suffix.
- Ensured all public methods have a comment.

I also reorganized the file to make it easier to follow:
- Grouped observers and listeners and renamed them appropriately.
- Made functions that the UI interacts with public.
- Renamed UI functions to make their use clearer.
- Moved all utility functions to the end of the file.

I have tried to keep each commit as atomic to one action as I could so that review could be made easier by looking at each commit individually. If there is a better process for submitting changes like this, please let me know (chaining a bunch of dependent PRs seemed like a rebasing nightmare).

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -20,15 +20,15 @@ const COLOR_LIST = [
 /**
  * Colors for 2D plots, using viridis colormap from matplotlib.
  */
-var COLOR_MAP = viridisData.map((rgb) => {
-  var [r, g, b] = rgb;
+const COLOR_MAP = viridisData.map((rgb) => {
+  const [r, g, b] = rgb;
   return d3.rgb(255 * r, 255 * g, 255 * b);
 });
 
 
 @component('labrad-plot')
 export class Plot extends polymer.Base {
-.
+
   @property({type: String, value: ''})
   xLabel: string;
 
@@ -105,7 +105,7 @@ export class Plot extends polymer.Base {
    */
   public addData(data: Array<Array<number>>) {
     if (data.length === 0) return;
-    var lastData = this.data.length > 0 ? this.data[this.data.length - 1] : null;
+    const lastData = this.data.length > 0 ? this.data[this.data.length - 1] : null;
     for (let row of data) {
       this.splice('data', this.data.length, 0, row);
     }
@@ -114,10 +114,10 @@ export class Plot extends polymer.Base {
 
 
   private createPlot_(area: HTMLElement, totWidth: number, totHeight: number): void {
-    var p = this;
+    const p = this;
 
-    var width = totWidth - p.margin.left - p.margin.right;
-    var height = totHeight - p.margin.top - p.margin.bottom;
+    const width = totWidth - p.margin.left - p.margin.right;
+    const height = totHeight - p.margin.top - p.margin.bottom;
 
     p.width = width;
     p.height = height;
@@ -261,7 +261,7 @@ export class Plot extends polymer.Base {
 
     for (let i of this.displayTraces) {
       // Extract data for trace i, starting with the last datapoint to avoid gaps.
-      var traceData = [];
+      const traceData = [];
       if (lastData) {
         traceData.push([lastData[0], lastData[i+1]]);
       }
@@ -341,7 +341,7 @@ export class Plot extends polymer.Base {
     this.limits.yMin = isNaN(this.dataLimits.yMin) ? 0 : this.dataLimits.yMin;
     this.limits.yMax = isNaN(this.dataLimits.yMax + this.dy) ? 0 : this.dataLimits.yMax + this.dy;
 
-    var zMin = this.dataLimits.zMin,
+    let zMin = this.dataLimits.zMin,
         zMax = this.dataLimits.zMax;
     if (zMin == zMax) {
       zMin -= 1;
@@ -349,10 +349,12 @@ export class Plot extends polymer.Base {
     }
 
     // Add a dot to the plot.
-    var p = this;
+    let w = 0,
+        h = 0;
+    const p = this;
     switch (this.drawMode2D) {
     case 'dots':
-      var w = 4, h = 4;
+      w = h = 4;
       for (let row of data) {
         this.chartBody
               .append('rect')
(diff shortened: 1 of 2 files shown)


adamcw:
I branched this off of https://github.com/labrad/scalabrad-web/pull/201, I'm not sure how to deal with these extra commits on this PR given your rebase workflow.

joshm-3b6b58:
Thanks for breaking this branch up logically and thanks for cleaning it up. Going thru the PR now.

Firstly the grapher list doesn't render correctly for me. I get a newline after the icon:
image

joshm-3b6b58:
The above comment is a problem in Pull #201 so, I'll start reviewing there?

adamcw:
Is there a way to review this despite the other branch being broken? They aren't even really dependent, I just rebased in the changes because it made testing quicker with the more performant list.

joshm-3b6b58:
Most of it is just cosmetic so shouldn't be too hard to review on it's own. I like to check that the app functions before giving the go ahead though...

adamcw:
I have rebased and dropped the commits from the other (now abandoned) branch.

client-js/app/elements/labrad-plot.ts
```diff
@@ -144,164 +105,38 @@ export class Plot extends polymer.Base {
...
-        this.$$('rect.background').style.fill = '#222222';
-        this.is1D = false;
-        this.is2D = true;
-        break;
+  /**
+   * Add new data to the plot and re-zoom.
+   * Fires when new data arrives via the socket.
+   */
+  public addData(data: Array<Array<number>>) {
+    if (data.length === 0) return;
```
joshm-3b6b58, line 124: How come we're checking if `data.length` has zero length? Does it ever fire if data is empty?

adamcw, line 124: I think I've dropped this line in my branch. I aimed for this PR to be purely cosmetic where possible. Though now you mention it, === should be == for style reasons.

client-js/app/elements/labrad-plot.ts
```diff
@@ -811,3 +789,66 @@ function prettyNumber(value: number, min: number, max: number, res: number): str
...
 
+
+/**
+ * Get a color for the given value on a scale with the given min and max
+ * limits (zMin <= z <= zMax).
+ */
+function getColor(z: number, zMin: number, zMax: number) {
+  var index = 128;
+  if (zMin !== zMax) {
+    index = Math.floor(255 * (z - zMin) / (zMax - zMin));
```
maffoo: I think you still want to use 256 here instead of 255, or else the range of the highest color value is contracted to occur _only_ when `z == zMax`. Then you can either restore the duplication of the top color value in `COLOR_MAP[256]` or else just add a check here to see if `z == zMax` and set index to 255 in that case.

client-js/app/elements/labrad-plot.ts
```diff
@@ -783,7 +823,8 @@ export class Plot extends polymer.Base {
           x = this.mouseToDataX_(event.pageX - rect.left),
           y = this.mouseToDataY_(event.pageY - rect.top);
 
-    this.currPos = `(${prettyNumber(x, xMin, xMax, dx)}, ${prettyNumber(y, yMin, yMax, dy)})`;
+    this.currPos = `(${prettyNumber(x, xMin, xMax, dx)}, ` +
+                   `${prettyNumber(y, yMin, yMax, dy)})`;
```
maffoo: I'd suggest making these separate expressions to keep the string interpolation simple:

``` typescript
const xStr = prettyNumber(x, xMin, xMax, dx),
      yStr = prettyNumber(y, yMin, yMax, dy);
this.currPos = `(${xStr}, ${yStr})`;
```

client-js/app/elements/labrad-plot.ts
```diff
@@ -813,7 +854,8 @@ function expDigits(value: number, res: number): number {
  * format, and then chooses the smaller representation. We also limit fixed
  * point format to numbers less than a million in absolute value.
  */
-function prettyNumber(value: number, min: number, max: number, res: number): string {
+function prettyNumber
+    (value: number, min: number, max: number, res: number): string {
```
maffoo: Leave opening paren on previous line with function name.

maffoo:
A few small comments; otherwise LGTM. Thanks for keeping the individual commits small and focused; that made the review much easier.

client-js/app/elements/labrad-plot.ts
adamcw: Good catch. Done.

client-js/app/elements/labrad-plot.ts
adamcw: I don't see why this behavior is desired. The highest color should only occur when the color is equal to `zMax`. We don't complain that `COLOR_MAP[0]` is only used when `z == zMin`.

adamcw: I see what you mean. I don't like the appending the same color twice trick at all though. I'd be more inclined to round half-up or down rather than using `Math.floor` to fix this problem.

client-js/app/elements/labrad-plot.ts
adamcw: A nice idea. Done.

client-js/app/elements/labrad-plot.ts
maffoo: I think a simple alternative to appending the last color a second time to the list is to add an explicit check for `z == zMax`:

``` typescript
function getColor(z: number, zMin: number, zMax: number) {
  let index = 128;
  if (zMin !== zMax) {
    if (z == zMax) {
      index = 255;
    } else {
      index = Math.floor(256 * (z - zMin) / (zMax - zMin));
    }
  }
  return COLOR_MAP[index];
}
```

adamcw: How about this:

``` typescript
function getColor(z: number, zMin: number, zMax: number) {
  if (zMin === zMax) {
    return COLOR_MAP[128];
  }
  if (z === zMax) {
    return COLOR_MAP[255];
  }
  const index = Math.floor(256 * (z - zMin) / (zMax - zMin));
  return COLOR_MAP[index];
}
```

maffoo: yes, much better!

maffoo:
LGTM

PR #206, opened 2016-06-27
Use iron-list for displaying menu in grapher.

adamcw:
Replaced the `table` in labrad-grapher with an `iron-list` making it substantially more performant even on large datasets.

### DIFF
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -8,7 +8,6 @@
 <link rel="import" href="../bower_components/iron-icons/communication-icons.html">
 <link rel="import" href="../bower_components/iron-icons/iron-icons.html">
 <link rel="import" href="../bower_components/iron-input/iron-input.html">
-<link rel="import" href="../bower_components/iron-list/iron-list.html">
 <link rel="import" href="../bower_components/iron-pages/iron-pages.html">
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
 <link rel="import" href="../bower_components/paper-button/paper-button.html">
--- a/client-js/app/elements/elements.html
+++ b/client-js/app/elements/elements.html
@@ -8,6 +8,7 @@
 <link rel="import" href="../bower_components/iron-icons/communication-icons.html">
 <link rel="import" href="../bower_components/iron-icons/iron-icons.html">
 <link rel="import" href="../bower_components/iron-input/iron-input.html">
+<link rel="import" href="../bower_components/iron-list/iron-list.html">
 <link rel="import" href="../bower_components/iron-pages/iron-pages.html">
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
 <link rel="import" href="../bower_components/paper-button/paper-button.html">
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -58,7 +58,10 @@ export class LabradGrapher extends polymer.Base {
 
   @listen("keypress")
   onKeyPress(event) {
-    var id = this.selectedId;
+    if (!this.selectedId) {
+      return "";
+    }
+    var id = this.selectedId.id;
     if (!id) return;
     switch (event.charCode) {
       case 115 /* s */:
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -38,9 +38,6 @@ export class LabradGrapher extends polymer.Base {
   @property({type: Object})
   selectedDatasetInfo: datavault.DatasetInfo = null;
 
-  @property({type: Number, value: 0})
-  kick: number;
-
   @property({type: Object})
   places: Places;
 
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -238,7 +238,7 @@ export class LabradGrapher extends polymer.Base {
     }
 
     if (selected.id.startsWith('dataset:')) {
-      var name = selected.id.substring(8);
+      const name = selected.id.substring(8);
       this.fetchInfo(name);
       return name;
     } else {
@@ -247,7 +247,7 @@ export class LabradGrapher extends polymer.Base {
   }
 
   async fetchInfo(name: string) {
-    var info = await this.api.datasetInfo({
+    const info = await this.api.datasetInfo({
       path: this.path,
       dataset: datavault.datasetNumber(name),
       includeParams: true
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -12,6 +12,8 @@ type ListItem = {
   trashed: boolean
 };
 
+type ListItemFilterFunction = (item: ListItem) => boolean;
+
 @component('labrad-grapher')
 export class LabradGrapher extends polymer.Base {
   @property({type: Array, notify: true})
@@ -199,7 +201,7 @@ export class LabradGrapher extends polymer.Base {
     }
   }
 
-  private filterListItemsFunction() {
+  private filterListItemsFunction(): ListItemFilterFunction {
     // Return trashed and parent items
     if (this.showTrash) {
       return (x) => (x.trashed || x.isParent);
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -215,7 +215,9 @@ export class LabradGrapher extends polymer.Base {
   }
 
   private applyFilterToListItems(): void {
-    // Will cause the `iron-list` view to scroll (jump) to the top.
+    // Will cause the `iron-list` view to scroll (jump) to the top when
+    // rendering even if no changes to the filter are applied. Only use when
+    // this behavior is desired, such as swapping filtering modes.
     const filterFunction = this.filterListItemsFunction();
     this.set('filteredListItems', this.listItems.filter(filterFunction));
   }
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -133,6 +133,7 @@ export class LabradGrapher extends polymer.Base {
         this.set('listItems.' + selectedIndex + '.trashed',
                  !this.selected.trashed);
 
+        // Whenever we toggle trash, it will be removed from the view.
         if (this.filterListItems_([this.selected]).length == 0) {
           this.splice('listItems', selectedIndex, 1);
         }
@@ -220,6 +221,7 @@ export class LabradGrapher extends polymer.Base {
   }
 
   private applyListFilter_(): void {
+    // Will cause `iron-list` to jump to the top.
     this.set('listItems', this.filterListItems_(this.listItems_));
   }
 
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -13,7 +13,7 @@ export class LabradGrapher extends polymer.Base {
   datasets: Array<{name: string; url: string; tags: Array<string>}>;
 
   @property({type: String, notify: true})
-  selectedId: string;
+  selected: {id: string};
 
   @property({type: Object})
   selectedDatasetInfo: datavault.DatasetInfo = null;
@@ -58,10 +58,10 @@ export class LabradGrapher extends polymer.Base {
 
   @listen("keypress")
   onKeyPress(event) {
-    if (!this.selectedId) {
+    if (!this.selected) {
       return "";
     }
-    var id = this.selectedId.id;
+    var id = this.selected.id;
     if (!id) return;
     switch (event.charCode) {
       case 115 /* s */:
@@ -169,13 +169,13 @@ export class LabradGrapher extends polymer.Base {
   }
 
   @computed()
-  selectedDataset(selectedId: string): string {
-    if (!selectedId) {
+  selectedDataset(selected: {id: string}): string {
+    if (!selected) {
       return "";
     }
 
-    if (selectedId.id.startsWith('dataset:')) {
-      var name = selectedId.id.substring(8);
+    if (selected.id.startsWith('dataset:')) {
+      var name = selected.id.substring(8);
       this.fetchInfo(name);
       return name;
     } else {
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -1,19 +1,39 @@
 import * as datavault from '../scripts/datavault';
 import {Places} from '../scripts/places';
 
+type ListItem = {
+  id: string,
+  name: string,
+  url: string,
+  isDir: boolean,
+  isDataset: boolean,
+  isParent: boolean,
+  starred: boolean,
+  trashed: boolean
+};
+
 @component('labrad-grapher')
 export class LabradGrapher extends polymer.Base {
   @property({type: Array, notify: true})
   path: Array<string>;
 
-  @property({type: Array})
+  @property({type: Array, value: () => { return []; }})
   dirs: Array<{name: string; url: string; tags: Array<string>}>;
 
-  @property({type: Array})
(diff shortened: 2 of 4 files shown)


client-js/app/elements/elements.html
```diff
@@ -8,6 +8,7 @@
 <link rel="import" href="../bower_components/iron-icons/communication-icons.html">
 <link rel="import" href="../bower_components/iron-icons/iron-icons.html">
 <link rel="import" href="../bower_components/iron-input/iron-input.html">
+<link rel="import" href="../bower_components/iron-list/iron-list.html">
```
joshm-3b6b58: So this html file is a vestige that we have to get rid of. We're putting imported elements in the file that imports them. Sorry I should make an issue of this I keep forgetting

joshm-3b6b58:
So there's a subtle bug here, that I'm not sure we care about. I'm creating a folder with a large number of entries to test performance, and I notice when a new entry is added the page pops back to the beginning of the list. Perhaps this is a limitation with `<iron-list>`? Maybe there's a different away to append to the list?

Edit: In looking a little deeper in this, this is probably because of the way `listItems` is generated, which I think recreates the list from the arrays every time a new entry is added. Fixing this is outside this PR, I believe.

Edit edit: actually this bug doesn't exist on current master. It stays at the current spot on the list when new entries are added.

adamcw:
This probably does require updating the data binding, I'll look into it and then assess if it seems suitable to be part of this PR.

client-js/app/elements/elements.html
adamcw: Done.

joshm-3b6b58:
Yeah, I've been looking into this a little bit. It looks like scroll to the top on update is the desired behavior for some reason? I can't tell.

joshm-3b6b58:
There are methods where this could be done manually where you can track the `lastVisibleIndex` and update the `scrollToIndex()`

adamcw:
This is now more of a monster of a PR to review than I would like, but I have updated the code to fix the jumping-to-the-top issue.

I completely reworked the data-binding to be incremental, allowing the list to be appended to rather than rewriting the entire thing which was causing iron-list to jump to the top.

I was able to remove the dependency on the kick++ hack which is also nice.

There is probably more refactoring that could go here, such as removing the dirs/datasets variables entirely, but this required more refactoring of `activities.ts` than I was inclined to do in one PR.

client-js/app/elements/labrad-grapher.ts
```diff
@@ -35,30 +52,46 @@ export class LabradGrapher extends polymer.Base {
...
+      this.datasets);
+    this.applyListFilter_();
   }
 
   starClicked() {
     this.showStars = !this.showStars;
     if (this.showStars) this.showTrash = false;
     this.updateButtons();
-    this.kick++;
+    this.applyListFilter_();
```
joshm-3b6b58: So I'm not familiar with this convention of ending this with underscores, what's the deal with that?

client-js/app/elements/labrad-grapher.ts
```diff
@@ -35,30 +52,46 @@ export class LabradGrapher extends polymer.Base {
...
+
   @listen("keypress")
   onKeyPress(event) {
-    var id = this.selectedId;
+    if (!this.selected) {
+      return "";
+    }
+
+    const selectedIndex = this.listItems.indexOf(this.selected);
+
```
joshm-3b6b58: Why the whitespace here?

client-js/app/elements/labrad-grapher.ts
```diff
@@ -35,30 +52,46 @@ export class LabradGrapher extends polymer.Base {
...
 
   starClicked() {
     this.showStars = !this.showStars;
     if (this.showStars) this.showTrash = false;
     this.updateButtons();
-    this.kick++;
+    this.applyListFilter_();
   }
 
   trashClicked() {
```
joshm-3b6b58: I know you didn't name this originally, but `trashClicked()` and `starClicked()` could use more descriptive function names. For a quick read through I can't tell if this is the function that filters the list or tags the datasets.

client-js/app/scripts/activities.ts
```diff
@@ -74,8 +74,8 @@ export class DatavaultActivity implements Activity {
 
   async start(): Promise<ActivityState> {
     console.info('Loading datavault:', this.path);
-    this.api.newDir.add(x => this.onNewDir(), this.lifetime);
-    this.api.newDataset.add(x => this.onNewDataset(), this.lifetime);
+    this.api.newDir.add(x => this.onNewDir(x), this.lifetime);
+    this.api.newDataset.add(x => this.onNewDataset(x), this.lifetime);
```
joshm-3b6b58, line 78: Why was this change necessary?

Edit: so before `onNewDir()` waited for a new dir to pop up but now `onNewDir` is called with the argument of the new dir when a new dir pops up?

adamcw, line 78: If you look at 115 and 124 you can see that the new dir and new dataset functions now actually receive the name of the new dataset. This lets them pass just the latest item to the grapher rather than regenerating the whole list by calling the api like it used to.

client-js/app/elements/labrad-grapher.ts
adamcw: I'll remove it.

client-js/app/elements/labrad-grapher.ts
adamcw: There is quite a bit more in this file I wanted to rename/refactor, but I was trying to keep this as small and single-function of a PR as possible.

joshm-3b6b58: OK good.

client-js/app/elements/labrad-grapher.ts
adamcw: Google style guide for JavaScript (and many languages) has all private members and private methods suffixed with an underscore. I can't seem to find exactly where it is called out, but it is common across the style guides.

https://google.github.io/styleguide/javascriptguide.xml?showone=Visibility__private_and_protected_fields_#Visibility__private_and_protected_fields_

https://google.github.io/styleguide/cppguide.html

I have tried to make new code at least follow the style guide, which leads to a weird mash of styles, but it's somewhat unavoidable without every PR first requiring a PR where the entire file gets refactored which are burdensome to review. If this is the workflow we'd prefer moving forward, we can do that.

joshm-3b6b58:
Very nice. The list renders much faster than before and behaves as I expect when new listings are added.
For 1150 entries this is the performance I see
New: 
image
Old: 
image

LGTM

adamcw:
Does that performance graph include you scrolling to the bottom of the list?

joshm-3b6b58:
This is just initial page load. Scrolling seems fine.

client-js/app/elements/labrad-grapher.ts
maffoo: It's actually recommended _not_ to do this for typescript since we have real visibility modifiers like `private`.

client-js/app/elements/labrad-grapher.ts
```diff
@@ -35,30 +52,46 @@ export class LabradGrapher extends polymer.Base {
...
     this.showTrash = !this.showTrash;
     if (this.showTrash) this.showStars = false;
     this.updateButtons();
-    this.kick++;
+    this.applyListFilter_();
   }
 
-  updateButtons() {
+  updateButtons(): void {
```
maffoo, line 79: Is the idea here just to have explicit return type annotations? If we're going to require that on `void` functions, we should be consistent and try to do this everywhere (for example, also on the `onKeyPress` method below). I like the idea of explicit return types on public methods, so I'd be for this.

client-js/app/elements/labrad-grapher.ts
```diff
@@ -35,30 +52,46 @@ export class LabradGrapher extends polymer.Base {
...
   @listen("keypress")
   onKeyPress(event) {
-    var id = this.selectedId;
+    if (!this.selected) {
+      return "";
+    }
+
+    const selectedIndex = this.listItems.indexOf(this.selected);
+
+    var id = this.selected.id;
```
maffoo: looks like this can also be `const`

client-js/app/elements/labrad-grapher.ts
```diff
@@ -77,6 +110,8 @@ export class LabradGrapher extends polymer.Base {
             tags: ['^star']
           });
         }
+        this.set('listItems.' + selectedIndex + '.starred',
```
maffoo: typescript has string interpolation:

``` typescript
this.set(`listItems.${selectedIndex}.starred`,
         !this.selected.starred);
```

client-js/app/elements/labrad-grapher.ts
```diff
@@ -134,41 +185,61 @@ export class LabradGrapher extends polymer.Base {
...
+    if (this.filterListItems_([item]).length == 1) {
+      this.push('listItems', item);
     }
-    return items;
+  }
+
+  private filterListItems_(items: Array<ListItem>) {
+    if (this.showTrash) {
+      return items.filter((x) => {
+        return (x.trashed || x.isParent);
```
maffoo: don't need `return` in arrow methods whose body is a single expression:

``` typescript
return items.filter((x) => (x.trashed || x.isParent));
```

client-js/app/elements/labrad-grapher.ts
```diff
@@ -134,41 +185,61 @@ export class LabradGrapher extends polymer.Base {
...
+
+  private addItemToList_(item: ListItem): void {
+    this.push('listItems_', item);
+    if (this.filterListItems_([item]).length == 1) {
+      this.push('listItems', item);
     }
-    return items;
+  }
+
+  private filterListItems_(items: Array<ListItem>) {
```
maffoo: Might be better to have a private method that returns the filter function, but does not accept an items array and call filter on it. You would then call `items.filter(this.filterFunction)` instead of `this.filterListItems(items)`, and you could also call `this.filterFunction(item)` instead of `this.filterListItems([item]).length !== 0`.

client-js/app/elements/labrad-grapher.ts
```diff
@@ -134,41 +185,61 @@ export class LabradGrapher extends polymer.Base {
...
   @computed()
-  selectedDataset(selectedId: string): string {
-    if (selectedId.startsWith('dataset:')) {
-      var name = selectedId.substring(8);
+  selectedDataset(selected: {id: string}): string {
+    if (!selected) {
+      return "";
+    }
+
+    if (selected.id.startsWith('dataset:')) {
```
maffoo, line 240: I'd suggest pulling out prefixes like `'dataset:'` and `'dir:'` into module-level constants.

client-js/app/elements/labrad-grapher.ts
```diff
@@ -134,41 +185,61 @@ export class LabradGrapher extends polymer.Base {
...
+      });
+    }
+
+    return items.filter((x) => {
+      return !x.trashed || x.isParent;
+    });
+  }
+
+  private applyListFilter_(): void {
+    // Will cause `iron-list` to jump to the top.
```
maffoo: You mean that the view scrolls to the top of the filtered list? Maybe clarify this comment a bit.

client-js/app/elements/labrad-grapher.ts
```diff
@@ -1,26 +1,43 @@
...
+  @property({type: Array, value: () => { return []; }})
   datasets: Array<{name: string; url: string; tags: Array<string>}>;
 
-  @property({type: String, notify: true})
-  selectedId: string;
+  @property({type: Array, value: () => { return []; }})
+  listItems: Array<ListItem>;
+
+  @property({type: Array, value: () => { return []; }})
+  listItems_: Array<ListItem>;
```
maffoo: I would suggest changing the names of these properties to something like `filteredListItems` and `allListItems` (or maybe just `listItems` for the latter, since it is clear that it is not filtered).

client-js/app/elements/labrad-grapher.ts
```diff
@@ -95,35 +130,51 @@ export class LabradGrapher extends polymer.Base {
...
+      id: 'dir:' + dir.name,
+      name: dir.name,
+      url: dir.url,
+      isDir: true,
+      isDataset: false,
+      isParent: false,
+      starred: this.isTagged_(dir, 'star'),
+      trashed: this.isTagged_(dir, 'trash')
+    };
+    this.addItemToList_(obj);
```
maffoo: I would probably drop the declaration and just pass the object literal to the function, but it's not a big deal either way. I'm curious if there's a reason you prefer having the declaration as a separate statement?

``` typescript
this.addItemToList({
  ...
});
```

maffoo:
Arg, sorry, hit wrong button. Just wanted to say please see comments :-)

client-js/app/elements/labrad-grapher.ts
adamcw: I like this idea. Done.

client-js/app/elements/labrad-grapher.ts
adamcw: Agreed, this is very confusing. I was thinking `listItems` and `listItemsView`, but `filteredListItems` sounds a better choice. Done.

client-js/app/elements/labrad-grapher.ts
adamcw: Originally obj was being passed to more than one function, a relic I forgot to reconsider. I agree it would be nicer to just pass the literal directly. Done.

client-js/app/elements/labrad-grapher.ts
adamcw, line 240: I wasn't sure if this sort of refactoring was appropriate or should be done in a separate PR.

maffoo, line 240: Separate PR would be fine; best to keep these small and focused. The state of this code is such that for almost every line that gets touched in a PR, one can see a few other things that need fixing... Glad you're helping fix things up :-)

client-js/app/elements/labrad-grapher.ts
adamcw: Done.

client-js/app/elements/labrad-grapher.ts
adamcw: Done.

client-js/app/elements/labrad-grapher.ts
adamcw: Done.

client-js/app/elements/labrad-grapher.ts
adamcw, line 79: I think we should aim to have types on everything, private or public, it lets the compiler do a lot of work for us for little effort. Done.

client-js/app/elements/labrad-grapher.ts
adamcw: Done.

adamcw:
A little hard to tell if I've responded to all the comments, but I believe it should be ready for another look.

client-js/app/scripts/activities.ts
```diff
@@ -112,14 +112,22 @@ export class DatavaultActivity implements Activity {
...
-  async onNewDir() {
-    var listing = await this.api.dir(this.path);
-    this.elem.dirs = this.getDirs(listing);
+  private onNewDir(item: {name: string, tags?: Array<string>}) {
+    const dir = {
+      name: item.name,
+      url: this.places.grapherUrl(this.path, item.name),
+      tags: (item.tags) ? item.tags : []
+    };
+    this.elem.newDir(dir);
```
maffoo: Could also omit `const` declarations here and in `onNewDataset` and just pass object literals to the appropriate functions.

maffoo:
One minor comment, otherwise LGTM.

client-js/app/scripts/activities.ts
adamcw: Done.

adamcw:
I feel the need to comment while we're on the topic that we may want to avoid `iron-icon` in general unless we actually want to use their data binding abilities and just serve SVG/PNG versions of the icons instead. I will have to do new perf testing given the `iron-list`, but I believe a big chunk of the rendering time is still setting those up. This is sufficient performance for now, but something to be aware of moving forward.

PR #209, opened 2016-06-29
Plotter style guide compliance.

adamcw:
Moving `labrad-plot.ts` closer to full style guide compliance.
- Removed all underscore suffixes on private variables.
- Changed `Array<T>` to `T[]` syntax.
- Removed use of the public keyword.
- Added `default` clauses to all `switch` statements, and fixed indenting.


adamcw:
The indenting in this file leaves much to be desired, but given that will flag a ton of lines, I will do that in another PR after this one.

maffoo:
LGTM

Issue #183, opened 2016-06-06
RMB initiates zoom window, leaves in funny state


joshm-3b6b58:
When you right click on a plot in the grapher, it will bring up the normal modifiy menu, but it will also initiate the zoom window. Pressing escape dismisses the modify menu, but also leaves the square for the zoom to window. 

Should be a fairly easy fix to check for which mouse button was pressed. Of course there may be dumb cross platform mac alt key issues.... :/

PR #210, opened 2016-06-29
zoomRectangle to only trigger on left mouse click.

adamcw:
Fixed issue reported in #183 where the zoom rectangle would end up in a strange state after a right-click.

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -4,7 +4,7 @@ import {viridisData} from '../scripts/colormaps';
 import * as datavault from "../scripts/datavault";
 
 
-const MOUSE_LEFT_BUTTON = 0;
+const MOUSE_MAIN_BUTTON = 0;
 
 
 /**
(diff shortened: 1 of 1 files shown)


client-js/app/elements/labrad-plot.ts
```diff
@@ -4,6 +4,9 @@ import {viridisData} from '../scripts/colormaps';
 import * as datavault from "../scripts/datavault";
 
 
+const MOUSE_LEFT_BUTTON = 0;
```
maffoo: Maybe call this `MOUSE_MAIN_BUTTON` with a comment that the main button is usually the left, but users can configure this (this aligns with [mozilla's docs](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button)).

I gotta say, it's annoying that the web apis have so many magic numbers and don't define standard symbolic names for these kind of things, but such is life.

maffoo:
One minor comment; otherwise LGTM.

adamcw:
Not sure how we both missed the `console.log` but I've removed that too.

Issue #164, opened 2016-03-09
Grapher: I can't command + click to open grapher links in a new tab


JulianSKelly:
This would be really nice.

joshm-3b6b58:
So, in the meantime right click and "Open in new tab" works. But I agree this has to be implemented correctly. Cross platform hotkeys is something we have to work on in general.

PR #211, opened 2016-06-29
Enabled cmd/ctrl click.

adamcw:
Fixed issue #164 where `cmd+click` and `ctrl+click` didn't have their default behavior of opening in a new tab.

Resolved by stopping the `tap` event from capturing the `click` event when `meta` or `ctrl` is held.

### DIFF
--- a/client-js/app/elements/app-link.html
+++ b/client-js/app/elements/app-link.html
@@ -1,6 +1,6 @@
 <script>
 (function () {
-  const MOUSE_LEFT_BUTTON = 0;
+  var MOUSE_MAIN_BUTTON = 0;
 
   Polymer({
     is: 'app-link',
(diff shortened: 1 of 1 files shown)


client-js/app/elements/app-link.html
```diff
@@ -1,5 +1,7 @@
 <script>
 (function () {
+  const MOUSE_LEFT_BUTTON = 0;
```
maffoo: Probably shouldn't use const here since this is just inline javascript, so we're not transpiling this to ES5 like we do with the typescript and I'm not sure about browser support of const. Of course, we can and probably should define this element in typescript like the other elements, in which case bringing const back would be fine.

maffoo:
One small comment, otherwise LGTM.

client-js/app/elements/app-link.html
adamcw: Ahh, a good point. The mix of TypeScript and JavaScript trips me up constantly. I will change this to `var` and plan to transition JS to TS in separate pull requests as I encounter it.

adamcw:
Also changed `MOUSE_LEFT_BUTTON` to `MOUSE_MAIN_BUTTON`.

Issue #145, opened 2016-02-08
Yellow traces are difficult to see


JulianSKelly:
![screen shot 2016-02-08 at 9 56 00 am](https://cloud.githubusercontent.com/assets/8152707/12894196/3c91ebbe-ce4a-11e5-8b14-db57a4f7e050.png)

adamcw:
With the white background and the darker yellow that seems to be in use compared to this screenshot, can we consider this resolved? Else, I can change the yellow to a more orange color if people would prefer. 

I would also perhaps recommend instead using colors that are more appropriate for color-blind users if we are already in the process of altering the colors:

http://mkweb.bcgsc.ca/biovis2012/

![Color Blind Palette](http://mkweb.bcgsc.ca/biovis2012/color-blindness-palette.png)

joshm-3b6b58:
I think we should make the line weight a little thicker as well.

PR #212, opened 2016-06-29
Updated 1D plots to be easier to read.

adamcw:
Fixes issue #145 by updating the 1D plots in grapher to use the same default colors as Google Charts as well as increasing the stroke width to 1.5 (up from 1.0).

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -6,14 +6,20 @@ import * as datavault from "../scripts/datavault";
 
 /**
  * Colors for traces in 1D plots.
+ * Chosen to contrast well even for color blind invididuals.
+ * Reference: http://mkweb.bcgsc.ca/biovis2012.
+ * GitHub Discussion: issues/145.
  */
 const COLOR_LIST = [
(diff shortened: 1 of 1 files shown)


adamcw:
I'm not sure that these colors are that visually appealing, we may want to consider a future enhancement to allow swapping between a color-blind and non-color-blind mode. People spend a lot of time looking at graphs, we don't want them to be ugly.

joshm-3b6b58:
Yeah these colours... 

Can we just use the Google chart colors? https://developers.google.com/chart/

adamcw:
I have changed the colors to the Google chart colors. They all start looking rather the same after six or so, clearly a pallet chosen for bar/pie chart graphics where you only need neighbors to be differentiated.

joshm-3b6b58:
Beyond six traces it's a little hard to draw any conclusion anyways, so I wouldn't worry about it. I like this scheme, we can always add more shades or whatever if people complain. 

@maffoo how many colors did the delphi grapher have anyways?

This LGTM

jwenner:
The Delphi grapher has 6: blue, red, green, cyan, purple, brown

Issue #131, opened 2016-01-09
Grapher: y-axis label can overlap tick labels


jwenner:
nit: The spacing of the y-axis label from the y-axis is such that the y-label can overlap tick labels. As an example, see WEBCOMPUTER/dataset/Jim/AliceBobRes/Cooldown151230/160107/00020/.

PR #213, opened 2016-06-29
Fix issue where y-axis ticks could overlap y-axis labels.

adamcw:
Fixed issue from #131 where the y-axis label could overlap with the y-axis ticks when the zoom level caused the ticks to get large.

Implemented as increasing the left-hand margin and setting the formatting of the y-axis ticks to use exponential notation.

This fix will still overlap if the exponent is greater than +/-99.


adamcw:
This is a simple approach to fixing the problem. Ideally it could dynamically resize the margin as you zoomed in and out, but I'm not sure if that would be disorienting for the user. I think this might be a good short-term fix until something a little more robust can be implemented.

maffoo:
I like the simple fix, but I'm not sure that 2 digits of precision is enough. For example, if you zoom in on a smallish region not centered on 0, then all the tick labels can show the same thing, and you have to do something like mouse over to actually see the coordinates. Maybe we should increase this somewhat, to 4 or 5 digits of precision? I think that will work a little better until we can implement a smarter layout scheme.

adamcw:
Hmm, I did wonder if 2 digits would be sufficient but worried this would lead to an unsightly margin. I will increase it and see how it goes. I believe resizing may not actually be that hard, so hopefully this is only a short term issue.

I did try to replace the `e+10` etc with a superscript, but this appeared to be harder than anticipated with how `d3` handles the axis binding. That approach would save a little space though.

adamcw:
I increased the precision to 5 digits, PTAL.

maffoo:
LGTM

Issue #154, opened 2016-02-19
Grapher: Live view scrunches datasets together


jwenner:
In live view, the datasets can be scrunched together such that the axes for the latest two datasets are covered up and the current position tags overlap. An example of this is at /grapher/Jim/AliceBobRes/Cooldown151230/160215/?live:

![image](https://cloud.githubusercontent.com/assets/8183919/13188007/a855a46e-d703-11e5-956d-b0d1278c5f01.png)

I believe this is the cause of @joshmutus's problem at #150. Note this is on Windows w/ Chrome.

PR #214, opened 2016-06-29
Fix issue where Grapher live view plots could overlap.

adamcw:
Fix for issue #154 where the plots on the Grapher live view could overlap if the page was less than the `450*num_plot` as each plot and its label require a minimum height to display.

The original implementation of `flex` would make each plot consistently 33% of the page when there were three plots, inconsiderate of minimum plot height.

This fix implements a minimum height, as well as applies the redraw function when plots are added so that plots are appropriately resized when new plots are added.

### DIFF
--- a/client-js/app/elements/labrad-grapher-live.html
+++ b/client-js/app/elements/labrad-grapher-live.html
@@ -5,12 +5,8 @@
     <style>
       :host {
         display: block;
-        height: 100%;
       }
       #plots {
-        display: flex;
-        flex-direction: column;
-        height: 100%;
       }
     </style>
     <div id="plots"></div>
@@ -21,12 +17,9 @@
   <template>
     <style>
       :host {
-        width: 100%;
-        display: flex;
-        flex-direction: column;
+        min-height: 450px;
       }
       #plot {
-        height: 400px;
         display: flex;
       }
     </style>
--- a/client-js/app/elements/labrad-grapher-live.html
+++ b/client-js/app/elements/labrad-grapher-live.html
@@ -5,8 +5,12 @@
     <style>
       :host {
         display: block;
+        height: 100%;
       }
       #plots {
+        display: flex;
+        flex-direction: column;
+        height: 100%;
       }
     </style>
     <div id="plots"></div>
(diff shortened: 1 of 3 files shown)


joshm-3b6b58:
Are the live plots working for you? They're not working on my machine, everything in "live view" shows up empty when I start a new dataset. Also, live plots that do show up aren't scaling correctly for me. ~~~I've checked out some older commits and it seems to be broken as far back as 399f9dd for me...~~~

Edit: I must've broken it in 7b4f8a67 :(

adamcw:
Live plots work just fine for me on master? That's where I branched this from.

joshm-3b6b58:
Weird. They are not working for me at all. Let me do a gulp clean

joshm-3b6b58:
OK I don't know what's going on, it works fine in one chrome window but not another. Pay no attention to my comments :|

joshm-3b6b58:
I'm getting the following error on compilation:

```
app/scripts/activities.ts(235,9): error TS2341: Property 'plot' is private and only accessible within class 'DatasetActivity'.
app/scripts/activities.ts(235,9): error TS2341: Property 'redraw_' is private and only accessible within class 'Plot'.
```

adamcw:
I guess those need to be public then, will fix.

adamcw:
Made the required functions public, PR now compiles without errors or warnings.

joshm-3b6b58:
So the formatting changes look really nice, but live view is utterly broken on my machine in weird ways. @maffoo does it work for you? I've managed to really confuse myself.

adamcw:
To confirm, live view is broken on more than just this branch for you? What is "broken" about it?

joshm-3b6b58:
It's broken in master.

When I got to live view and start a dataset running the nothing renders in the current plot. 

When I got to that specific dataset it shows points updating, but autoscales the graph wrong.

maffoo:
I've had some strange behavior in the past that was fixed by completely removing the `bower_components` folder and rerunning `npm run bower install`. @joshmutus, you might try that.

maffoo:
LGTM

Issue #147, opened 2016-02-09
Chrome link preview covers last row


jwenner:
The link preview on Chrome (showing the link in the bottom left corner) covers up the last row of text. This is particularly noticeable on the Node page, where it covers up the name of the Serial Server.

maffoo:
@jwenner, I've noticed this, but if you actually mouse over the link preview, it moves out of the way, so it doesn't cover a link you are actually trying to click. Does that make this less of a problem for you?

jwenner:
Maybe less of a problem, but still not negated. Would it be possible to add one blank line at the bottom of the page (likely outside the table) to allow for anything like this bar?

DanielSank:
I agree that this is annoying and has confused me several times. Definitely not a priority IMHO though.

adamcw:
There is no way to disable the status bar in Google Chrome.

The only "fix" for this issue would be to add a fixed footer at the bottom of the page, but that's essentially just permanently covering whatever is at the bottom of the screen, instead of temporarily.

You can't just add padding to the bottom of the page, because that doesn't stop the issue occurring when the page is longer than the window, as the padding wont even be on the page.

I'm marking this wont-fix, but perhaps it is an argument to increase browser support to at least Firefox (Issue #93) so users have some flexibility to work around browser features (such as this status bar) that they don't like.

jwenner:
@adamcw, what I was imagining was just adding a blank line at the bottom. While I agree that this wouldn't fix if the page is longer than the window, I (and probably 99.999%+ of users) am used to scrolling down in such a case. The real issue is that, for the last line, I can't scroll down to make the last line always visible.

adamcw:
@jwenner I'm not sure I like implementing a footer for the sake of a browser feature, but I see your point. I'll make a PR with the change and see what the others think.

PR #215, opened 2016-06-29
Add a small footer to the UI so that important information isn't obscured by browser status bars.

adamcw:
Fixes issue #147 by adding a small 25px footer to all pages so that any scrollable element cannot be obscured by the status bar that appears when hovering over links in browsers such as Google Chrome.

This change also begins to reduce the reliance on absolute positioning, moving towards a more consistent flex-based approach to positioning.

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -177,12 +177,13 @@ export class Plot extends polymer.Base {
             .on('zoom', () => this.handleZoom_());
 
     // Plot area.
+    const marginLeft = p.margin.left;
+    const marginTop = p.margin.top;
     p.svg = d3.select(area)
-            .append('svg:svg')
-            .attr('width', width + p.margin.left + p.margin.right)
-            .attr('height', height + p. margin.top + p.margin.bottom)
-            .append('g')
-            .attr('transform', `translate(${p.margin.left}, ${p.margin.top})`);
+              .append('svg:svg')
+                .attr('class', 'flex')
+                .append('g')
+                  .attr('transform', `translate(${marginLeft}, ${marginTop})`);
 
     // Background rectangle.
     p.svg.append('rect')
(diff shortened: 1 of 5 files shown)


maffoo:
A couple of things I noticed:
1. In the plot view the content + footer is now slightly too big for the available space, so the view scrolls slightly. This is not a huge deal in itself, but it's different from before and it would be nice to understand why. Also, this causes the footer to appear slightly smaller in the dataset view compared to the grapher view, since when the page first renders the footer is partially offscreen. 
2. On the registry UI, if the registry listing is longer than the browser window so that you have to scroll to see it, the footer doesn't display at all. I haven't dug into what's going on here, but given that you had to modify the grapher element, it might be that other content elements also need to be modified. It would be unfortunate if this were the case.

adamcw:
Ahh, I didn't catch those edge cases. I will look into it. 

The styling is a little delicate as it stands, so it's pretty easy to break things with seemingly innocent changes. I want to do something about that, but I haven't quite come up with a decent plan yet. 

Style scopes are a good idea when you want independent shadow-DOM components, but when trying to design a unified single-page web application, it mostly leads to needing similar CSS in multiple places that makes keeping things uniform difficult.

Will probably need a conscious day or two of code-health to clean up the CSS across the board and refactor out common styling into generic modules.

adamcw:
I couldn't replicate what you mentioned in 1, can you please check if the latest changes fix it for you?

I rebased the branch with Master to get the Live View stacking changes.

client-js/app/elements/labrad-plot.ts
```diff
@@ -177,12 +177,13 @@ export class Plot extends polymer.Base {
...
+    const marginLeft = p.margin.left;
+    const marginTop = p.margin.top;
     p.svg = d3.select(area)
-            .append('svg:svg')
-            .attr('width', width + p.margin.left + p.margin.right)
-            .attr('height', height + p. margin.top + p.margin.bottom)
-            .append('g')
-            .attr('transform', `translate(${p.margin.left}, ${p.margin.top})`);
+              .append('svg:svg')
+                .attr('class', 'flex')
```
adamcw, line 213: The indenting change might make this hard to see, but I removed width and height here on purpose so that the SVG can resize with its parent container.

maffoo:
LGTM

Issue #187, opened 2016-06-14
Show error message when datavault server is not connected


maffoo:
Currently, the UI just doesn't render if you go to a `/grapher/` link and the datavault server is not connected.

adamcw:
@maffoo I'm closing this issue. I'd recommend you create a new one regarding auto-reconnect features to be more explicit with the request and we'll mark it v2.1 or similar.

PR #216, opened 2016-06-29
Display error dialog on Grapher when Data Vault is unavailable.

adamcw:
Fixes issue #187 where the Grapher just silently failed when the Data Vault was unavailable (except for a console error). 

Added a dialog box that displays the response from the Labrad server, as well made it so that the Grapher UI still renders somewhat nicely respite there being no data available.

### DIFF
--- a/client-js/app/elements/labrad-grapher.html
+++ b/client-js/app/elements/labrad-grapher.html
@@ -197,5 +197,14 @@
         </paper-header-panel>
       </div>
     </div>
+    <paper-dialog id="errorDialog" modal>
+      <h1>Unable to connect to Data Vault</h1>
+      <div>
+        <span>{{connectionError}}</span>
+      </div>
+      <div class="buttons">
+        <paper-button dialog-dismiss>Close</paper-button>
+      </div>
+    </paper-dialog>
   </template>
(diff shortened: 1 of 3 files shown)


maffoo:
We have a similar problem for the dataset activity and grapher live view activity. Since people may just be opening to a url directly, rather than clicking through the grapher UI, it would be good to have a decent message in all those cases. I was actually imagining something similar to what happens now if the server can't connect to the manager, where the whole UI is replaced with an error message that can't be dismissed and the system tries in the background to connect so that if the datavault server connects, the UI will come up automatically. We'd probably not want to block the entire UI, since the user should be able to navigate to other parts of the web interface (the registry, say), but otherwise it would work basically like manager connection problems.

adamcw:
The non-blocking aspect was why I added a cancel button to the modal.

The reconnect feature is nice, but the issue was specifically just asking for a modal, so I kept this PR simple and to the point at this stage.

If this PR is approved I plan to do similar things for the dataset and grapher live activities. I may even factor out this shared functionality as it needn't be copied so many places.

maffoo:
Ok, LGTM. This is definitely an improvement over the status quo.

Issue #90, opened 2015-10-29
Add colorbar to 2D graphs


maffoo:

jwenner:
What's the status of this? Thanks.

joshm-3b6b58:
It's not being worked on. Shouldn't too too difficult to add. What priority
would you rank this with the other issues listed?

maffoo:
I think @ccwa is working on this, actually. Perhaps she can weigh in on the
current status.

joshm-3b6b58:
Marking this as a bug since it's needed to reach feature parity.

adamcw:
As I have never seen the Delphi grapher in action, can someone describe this feature so I can see how difficult it may be to implement?

adamcw:
If @ccwa has made progress, it would be great to see as I will likely need to port it to use WebGL once my branch is merged into master.

jwenner:
@adamcw, I've included a screenshot of a 2D plot from the Delphi grapher. The colorbar is shown to the right (with the tick marks for 100, 200, 300, 400, 500). This provides a way to associate colors on a 2D plot with the values of the dependent variable which is plotted.

In addition, note that the limits on the colorbar dynamically update as new data are added. In particular, as the limits of the dependent variable are widened due to data being outside the old limits, the colorbar ticks and coloring is updated accordingly.

I hope this helps. Let me know if you have more questions.

![image](https://cloud.githubusercontent.com/assets/8183919/16393268/ef32d570-3c64-11e6-8d29-841a7be19c5b.png)

adamcw:
So it's essentially an axis for the third variable, the color?

adamcw:
Turns out setting a base href on a page messes with SVG linking, and it took multiple hours to diagnose why the gradient on my SVG wasn't showing up...

This looks fairly straight forward though, at least in the realm of d3/SVG. I think this will be a good short term solution to get to feature parity sooner rather than later.

Issue #185, opened 2016-06-06
make grapher faster for large datasets


joshm-3b6b58:
It really bogs down, esp in 2D. 

@maffoo is this a bug or an enhancement?

Others, are speed rendering issues hampering y'all?

maffoo:
I would call this an enhancement, but it's definitely important to make things more usable for people.

jwenner:
The rendering speed would be hampering me if I were primarily using the web grapher. But, I'm still primarily using the Delphi grapher.

> is this a bug or an enhancement?

Yes. (i.e., both)

adamcw:
This is turning into a major rewrite/restructure of labrad-plot.ts. Keep this in mind if you decide to make changes to this file, as the resolution will be non-trivial.

adamcw:
This issue will also resolve #208 by removing the need for a clipping plane at all.

PR #219, opened 2016-06-30
Add a color bar axis to the Grapher for 2D plots.

adamcw:
Fixes issue #90 where there was no axis for the colors displayed on 2D plots using the Grapher.

Added a new element to 2D plots showing the z-axis as a color gradient. The scale adjusts as new data is introduced. Size and positioning can be tweaked via constants in the file.

Also moved the styling for the `zoomRectangle` from global CSS, to a local fill color on the SVG element. This was done because otherwise all rectangles had to be the same color.

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -26,6 +26,18 @@ const COLOR_MAP = viridisData.map((rgb) => {
 });
 
 
+const COLOR_BAR_WIDTH = 15;
+const COLOR_BAR_LEFT_MARGIN = 15;
+const COLOR_BAR_RIGHT_MARGIN = 5;
+const COLOR_BAR_AXIS_WIDTH = 40;
+const COLOR_BAR_STROKE_WIDTH = 0.75;
+const COLOR_BAR_OUTER_WIDTH = (
+  COLOR_BAR_LEFT_MARGIN + COLOR_BAR_WIDTH + COLOR_BAR_RIGHT_MARGIN
+);
+const COLOR_BAR_WIDGET_SIZE = (
+   COLOR_BAR_OUTER_WIDTH + COLOR_BAR_AXIS_WIDTH
+);
+
 @component('labrad-plot')
 export class Plot extends polymer.Base {
 
@@ -83,7 +95,7 @@ export class Plot extends polymer.Base {
   };
   private margin = {
     top: 50,
-    right: 80,
+    right: 10,
     bottom: 50,
     left: 40
   };
@@ -137,8 +149,9 @@ export class Plot extends polymer.Base {
       area: HTMLElement, totWidth: number, totHeight: number): void {
     const p = this;
 
-    if (p.numIndeps == 1) {
-      p.margin.right = 10;
+    // Make room for the color bar if necessary.
+    if (p.numIndeps == 2) {
+      p.margin.right += COLOR_BAR_WIDGET_SIZE;
     }
 
     const width = totWidth - p.margin.left - p.margin.right;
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-plot.ts
```diff
@@ -135,6 +149,11 @@ export class Plot extends polymer.Base {
       area: HTMLElement, totWidth: number, totHeight: number): void {
     const p = this;
 
+    // Make room for the color bar if necessary.
+    if (p.numIndeps == 2) {
+      p.margin.right += COLOR_BAR_WIDGET_SIZE;
```
maffoo: This changes the margin every time `createPlot` is called, which can happen multiple times. The margin gets larger and larger over time and the plot smaller and smaller, which is a fun effect but probably not what we want :-)

client-js/app/elements/labrad-plot.ts
```diff
@@ -222,6 +247,55 @@ export class Plot extends polymer.Base {
...
+          .attr("x1", "0%")
+          .attr("y1", "0%")
+          .attr("x2", "0%")
+          .attr("y2", "100%")
+          .selectAll("stop")
+          .data([
+            {offset: "0%", color: COLOR_MAP[0]},
+            {offset: "50%", color: COLOR_MAP[128]},
+            {offset: "100%", color: COLOR_MAP[255]}
+          ])
```
maffoo: Is this relying on d3 to do the interpolation between `COLOR_MAP` values? That's probably ok, though the matplotlib guys who spent all that time carefully designing the viridis map would probably be sad. It might be nice to use a few more points, just to make improve the fidelity of the gradient and the actual color map (note that this gradient data can be precomputed once when the color map data is first loaded).

client-js/app/elements/labrad-plot.ts
```diff
@@ -222,6 +247,55 @@ export class Plot extends polymer.Base {
...
+          .enter().append("stop")
+          .attr("offset", function(d) { return d.offset; })
```
maffoo: `(d) => d.offset`

client-js/app/elements/labrad-plot.ts
```diff
@@ -222,6 +247,55 @@ export class Plot extends polymer.Base {
...
+          .attr("stop-color", function(d) { return d.color; })
```
maffoo: `(d) => d.color`

client-js/app/elements/labrad-plot.ts
```diff
@@ -222,6 +247,55 @@ export class Plot extends polymer.Base {
...
+          .attr("stop-opacity", 1);
+
+      // Appending the location href is necessary due to the use of `base href`
+      // for the overall app to make Polymer paths work.
+      const gradientFill = "url('" + location.href + "#ColorBarGradient')";
```
maffoo: can use string interpolation:

``` typescript
const gradientFill = `url('${location.href}#ColorBarGradient')`;
```

maffoo:
Overall, the colorbar widget looks nice, but I think the gradient is inverted. Bright yellow goes with zMax, dark blue with zMin, but it's being drawn the other way around.

client-js/app/elements/labrad-plot.ts
adamcw: This is relying on the SVG implementation in the browser to do a uniform linear gradient.

I forgot that Viridis went to special lengths to make the gradient uniform in color. I will sample more points and see how that goes.

The inefficiency you mention is true for everything in `createPlot` that's why for #185 it only gets called once, so I'm not too concerned about this part in particular at this stage.

client-js/app/elements/labrad-plot.ts
adamcw: Ahh, of course that's what it does. Not sure how I missed that after the refactor. Done.

adamcw:
I've improved the color gradient, flipped it around, and also increased the number of ticks on the scale. PTAL.

maffoo:
LGTM

Issue #208, opened 2016-06-28
Zoomed plots render outside chart


joshm-3b6b58:
This is on current master.

Load a plot, looks like this:
![image](https://cloud.githubusercontent.com/assets/8173925/16426658/def8f896-3d1e-11e6-9421-217620ce8647.png)

Zoom into area of interest, plot renders outside the chart:
![image](https://cloud.githubusercontent.com/assets/8173925/16426682/f337f2c6-3d1e-11e6-8f4e-706f720f23ba.png)

joshm-3b6b58:
Also happens when plots are dragged :(
![image](https://cloud.githubusercontent.com/assets/8173925/16426827/7edead60-3d1f-11e6-9609-0631eb98656a.png)

joshm-3b6b58:
The code that should keep the graph from overflowing the chart is here: https://github.com/labrad/scalabrad-web/blob/master/client-js/app/elements/labrad-plot.ts#L379

joshm-3b6b58:
Weirdly things render fine on the current version of matrix-reloaded.

I hope this doesn't have something to do with updating D3...

Edit: I reverted D3 to 3.5.6 on my machine and still see the issue

adamcw:
This was a known bug before the updating of D3.

It it is fixed in my WebGL branch so I didn't bother making an issue for it, will be fixed once that gets through review.

(WebGL doesn't use a clipping frame at all, so this bug was basically fixed for free)

joshm-3b6b58:
Well it doesn't happen in 7b4f8a67500 either, I'm trying to track down where it popped up

Issue #184, opened 2016-06-06
Copy graph to clipboard 


joshm-3b6b58:
No more print screening would be lovely.

adamcw:
I've looked into this with my new WebGL/SVG-hybrid plot renderer, and initial outcomes look a little grim. You can save the graph portion, but not the axes. This might be possible by converting the SVG portion to Canvas, and then rendering the WebGL portion within the Canvas, but that would be a lot of work for this single feature. It also may completely tank the performance.

Ctrl+Shift+4 on Mac, or Ctrl+Prnt Screen on Linux is easy enough that I'm not sure this is likely to be a high priority unless there is something I'm missing?

joshm-3b6b58:
I think this is a high priority. 

The old grapher didn't support export in any sensible way and one of the most frequent use cases was exporting plots into other docs and emails. Sure print screen works, but it makes the grapher seem un-polished and costs users lots of time in aggregate.

adamcw:
I have investigated fairly thoroughly, and cannot see a solution that is both simple and performant.

The only approach to allowing this feature while maintaining adequate performance seems to be rewriting all the SVG plot functionality into WebGL directly. 

The primary method that I explored was that you can turn SVG -> Blob -> Object URL -> Image -> Texture and then render a Plane with the texture of your SVG. 

The problems with this is many, but the biggest being that this translation process takes a considerable amount of processing time. It works well for a mostly static HUD, but when it needs to be updated every frame during a scroll/pan or as new data is added, it doesn't work out well.

Recreating the plot in pure WebGL is definitely possible, it's mostly a matter of it probably being at least couple of weeks dedicated development, which I think could be better spent working on features such as #221 and #229, among others at this point in time.

adamcw:
New idea... we render as both SVG and WebGL, but have a screenshot button that merges the two for a single frame into a canvas and then downloads that. That could work!

PR #223, opened 2016-07-12
WebGL Implementation of main plotting interface in Grapher.

adamcw:
Resolves issues #185 and #208. Makes use of WebGL to do the data plotting portion of each graph in order to greatly improve the performance of live data view as well as zoom/pan operations.

`d3.js` was retained for the axes and and interface as SVG, while `THREE.js` library was introduced to handle the plotting of the data via WebGL.

This dual library usage was used to get the most performance gain for the least development time, as implementing axes and other features in WebGL will be time consuming. It is a longer term goal to move all plotting towards WebGL in order to allow screenshotting (Issue #184).

Performance was also improved by causing only the parts of the graph that need to be re-rendered under actions like zooming to be re-rendered. This also makes it possible to now retain the zoom level during live updates, the changing of plotting types, and the toggling of traces.


client-js/app/elements/labrad-plot.ts
```diff
@@ -1,9 +1,17 @@
 import 'd3';
-
+import * as three from 'three';
 import {viridisData} from '../scripts/colormaps';
 import * as datavault from "../scripts/datavault";
 
+/**
+ * Import the THREE namespace from the module.
+ */
+const THREE = three.default;
```
adamcw: This throws a compiler error, but was the only way I could get the library to play nicely with TypeScript. If someone can work out how to import the library without this hack, please let me know, I wasted a few hours trying and ultimately moved on.

client-js/app/elements/labrad-plot.ts
```diff
@@ -351,14 +583,111 @@ export class Plot extends polymer.Base {
...
+      }
+
+      if (dataPoints.length > 1) {
+        const geometry = new THREE.BufferGeometry();
+        geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+        geometry.addAttribute('data', new THREE.BufferAttribute(dataPoints, 2));
+
+        let material = new THREE.LineBasicMaterial({color: COLOR_LIST[i % COLOR_LIST.length]});
+        let mesh = new THREE.Line(geometry, material);
+        // TODO [awhiteside]: Increase line width to retain parity with SVG
```
adamcw: Oops, left a TODO in.

client-js/app/elements/labrad-plot.ts
```diff
@@ -351,14 +583,111 @@ export class Plot extends polymer.Base {
...
+      }
+    }
+
+    this.sceneObjects.push(ob);
+    this.scene.add(ob);
+  }
+
+
+  /**
+   *
```
adamcw: A blank comment.

client-js/app/elements/labrad-plot.ts
```diff
@@ -351,14 +583,111 @@ export class Plot extends polymer.Base {
...
+    ob.add(mesh);
+
+    this.sceneObjects.push(ob);
+    this.scene.add(ob);
   }
 
 
-  private plotData1D(data: number[][], lastData?: number[]) {
+  /**
+   *
```
adamcw: Another blank comment.

client-js/app/elements/labrad-plot.ts
```diff
@@ -385,43 +714,36 @@ export class Plot extends polymer.Base {
...
-      }
-    }
   }
 
 
-  private plotData2D(data: number[][], lastData?: number[]) {
-    // Update data limits.
-
+  /**
+   *
```
adamcw: Another blank comment, now this is just embarrassing.

adamcw:
I have mostly created this PR to start a dialog on breaking it up further. I figured this would be best done as a semi-team process as this sort of thing is likely to be common as large scale refactors/redesigns become required throughout the codebase(s).

This is ultimately the smallest PR that maintains feature parity as well as builds/runs. This opens the question of if stacking smaller, non-building, PRs may be preferred where they are only merged once all have been reviewed and approved. I don't believe this concept is currently an accepted workflow so I thought a discussion first may be useful.

client-js/app/elements/labrad-plot.ts
maffoo: The problem here is that the typescript definition files for these libraries (and most libraries actually) were set up assuming they would be imported using CommonJS style imports like in node, rather than es6 imports. To fix this, you have to change the typings file slightly, like so: https://github.com/labrad/scalabrad-web/commit/7096169119bc2ad8ff4c949d9adeda028a0f0a6e

maffoo:
I get a couple of typescript compilation errors, but they're not too hard to fix: https://github.com/labrad/scalabrad-web/commit/e341a3387cf67cc01ca6a6565a458ef979f8f903

adamcw:
Interesting, I do not get any compilation warnings. This might suggest we have different versions of TypeScript running?

client-js/app/elements/labrad-plot.ts
adamcw: Yes, I did track down that I could probably fix it by changing the typings file, but I didn't like the idea of changing it manually. Doing this will lead to problems whenever someone tries to update the typings in the future without knowing to update the format of the exports.

maffoo: I agree it's not ideal to keep these local changes, but this is the reason why we've got the `.d.ts` definition files checked into the repo, rather than installing them using one of the tools to manage typescript definitions (originally it was `tsd`, then `typings`, now the latest thing is to install types packages from npm like [`@types/three`](https://www.npmjs.com/package/@types/three), who knows what it will be next month...). In practice we don't update these definitions very often, but if we do make an update and forget to change the module declaration, at least we get compiler errors to remind us to fix it :-)

client-js/app/elements/labrad-plot.ts
```diff
@@ -154,49 +240,129 @@ export class Plot extends polymer.Base {
...
+  attached() {
+    this.render();
+    window.addEventListener('resize', (e) => this.resizePlot());
+  }
+
+
+  /**
+   * Resets the current zoom level to fit the data.
+   */
+  resetZoomControl() {
```
maffoo: Looks like this is not referenced anywhere.

client-js/app/elements/labrad-plot.ts
```diff
@@ -134,13 +143,90 @@ export class Plot extends polymer.Base {
...
+   * moving the camera around the 3D space provided by `THREE.js`. As such,
+   * camera properties such as position and zoom never change.
+   */
+  private camera = new THREE.PerspectiveCamera(
+    this.cameraOpts.fov,
+    this.cameraOpts.aspect,
+    this.cameraOpts.near,
+    this.cameraOpts.far);
+
+  /** The scene to render graphs to. */
```
maffoo: A lot of the comments here seem to just narrate the code and can probably be removed.

client-js/app/elements/labrad-plot.ts
```diff
@@ -154,49 +240,129 @@ export class Plot extends polymer.Base {
...
+   * Creates and renders the SVG and WebGL portions of the graph.
+   */
+  private render(): void {
+    if (this.isRendering) {
+      return;
+    }
+
+    this.isRendering = true;
+
+    // Initialize SVG portions of plot.
```
maffoo: Narration, can be removed.

client-js/app/elements/labrad-plot.ts
```diff
@@ -134,13 +143,90 @@ export class Plot extends polymer.Base {
...
+  };
+
+
+  /**
+   * A perspective camera to view the graph.
+   *
+   * The camera is currently only used for the sake of being able to observe
+   * the scene. Zooming/Panning is handled by `d3` rather than physically
+   * moving the camera around the 3D space provided by `THREE.js`. As such,
+   * camera properties such as position and zoom never change.
```
maffoo: You do update the camera aspect ratio in `resizeWebGLPlot`; might be good to note here that not _all_ camera properties are fixed.

client-js/app/elements/labrad-plot.ts
```diff
@@ -303,26 +451,111 @@ export class Plot extends polymer.Base {
...
+
+
+  /**
+   * Resize the SVG portion of the plot.
+   */
+  private resizeSVGPlot(): void {
+    const plot = this.$.plot;
+    const plotBounds = plot.getBoundingClientRect();
+    const plotWidth = Math.max(plotBounds.width, 400);
+    const plotHeight = Math.max(plotBounds.height, 400);
```
maffoo: Maybe define constants `PLOT_MIN_WIDTH`, `PLOT_MIN_HEIGHT` to use here.

client-js/app/elements/labrad-plot.ts
```diff
@@ -303,26 +451,111 @@ export class Plot extends polymer.Base {
...
+    this.svg.select('#rect')
+              .attr('width', this.width)
+              .attr('height', this.height);
+
+    this.svg.select('#x-axis')
+              .attr('transform', `translate(0, ${this.height})`)
+              .call(this.xAxis);
+
+    const xLabXOffset = this.width / 2;
+    const xLabYOffset = this.height + this.margin.bottom - 10;
```
maffoo: I was confused about "Lab" here, thought it meant "lab frame" or something like that. Maybe just say "Label" since other physicists may look at this code :-)

client-js/app/elements/labrad-plot.ts
```diff
@@ -154,49 +240,129 @@ export class Plot extends polymer.Base {
...
+    this.resizeSVGPlot();
+    this.resizeWebGLPlot();
+    this.projectGraphPositions();
+  }
+
+
+  /**
+   * Begins the scene rendering loop.
+   */
+  private renderScene(): void {
```
maffoo: The name and documentation here is a bit confusing because this does not just begin the render loop, this _is_ the render loop. Also, we need to shut down the render loop when this plot element is detached.

client-js/app/elements/labrad-plot.ts
```diff
@@ -351,14 +584,113 @@ export class Plot extends polymer.Base {
...
+      if (dataPoints.length > 1) {
+        const geometry = new THREE.BufferGeometry();
+        geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+        geometry.addAttribute('data', new THREE.BufferAttribute(dataPoints, 2));
+
+        let material = new THREE.LineBasicMaterial({
+          color: COLOR_LIST[i % COLOR_LIST.length],
+          linewidth: 1.5
+        });
+        let mesh = new THREE.Line(geometry, material);
```
maffoo: These last two could use `const` instead. Also, maybe call the latter `line` instead of `mesh`.

client-js/app/elements/labrad-plot.ts
```diff
@@ -351,14 +584,113 @@ export class Plot extends polymer.Base {
...
+    }
+
+    const geometry = new THREE.BufferGeometry();
+    geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+    geometry.addAttribute('data', new THREE.BufferAttribute(dataPoints, 2));
+    geometry.addAttribute('color', new THREE.BufferAttribute(colors, 3));
+
+    let material, mesh;
+    if (this.drawMode2D == 'dots') {
+      material = new THREE.PointsMaterial({size: 1, vertexColors: THREE.VertexColors});
```
maffoo: Single-pixel dots are almost invisible for me; maybe make these a bit larger?

client-js/app/elements/labrad-plot.ts
```diff
@@ -385,43 +717,36 @@ export class Plot extends polymer.Base {
...
 
 
-  private plotData2D(data: number[][], lastData?: number[]) {
-    // Update data limits.
-
+  /**
+   * Set the bounds for a 2D plot.
+   */
+  private dataLimits2D(data: number[][]) {
+    console.info("dataLimits2D", data.length);
```
maffoo: can probably remove.

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,206 +787,186 @@ export class Plot extends polymer.Base {
...
-                .attr('y', (d) => p.yScale(d[1]) - h)
-                .attr('width', w)
-                .attr('height', h)
-                .style('fill', (d) => getColor(d[p.displaySurface], zMin, zMax));
-        }
-        break;
+  /**
+   * Projects a coordinate in graph coordinates (relative to the axes of the
+   * data) to screen coordinates (relative to the canvas). Additionally
+   * determines the width and height of the coordinate based on its type.
```
maffoo: Should indicate that this mutates the `outputObject` parameter, rather than returning a value.

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,206 +787,186 @@ export class Plot extends polymer.Base {
...
-
-      default:
-        // Nothing to do.
-        break;
-    }
+  /**
+   * Projects a coordinate in screen coordinates (relative to the canvas) to
+   * world coordinates (relative to the 3D projection of the scene).
+   */
+  private projectScreenCoordToWorldCoord(x: number, y: number, outputVector: THREE.Vector3) {
```
maffoo: I would reorder to put this after `projectGraphPositions`, which describes the various coordinate systems.

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,206 +787,186 @@ export class Plot extends polymer.Base {
...
+    const vector = new THREE.Vector3();
+    const screenRect = { x: 0, y: 0, w: 0, h: 0 };
+
+    // Project the (0, 0) screen position to world coordinates so that widths
+    // and heights in world space can be calculated from the delta.
+    this.projectScreenCoordToWorldCoord(0, 0, vector);
+    const {
+      x: xWorldZero,
+      y: yWorldZero
+    } = vector;
```
maffoo: Looks strange to break a destructuring assignment across multiple lines like this; I'd suggest keeping it on one line.

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,206 +787,186 @@ export class Plot extends polymer.Base {
...
+
+    for (let obj of this.sceneObjects) {
+      for (let child of obj.children) {
+        let array = child.geometry.attributes.position.array;
+        let data = child.geometry.attributes.data.array;
```
maffoo: can use `const` for these.

client-js/app/elements/labrad-plot.ts
```diff
@@ -351,14 +584,113 @@ export class Plot extends polymer.Base {
...
+      for (let row of data) {
+        dataPoints[offset] = row[0];
+        dataPoints[offset + 1] = row[i + 1];
+        offset += 2;
+      }
+
+      if (dataPoints.length > 1) {
+        const geometry = new THREE.BufferGeometry();
+        geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+        geometry.addAttribute('data', new THREE.BufferAttribute(dataPoints, 2));
```
maffoo: It would be helpful to have a comment somewhere describing what is in the attributes added to the geometry for 1d and 2d plots and where this info is updated (for example, note that 'position' attribute is created here but values are set in `projectGraphPositions`.

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,206 +787,186 @@ export class Plot extends polymer.Base {
...
+  /**
+   * Queues zoom handling on the next animation frame.
+   */
+  private zoomEventHandler(): void {
+    if (!this.isZooming) {
+      requestAnimationFrame(() => this.handleZoom());
+      this.isZooming = true;
+      this.haveZoomed = true;
     }
   }
```
maffoo, line 1094: This could be written inline for added clarity. We'd need a little utility function that returns a promise for the next animation frame:

``` javascript
function requestAnimationFramePromise(): Promise<void> {
  const {obligation, promise} = obligate<void>();
  requestAnimationFrame(() => obligation.resolve());
  return promise;
}

...

private async handleZoom(): void {
  if (this.isZooming) return;
  this.isZooming = true;
  await requestAnimationFramePromise();
  // do zooming work
  this.isZooming = false;
}
```

client-js/app/elements/labrad-plot.ts
```diff
@@ -736,30 +1033,69 @@ export class Plot extends polymer.Base {
...
 
 
-  private mouseToDataX(x: number): number {
-    x = clip(x - this.margin.left, 0, this.width);
-    return this.xScale.invert(x);
+  /**
+   * Updates the control event listeners depeding on mode.
```
maffoo: typo: depending

client-js/app/elements/labrad-plot.ts
```diff
@@ -736,30 +1033,69 @@ export class Plot extends polymer.Base {
...
+   */
+  private updateControlEventListeners() {
+    if (!this.isRendering) {
+      return;
+    }
```
maffoo, line 1167: can write on single line without curlies, like line 1064 below.

client-js/app/elements/labrad-plot.ts
```diff
@@ -950,19 +1288,24 @@ export class Plot extends polymer.Base {
   }
 
 
-  @listen('plot.mousemove')
-  private listenPlotMouseMove(event) {
-    const rect = event.currentTarget.getBoundingClientRect();
+  @listen('canvas.mousemove')
+  private listenPlotMouseMove(e) {
```
maffoo: I would suggest calling this `onMouseMove` instead.

maffoo:
This is really awesome! I've made a number of comments, but they're mostly minor. We obviously need to sort out the compilation problems, but once I got that to work on my machine the performance is great, and certainly hugely improved over straight d3+svg.

maffoo:
As for how to get this merged, I think it might be possible to restructure this into a series of more incremental changes, but I don't really think it's worth the effort in this case. I'm fine with merging this as one big change once we get through the review.

I think when doing refactors it's best to keep things building on master at every change, rather than incremental but breaking changes. Sometimes this requires building some adaptation code to let things run in the intermediate state, which then gets torn out once the change is done, but I think it's worth the effort.

client-js/app/elements/labrad-plot.ts
adamcw: You mean attached()? or resetZoomControl()? 

attached() is called when the component loads.
resetZoomControl() is called from the html UI element.

client-js/app/elements/labrad-plot.ts
adamcw: I think it only looks strange because destructuring is new syntax. It's ultimately a necessary style format that is required when you have longer destructurings. 

If you think of it as each being a variable assignment, this formatting makes it consistent with actual variable declarations, which are always done on their own lines.

``` typescript
    const {
      x: xWorldZero,
      y: yWorldZero
    } = vector;
```

is the same as:

``` typescript
    const xWorldZero = vector.x;
    const yWorldZero = vector.y;
```

Now we mention it, maybe these are all better off as just straight variable declarations.

client-js/app/elements/labrad-plot.ts
adamcw, line 1094: `zoomEventHandler` is being called by `d3` on every `scroll` event. The purpose of `requestAnimationFrame` is to somewhat smooth the calls to avoid long frames and lag.

I'm not your code flow is any easier to understand, nor does it really support the fact that `handleZoom` is called from more places than `zoomEventHandler`.

client-js/app/elements/labrad-plot.ts
adamcw: I think I will leave this line as-is, then submit another PR making the required typings changes and then removing this line.

client-js/app/elements/labrad-plot.ts
maffoo: I meant `resetZoomControl()`. I don't see it referenced anywhere here or in the element html (https://github.com/labrad/scalabrad-web/blob/185-graphing-performance-webgl/client-js/app/elements/labrad-plot.html); am I missing something?

client-js/app/elements/labrad-plot.ts
maffoo, line 1094: I meant for this one function to replace `zoomEventHandler`, so there would just be one function. AFAICT there's only one place currently where `handleZoom` is called directly and that is in `resetZoom`, but presumably that should be protected by the `isZooming` check, too? Anyway, that was my thinking, to just have one entry point, but it's not a huge deal.

client-js/app/elements/labrad-plot.ts
adamcw: Ahh, I must have merged it poorly. The intent was to have `resetZoom` be private, then have `resetZoomControl` be the pubic hook the UI uses. 

It's basically a separation of the API the UI calls and the API the internal module calls.

Ultimately, it's a bit of a waste since it's just an alias, but the idea would be that the UI hooks can be altered independently, as well as having UI elements call descriptive names from the HTML, while internally calling a UI independent name.

client-js/app/elements/labrad-plot.ts
adamcw, line 1094: I went a different route and merged the zoom loop with the render loop, so now there is just one core loop occurring each frame which I think is much nicer. It also leads to slightly improved performance.

adamcw:
@maffoo Quite a few comments, so quite a few changes. PTAL.

I will submit the typings changes fixing both the hack and the other compiler warnings you mentioned in a separate PR.

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,214 +826,212 @@ export class Plot extends polymer.Base {
...
+            // When dealing with dots, simply copy the world coordinates
+            // into the position array at the correct offset.
+            positions[positionOffset] = xWorld;
+            positions[positionOffset + 1] = yWorld;
+          } else {
+            // Project the width and height in screen space to world space,
+            // then calculate the size as a delta from world zero.
+            this.projectScreenCoordToWorldCoord(wScreen, -hScreen, vector);
+
+            const wWorldEnd= vector.x,
```
maffoo: nit: spacing

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,214 +826,212 @@ export class Plot extends polymer.Base {
...
-        this.svg.call(this.zoom);
-        this.$.plot.style.cursor = 'auto';
-        this.$.plot.style.cursor = 'all-scroll';
-        break;
+  /**
+   * Projects a coordinate in graph coordinates (relative to the axes of the
+   * data) to screen coordinates (relative to the canvas). Additionally
+   * determines the width and height of the coordinate based on its type.
+   *
+   * Mutates `outputObject` returning the projected `x` and `y` coordinate as
```
maffoo: nit: maybe say "by setting" instead of "returning"

client-js/app/elements/labrad-plot.ts
```diff
@@ -470,214 +826,212 @@ export class Plot extends polymer.Base {
...
-    if (!this.svg) return;
-    switch (this.numIndeps) {
-      case 1:
-        this.$$('rect.background').style.fill = '#ffffff';
-        this.$.modes2d.style.visibility = 'hidden';
-        this.is1D = true;
-        this.is2D = false;
+      case 'rectfill':
+        wScreen = Math.abs(this.xScale(this.dx0) - this.xScale(0));
+        hScreen = Math.abs(this.yScale(this.dy0) - this.yScale(0));
```
maffoo: I think you need to modify y here, otherwise the rectfill and vargrid modes render uniform grids in different places (you can see this in the uniform-grid sample plots where the plot shifts vertically when switching between these two modes):

``` typescript
y += this.dy0
```

maffoo:
A few minor small comments, but mostly this LGTM. I agree it's a good idea to do the typescript changes to `.d.ts` files in a separate commit to not mix that in with the other changes here; let's merge that first so we can keep master building cleanly. Also, if you rebase this on master the tests and build should run on teamcity, so we'll see a clean build before merging.

adamcw:
Can you confirm what version of TypeScript you are running and the exact commands you are running to see the compiler errors regarding the typings? I still cannot see the errors, and I don't like the idea of submitting a PR blind.

maffoo:
Just did a clean `npm install` and here's what I get:

```
$ npm list | grep typescript
├─┬ gulp-typescript@2.13.6
├── typescript@1.8.10

$ npm run gulp

> labrad-browser@0.0.0 gulp /Users/maffoo/src/scalabrad-web/client-js
> gulp

[11:03:42] Using gulpfile ~/src/scalabrad-web/client-js/gulpfile.js
[11:03:42] Starting 'clean'...
[11:03:42] Finished 'clean' after 15 ms
[11:03:42] Starting 'default'...
[11:03:42] Starting 'compile-ts'...
error TS5023: Unknown compiler option 'experimentalAsyncFunctions'.
app/elements/labrad-plot.ts(9,21): error TS2339: Property 'default' does not exist on type 'typeof THREE'.
[11:03:44] TypeScript: 1 semantic error
[11:03:44] TypeScript: emit succeeded (with errors)
[11:03:44] Finished 'compile-ts' after 2.13 s
[11:03:44] Starting 'bundle'...
     Building the single-file sfx bundle for app/scripts/app...
ok   Built into .tmp/scripts/bundle.js, unminified.


[11:03:48] Finished 'bundle' after 3.49 s
[11:03:48] Starting 'copy'...
[11:03:48] Starting 'styles'...
[11:03:49] styles all files 382 B
[11:03:49] Finished 'styles' after 591 ms
[11:03:49] copy all files 26.17 MB
[11:03:49] Finished 'copy' after 1.42 s
[11:03:49] Starting 'elements'...
[11:03:49] Finished 'elements' after 6.58 ms
[11:03:49] Starting 'jshint'...
[11:03:49] Starting 'images'...
[11:03:50] Starting 'fonts'...
[11:03:50] Starting 'html'...
[11:03:50] Finished 'fonts' after 430 ms

app/elements/app-link.html
  line 20  col 35  Expected '{' and instead saw 'return'.
  line 28  col 35  Expected '{' and instead saw 'return'.

  ⚠  2 warnings

[11:03:51] html all files 1.19 kB
[11:03:51] Finished 'html' after 576 ms
[11:03:51] 'jshint' errored after 1.29 s
[11:03:51] Error in plugin 'gulp-jshint'
Message:
    JSHint failed for: app/elements/app-link.html
[11:03:51] 'default' errored after 8.34 s
[11:03:51] Error in plugin 'run-sequence(jshint)'
Error
    at finish (/Users/maffoo/src/scalabrad-web/client-js/node_modules/run-sequence/index.js:56:13)
    at Gulp.onError (/Users/maffoo/src/scalabrad-web/client-js/node_modules/run-sequence/index.js:67:4)
    at Gulp.emit (events.js:129:20)
    at Gulp.Orchestrator._emitTaskDone (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp/node_modules/orchestrator/index.js:264:8)
    at /Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp/node_modules/orchestrator/index.js:275:23
    at finish (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:21:8)
    at DestroyableTransform.<anonymous> (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:52:4)
    at DestroyableTransform.f (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp/node_modules/orchestrator/node_modules/end-of-stream/node_modules/once/once.js:17:25)
    at DestroyableTransform.emit (events.js:129:20)
    at DestroyableTransform.flush (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp-jshint/src/reporters/fail.js:42:14)
    at DestroyableTransform._flush (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp-jshint/src/stream.js:33:13)
    at DestroyableTransform.<anonymous> (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp-jshint/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:135:12)
    at DestroyableTransform.g (events.js:199:16)
    at DestroyableTransform.emit (events.js:129:20)
    at finishMaybe (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp-jshint/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:371:12)
    at endWritable (/Users/maffoo/src/scalabrad-web/client-js/node_modules/gulp-jshint/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:378:3)
[11:03:51] images all files 25.3 kB
[11:03:51] Finished 'images' after 1.09 s

npm ERR! Darwin 15.5.0
npm ERR! argv "node" "/usr/local/bin/npm" "run" "gulp"
npm ERR! node v0.12.4
npm ERR! npm  v2.10.1
npm ERR! code ELIFECYCLE
npm ERR! labrad-browser@0.0.0 gulp: `gulp`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the labrad-browser@0.0.0 gulp script 'gulp'.
npm ERR! This is most likely a problem with the labrad-browser package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     gulp
npm ERR! You can get their info via:
npm ERR!     npm owner ls labrad-browser
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/maffoo/src/scalabrad-web/client-js/npm-debug.log
```

The compiler option and jshint problems are fixed on master, so rebasing will pick up those changes, but there's still the typescript problem with the 'default' property on three.js.

joshm-3b6b58:
Looks awesome! I'm going thru and testing various UI stuff. 

When displaying a 2D plot with 'vargrid' or 'rect fill' the drag zooming rectangle doesn't render. I suspect it's showing up behind the plot.

joshm-3b6b58:
Have you tried loading larger 2D datasets? It behaves a little unpredictably for me. It'll load the first few columns right away, but won't stream the new data as it comes. I have to do some funny dance of zooming in and hitting the home button to get the plot to update.

adamcw:
@joshmutus 2D Datasets not working is a strange regression, I'm looking into it now.

The zoom rectangle will probably need to be rendered in WebGL rather than SVG. I could render it as a second SVG layer, but since we want to move everything toward WebGL anyway, may as well do it that way now.

joshm-3b6b58:
@adamcw Where you able to recreate the behaviour? I can send you the dataset that chokes up on my system if you like

adamcw:
I've fixed the rendering portion if that is what you refer to. I restructured the rendering pipeline due to comments and it meant that not everything was updated when it had to be after new data was added.

joshm-3b6b58:
@adamcw renders super nicely for large datasets now. Really really snappy. Thanks.

I have a weird problem, the plot doesn't show up when I'm looking at the app through remote desktop... None of the plot widget loads at all. Neither the axes or the data:
screenshot from 2016-07-13 16 02 42

Renders fine on the same machine, just not through Chrome remote desktop session:
screenshot from 2016-07-13 16 01 18

Any idea what could be happening there?

adamcw:
That error is when WebGL is disabled in the browser (I've encountered it before trying to do headless testing of WebGL applications). For some reason via remote desktop, WebGL is being disabled, Chrome is probably detecting that you don't have the ability to do proper hardware acceleration through the Remote Desktop and shuts it off.

joshm-3b6b58:
Yikes. Is there a way to force it on? This might be inconvenient if you have to remote desktop into a system to access the webserver, but I suppose that should never happen?

joshm-3b6b58:
This did the trick. We might want to catch this error and bring up a dialog to point to this workaround.

> Step 1: Open Google Chrome
> Step 2: Type chrome://flags in the address bar
> Step 3: Press Ctrl + f and type ” Rendering list “, “Override software rendering list” should come up, >Now click on Enable and restart the browser.

adamcw:
I determined, that since the zoomRectangle will never be in screenshots anyway, there was no reason not to just render it as a div on top of the plot. PTAL.

client-js/app/elements/labrad-plot.ts
```diff
@@ -134,69 +155,256 @@ export class Plot extends polymer.Base {
...
+    this.graphUpdateRequired = true;
+  }
+
+
+  /**
+   * Redraw the plot and attach resize to the window resize event.
+   * Fires when the component is attached to the DOM.
+   */
+  attached(): void {
+    this.plotID = Math.random();
```
maffoo: Typescript compiler throws an error here.

adamcw: Line shouldn't have even been there, fixed.

client-js/app/elements/labrad-plot.ts
```diff
@@ -1,9 +1,25 @@
 import 'd3';
-
+import {THREE} from 'three';
```
maffoo: I get runtime errors from this. I think this needs to get the _default_ import, rather than an item named 'THREE' from inside the module. That is:

``` typescript
import THREE from 'three';
```

maffoo:
One comment about import (ts compiler doesn't complain, but it fails at runtime; we can investigate this separately). Otherwise LGTM.

PR #224, opened 2016-07-12
Fix error causing 1D plots not to render properly

adamcw:
PR #219 actually broke 1D plots because they do not have a zAxis defined. This fixes 1D plots in `master`.


maffoo:
Good catch. I had noticed 1D autoscaling was broken, but hadn't dug into why. LGTM.

Issue #220, opened 2016-06-30
Convert remaining javascript files to typescript


maffoo:
We still have a few polymer elements that include script tags containing plain javascript. Having this mix of languages can be confusing (see #211 for example). We should convert these to typescript.

adamcw:
This probably needs to be done sooner rather than later. I've added it to the v2.0 timeline to avoid code health being deprioritized.

adamcw:
@maffoo do you know off the top of your head the files that need to be converted? Or should I take an inventory?

maffoo:
They're actually .html files with script tags inside. Look for .html files in the elements directory without a matching .ts file. That should be all.

adamcw:
As far as I can tell we can delete:
- app/elements/dir-item.html
- app/elements/menu-button.html
- app/elements/expandable-input.html

But we need to convert:
- app/elements/selectable-table.html
- app/elements/labrad-setting.html
- app/elements/labrad-breadcrumbs.html

These are html files without ts files but don't have anything to convert:
- app/elements/elements.html
- app/elements/app-theme.html

PR #225, opened 2016-07-12
Convert app-link element to use typescript.

maffoo:
See #220


client-js/app/elements/app-link.ts
```diff
@@ -0,0 +1,30 @@
...
+  @property({type: Array})
+  path: string[];
+
+  isSpecialClick(e): boolean {
+    return (e.ctrlKey || e.metaKey || e.button !== MOUSE_MAIN_BUTTON);
+  }
+
+  @listen('click')
+  onClick(e) {
+    if (this.isSpecialClick(e)) return;
```
adamcw, line 16: Here you have a one line return without curly braces, on line 25 you include the curly braces.

client-js/app/elements/app-link.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    this.fire('app-link-click', {path: this.path});
```
adamcw, line 18: Is the styleguide here:

``` typescript
this.fire('app-link-click', {path: this.path});
```

or 

``` typescript
this.fire('app-link-click', { path: this.path });
```

adamcw:
Two small comments, else LGTM.

client-js/app/elements/app-link.ts
maffoo, line 16: LOL; I'm inclined to remove curlies below, but how do you feel about these one-line if statements?

client-js/app/elements/app-link.ts
maffoo, line 18: For js at least it's the former: https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting

client-js/app/elements/app-link.ts
adamcw, line 16: I personally prefer braces for consistency, and if you include braces, I prefer not to one-line them.

``` typescript
if (this.isSpecialClick(e)) {
  return;
}
```

Everyone is different on this though.

client-js/app/elements/app-link.ts
adamcw, line 18: Seems to be the same for Python too. This is fine then.

client-js/app/elements/app-link.ts
maffoo, line 16: The js style guide points to the c++ style guide which allows single line conditionals (https://google.github.io/styleguide/cppguide.html#Conditionals), so I went with that. I agree that everyone is different on this; we should set up automated linting/formatting and just use that :-)

PR #226, opened 2016-07-12
Use same typescript options for compiling app and tests.

maffoo:
Makes gulpfile more DRY and also fixes a warning about non-existent `experimentalAsyncFunctions` option.


adamcw:
LGTM. Nice.

PR #228, opened 2016-07-12
Make gulp `test` task run tests, and enable `npm test`.

maffoo:
`web-component-tester` was taking over the test task, but we're not using it at the moment so I removed it. If someone gets ambitious and decides to write tests for the custom elements, we can make the test task also run those tests.

Also hooked up `npm test` to run the gulp tests, since that is the npm convention, though not super important for us.


adamcw:
LGTM.

PR #234, opened 2016-07-15
Replace gulp-minify-html with gulp-htmlmin.

maffoo:
`gulp-minify-html` is deprecated according the npm page and they suggest using `gulp-htmlmin` instead. This was prompted by the fact that our gulp build was failing for me on a fresh npm install due to bad dependencies in `gulp-minify-html`.

See: https://www.npmjs.com/package/gulp-minify-html


adamcw:
LGTM.

Issue #233, opened 2016-07-15
1D Plot Live Limits


adamcw:
Are the limits on the 1D plots in live view really the intended output? At least on the sample data you basically just see a flash of random lines. This is unlike the 2D limits which scale to fit all the data in the frame.

Is this a long standing bug, or does the non-test use case actually find this useful?

maffoo:
No, this is not intended. The limits should work like for 2D plots, scaling to fit the full data set and adjusting as more data is received. I'm not sure when this broke.

PR #236, opened 2016-07-15
Fix limits for live 1D plots to show all the data

adamcw:
Fixes issue #233. 1D live plots were having the data bounds being reset to only show the last stream of data, rather than scaling to view the full plot like 2D plots.

Simply removed the limits reset occurring each time new 1D data points are plotted.


maffoo:
LGTM

PR #237, opened 2016-07-15
Changed typings exports to use es6 defaults for d3 and THREE.

adamcw:
In order to properly import external modules nicely, we need them to support ES6 export syntax. Updated the typings files for `d3` and `THREE` to use this export style.


maffoo:
LGTM

PR #238, opened 2016-07-15
Shutdown plot render loop and remove event listeners on detach.

maffoo:
This ensures that we don't do unnecessary work or prevent the widget from getting GC'd after it is detached.

@adamcw

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -1,7 +1,8 @@
 import 'd3';
 import THREE from 'three';
 import {viridisData} from '../scripts/colormaps';
-import * as datavault from "../scripts/datavault";
+import * as datavault from '../scripts/datavault';
+import {Lifetime} from '../scripts/lifetime';
 
 
 /**
@@ -89,6 +90,8 @@ export class Plot extends polymer.Base {
 
   @property({type: Array})
   deps: {label: string, legend: string, unit: string}[];
+
+  private lifetime = new Lifetime();
 
   private data: number[][] = []
   private lastData: number[] = null;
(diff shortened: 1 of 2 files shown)


adamcw:
If you want to do it like this, you will have to remove the `finishRender` line from `activities.ts` that I already set up to stop the render loop.

client-js/app/elements/labrad-plot.ts
```diff
@@ -274,7 +277,17 @@ export class Plot extends polymer.Base {
    */
   attached(): void {
     this.render();
-    window.addEventListener('resize', (e) => this.resizePlot());
+    this.lifetime.defer(() => { this.finishRender = true; });
+    const resizeListener = (e) => this.resizePlot();
+    window.addEventListener('resize', resizeListener);
+    this.lifetime.defer(() => {
```
adamcw, line 283: Can you not merge both of these `this.lifetime.defer` functions?

maffoo, line 283: Yes, could do that. The reason I prefer doing it this way is to follow the way go's defer statement is used: do operation A then defer the cleanup of A; next do operation B and defer the cleanup of B. That way each setup/cleanup pair is small and self-contained.

maffoo:
@adamcw, made `finishRender` private and removed the reference from `activities.ts`. PTAL

adamcw:
Detaching the event listeners and moving `finishRender` locally are both things I approve of, so LGTM. However, I'd prefer if we could replicate the supposed slow-down on master that inspired this PR before merging it.

maffoo:
I am also unable to reproduce it now. I'm willing to chalk it up to my having done something silly, like not having bower packages up to date. Sorry for the false alarm...

adamcw:
All good then. Merge away.

PR #239, opened 2016-07-20
Updated Grapher menu styling and function.

adamcw:
Updated the Grapher menu to behave like the updated registry menu in Pull #202.
- Icons are now clickable as well as the text, including the back arrow.
- Back arrow is now always a dark grey color instead of turning purple.


maffoo:
LGTM

PR #240, opened 2016-07-20
Fix a compiler error introduced in PR #216.

adamcw:
Fix a compiler error introduced in PR #216 due to no definition of connectionError.


maffoo:
Oh, I already pushed this to master with an over-the-shoulder review from you since we chatted about it :-)

adamcw:
I see that, I guess I'll close this.

Issue #242, opened 2016-07-25
Automatic zoom if only some dependent traces displayed


jwenner:
For a 1D dataset with multiple dependent traces, the original y-axis limits are kept even if I select to display only some of the dependent traces. It would be nice if the y-axis limits were recomputed for only those traces being displayed. This is particularly useful when the difference between values between traces is much greater than the variation within a trace.

Note: currently on version 1.1.0-0: 280f46b

adamcw:
If this is how it works in the Delphi grapher, this is a bug, otherwise it is a feature. Can you confirm?

joshm-3b6b58:
This is a bug. Axes were recomputed after traces were changed on version before WebGL was implemented. This was just something that was missed on review.

adamcw:
Ahh. The data generator doesn't really produce anything to test this functionality as it stands. I'll have to write a new case.

joshm-3b6b58:
True story.

adamcw:
The code that broke live streaming of 1D plots is the code that made the zoom level work when you change traces. Working on a way to make both work at the same time.

adamcw:
A simple fix. Need to reset the data limits when redrawing the scene. Old code reset whenever `plot1DData` was called, which meant that the zoom was reset on every incoming live packet, but also worked when traces were modified. 

Since live data now doesn't redraw the whole scene, the reset can be put into `redrawScene` and avoid the live streaming issue.

joshm-3b6b58:
So I recently had to get a new workstation, so my scalabrad-web stuff is kind of toast and I won't be able to get this to review for a day or two...

adamcw:
Seems @maffoo is on the case!

PR #243, opened 2016-07-25
Fix 1D plots to properly adjust zoom level when changing traces.

adamcw:
Fixes issue #242 where 1D plots do not re-zoom to properly frame the data when traces are toggled on and off.

Includes an update to `data_demo.py` to create sine waves with enough amplitude variance to test this feature in the future to avoid further regressions.

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -860,6 +860,11 @@ export class Plot extends polymer.Base {
     this.limits.yMin = isNaN(this.dataLimits.yMin) ? 0 : this.dataLimits.yMin;
     this.limits.yMax = isNaN(this.dataLimits.yMax + this.dy) ?
         0 : this.dataLimits.yMax + this.dy;
+
+    if (this.dataLimits.zMin === this.dataLimits.zMax) {
+      this.dataLimits.zMin -= 1;
+      this.dataLimits.zMax += 1;
+    }
   }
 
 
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-plot.ts
```diff
@@ -367,8 +367,18 @@ export class Plot extends polymer.Base {
 
     this.lastData = null;
 
+    this.dataLimits.xMin = 0;
+    this.dataLimits.xMax = 0;
+    this.dataLimits.yMin = 0;
+    this.dataLimits.yMax = 0;
```
maffoo: Should use `NaN` as the sentinel values when data limits need to be recomputed. Might be good to modify the test case to put some traces entirely above or below 0 to check this. Also, probably want to reset `zMin` and `zMax` to take care of the 2D case, and add a data_demo case for that too.

adamcw: Existing 2D_multi data test is sufficient to test the 2D case. It elucidated the fact the z-axis wasn't being adjusted correctly when all values were equal, so I fixed that also.

adamcw:
PTAL.

server/src/test/scripts/data_demo.py
```diff
@@ -36,7 +36,6 @@ def demo_1d_simple(dv):
         dv.add([x, y])
         time.sleep(0.002)
 
-
```
maffoo, line 39: style guide says: two blank lines between top-level definitions (https://google.github.io/styleguide/pyguide.html#Blank_Lines)

maffoo:
One minor comment about whitespace. Otherwise LGTM.

Issue #26, opened 2015-08-02
Application secret not so secret


DanielSank:
Should [this](https://github.com/labrad/scalabrad-web/blob/6d30720892289f4aa24f8aa224dd137e48cc11c2/server/src/main/resources/application.conf#L12) be in version control?

maffoo:
The way this is typically done is to deploy the app on a server with a different config file that overrides this secret. So I think of this as the dev mode secret and it's not a problem to have it in version control.

(Also, this is used by play for things like encrypting session cookies, which we aren't using at all; this value actually does not get used at all by the server here...)

DanielSank:
Does deploying without writing an override config work? If so that seems kind of dangerous.

DanielSank:
bump

adamcw:
@maffoo Can you respond to this so we can either action or close it?

maffoo:
`application.conf` is no longer needed since play framework is gone. We should delete it along with the `routes` file.

PR #245, opened 2016-07-25
Remove application.conf and routes files from server/.

adamcw:
Resolves issue #26. Since the play framework is no longer in use, the `application.conf` and `routes` files are no longer necessary.


maffoo:
LGTM

PR #246, opened 2016-07-26
Fix registry modal sizing when working with very long keys.

### LINKED ISSUE #244: Impossible to Edit Long Registry Keys
#156 redux. For a registry key which is longer than the current screen size, the registry key flows past the end of the key edit box. When I try to drag value to see values currently off-screen, instead the main registry page scrolls. Note that I had shrunk the Chrome window size to test this. On version 280f46b.

image

adamcw:
Fixes #244 where very long registry keys would scroll off the page due to the dialog not resizing properly.

This PR also fixes the unreported issue that if you pasted a long key into the new key or edit key windows, the modal window would not reposition, causing it to go off screen.

This bug was caused by an interplay of the various Polymer elements being used. The main culprit being that the `iron-resize` event that `paper-dialog` listens to so that it knows when to resize/reposition does not get fired by a `iron-autogrow-textarea` which is a lower level component that `paper-textarea` wraps.

This is resolved by first simplifying the component interaction stack by removing the use on `paper-dialog-scrollable` and moving the job of scrolling to the dialog itself. Secondly, the value update event on the textarea is hooked, and used to fire an `iron-resize` event that the dialog can catch, causing both the position and sizing to be set correctly.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -75,6 +75,9 @@
       color: black;
     }
 
+    paper-dialog {
+      overflow-y: auto;
+    }
     paper-dialog.size-position {
       position: fixed;
       top: 16p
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-registry.ts
```diff
@@ -270,7 +292,8 @@ export class LabradRegistry extends polymer.Base {
     newKeyElem.value = '';
     newValueElem.value = '';
     dialog.open();
-    window.setTimeout(() => newKeyElem.$.input.focus(), 0);
+
+    this.bindIronAutogrowTextAreaResizeEvents(dialog, newValueElem);
```
maffoo: This is getting called every time the dialog opens, which means we add a new 'iron-overlay-closed' listener to the dialog every time the dialog is opened. I would suggest instead doing initialization work in the attached callback of the registry element itself; you can add a listener to each textarea that looks for the value changes and fires a resize event, and there's no need to remove this listener since it will just go away when the textarea element is destroyed.

client-js/app/elements/labrad-registry.html
```diff
@@ -230,34 +233,30 @@
...
-          <paper-textarea id="newValueInput" label="Value"></paper-textarea>
-        </form>
-      </paper-dialog-scrollable>
-      <div class="buttons"> 
+      <h1>Create New Key</h1>
+      <form>
+        <paper-input id="newKeyInput" label="Key"></paper-input>
+        <paper-textarea id="newValueInput" label="Value"></paper-textarea>
+      </form>
+      <div class="buttons">
```
maffoo, line 241: The idea previously was to keep the buttons visible in the dialog, even if the text has to scroll. There may have been some bugs with that which need fixing, but I think that's nicer than having the buttons scroll offscreen when the text overflows.

adamcw:
@maffoo PTAL.

maffoo:
LGTM

PR #248, opened 2016-07-27
Added close buttons to the manager to kill connections.

### LINKED ISSUE #247: Expose "kill client" for manager browser
It occasionally happens that a server is running outside of a node.  In that case, an updated copy of the server cannot be started unless the old one gets shut down.  The manager interface to the browser should allow you to kill connections so that you don't have to hunt them down manually

See https://github.com/labrad/scalabrad/issues/62

adamcw:
Fix #247. Implements a close button for each connection on the Manager screen that will kill that connection. 

Allows for the killing of servers that have detached from nodes without having to hunt them down manually.

### DIFF
--- a/client-js/app/elements/labrad-manager.html
+++ b/client-js/app/elements/labrad-manager.html
@@ -55,10 +55,10 @@
             <td><span>{{item.id}}</span></td>
             <td>
               <template is="dom-if" if="{{item.server}}">
-                <span>server</span>
+                <span>Server</span>
               </template>
               <template is="dom-if" if="{{!item.server}}">
-                <span>client</span>
+                <span>Client</span>
               </template>
             </td>
             <td>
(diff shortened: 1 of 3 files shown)


client-js/app/elements/labrad-manager.ts
```diff
@@ -1,9 +1,14 @@
...
 @component('labrad-manager')
 export class LabradManager extends polymer.Base {
 
   @property({type: Array, notify: true})
   connections: Array<ConnectionInfo>;
 
+  mgr: ManagerApi;
+
+  closeConnection(event) {
+    this.mgr.connectionClose(event.currentTarget.connId);
```
maffoo: It would be good to ask for confirmation before actually closing the connection. Can just use `window.confirm` for simplicity. This is not an operation we will do often, and we should be careful to avoid killing connections accidentally.

maffoo:
Add confirmation before killing connections, then LGTM.

client-js/app/elements/labrad-manager.ts
adamcw: I did wonder that, I tried to make the actual buttons tiny to avoid it. 
`window.confirm` is a big gross, I'll do it properly with a dialog and resubmit.

maffoo: I like `window.confirm` a lot compared to messing around with polymer dialogs, though that probably says more about polymer dialogs...

adamcw: True, but I'd rather not need to open an issue to replace `window.confirm` with a nicer dialog that will never get cleared out.

adamcw:
@maffoo The dialog was actually a decently large addition, so PTAL.

maffoo:
LGTM

client-js/app/elements/labrad-manager.html
```diff
@@ -70,9 +77,26 @@
...
+              </span>
+            </td>
           </tr>
         </template>
       </tbody>
     </table>
+    <paper-dialog id="closeConnectionConfirmation" modal>
+      <h1>Close Connection</h1>
+      <p>Are you sure you want to close the connection to {{selectedConnection.name}}?</p>
+      <div class="buttons">
```
ejeffrey, line 101: Show both the connection name and ID here.

ejeffrey:
One minor note, then LGTM

client-js/app/elements/labrad-manager.html
adamcw, line 101: Good idea, done.

Issue #182, opened 2016-05-25
Grapher: Allow for basic fitting


jwenner:
The old Delphi grapher has support for fitting parabolas and exponentials. There are times when such manual fitting is useful. Note that it's possible that this might be implemented in a similar manner to #181, by doing the fitting and then using #181 to display the fit.

joshm-3b6b58:
This is a good idea. The question is how do we want to support it? Do you want to add a list of fit functions and have them accessible by buttons somewhere a-la old grapher? Do we need to support any customizing fit functions?

btchiaro:
I think it would be nice to have a button for the standard ones like
exponentials, but it would be nice to support arbitrary fit functions with
a text box or something.

jwenner:
Perhaps it might be good to have buttons for the standard fit functions which then populate a text box used to define the fit function. That way, the parameter names are clearly defined, and it's possible to verify the fit function used (and modify it if necessary).

joshm-3b6b58:
Can you enumerate the standard fit functions you'd like?

jwenner:
For the standard functions, I would think at least exponential and parabola. This would at least maintain feature parity with the old Grapher. Beyond this, I could see linear, Gaussian, or sine, but I don't know how much we would want these (at least, assuming we can enter mathematical functions in a text box).

jwenner:
By the way, I'd say this is only of medium priority. In particular, I'd say this is probably on the low end of achieving feature compatibility with the old grapher.

adamcw:
I have begun work on this in a branch, at the moment the goal is just feature parity so I am porting Parabola and Exponential fits from the Delphi grapher only.

https://github.com/labrad/scalabrad-web/tree/182-basic-fit-functions

I looked into more generic fitting methods, and implementations in JavaScript are sparse, and ones with proper licensing were non-existent. They are sufficiently complex that I left fitting to a generic function as a task for the future based on demand.

The branch so far ports the basic fitting functionality, but has no UI to turn them on/off, nor does it display the coefficients of the fitted function anywhere. At the moment I am rendering fits in the same color as the line it is fitting, but twice the thickness. I figure this is appropriate given they will be toggled on and off by the user, so you will want them to be clearly seen, but also easily identifiable in terms of which trace they apply to.

PR #249, opened 2016-07-28
Add support for fitting Parabolas and Exponentials to plots in Grapher.

adamcw:
Implements #182. Added a dialog to Grapher for 1D plots presenting the user with checkboxes to enable/disable fitting Parabolas/Exponentials to the visible traces.

Parabolas and exponential functions were implemented as these were the fits available in the Delphi grapher before Scalabrad-web was developed.


adamcw:
I have opened this PR to get feedback. I am uncertain as to if/how the function coefficients would be best displayed to the user. 

Is this information that should be part of the graph, or could it be shown within the fit dialog?

Also, I figured the `services:fitness-center` icon from the `iron-icon` set was the best we could do, but if someone has a better suggestion I am open to it. I didn't spend too much time laboriously checking every icon for suitability.

maffoo:
A simple way to display the coefficients would be to add them in the line that shows the cursor position, so it would say something like:

```
Cursor Position: (123, 456)    Polynomial Fix: y = 0.123 x^2 + 4.5 x - 67.8
```

I realize that doesn't work very well if you want to fit multiple traces, but the delphi grapher doesn't actually support that, so if we're just going for parity we only need to fit the first (or only) displayed trace.

One important aspect of the delphi grapher's fitting that is missing here is that it allows the user to use a drag box to define a region of data to consider for the fit, rather than taking the whole data set. This is important for the typical use cases because often we want a parabolic fit to find the extremum of one peak or dip in a more complicated dataset, for example.

adamcw:
Closing as development was moved to a separate stacked pull request due to the significant changes this feature required.

PR #250, opened 2016-07-29
Show exceptions on Nodes when they occur during start, stop or restart of a server.

### LINKED ISSUE #169: Node: display error message when server fails to start
Currently, errors that prevent the node from starting a server just get logged to the dev console. These error messages should be displayed to the user.

adamcw:
Fixes #169. Shows the user that an error has occurred if an exception is thrown during the starting, stopping or resetting of a server from the Node panel. Also show the whole exception that was raised in a collapsible panel.

Previously, when an exception occurred while starting, stopping or resetting a server, it was logged to the console, but not displayed to the user.

Also adjusts the Node UI to fill the screen and fixes the "jump" that would occur when the loading spinner was active on a line of controls.

### DIFF
--- a/client-js/app/elements/labrad-nodes.ts
+++ b/client-js/app/elements/labrad-nodes.ts
@@ -3,6 +3,21 @@ import {ManagerApi, ServerConnectMessage, ServerDisconnectMessage} from '../scri
 import {NodeApi, NodeStatus, ServerStatus, ServerStatusMessage} from '../scripts/node';
 import {Places} from '../scripts/places';
 
+
+@component('labrad-exception-handler')
+export class LabradExceptionHandler extends polymer.Base {
+  @property({type: String, value: ''})
+  error: string;
+
+  @property({type: String, value: ''})
+  exception: string;
+
+  toggleException() {
+    const collapse = <any>(this.querySelector('#exceptionCollapse'));
+    collapse.toggle();
+  }
+}
+
 @component('labrad-instance-controller')
 export class LabradInstanceController extends polymer.Base {
   @property()
@@ -26,6 +41,9 @@ export class LabradInstanceController extends polymer.Base {
   @property({type: Boolean, value: false})
   active: boolean;
 
+  @property({type: Object, notify: true})
+  server: ServerInfo;
+
   @property()
   api: NodeApi;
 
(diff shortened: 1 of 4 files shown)


client-js/app/elements/labrad-nodes.ts
```diff
@@ -3,6 +3,21 @@ import {ManagerApi, ServerConnectMessage, ServerDisconnectMessage} from '../scri
...
+@component('labrad-exception-handler')
+export class LabradExceptionHandler extends polymer.Base {
+  @property({type: String, value: ''})
+  error: string;
+
+  @property({type: String, value: ''})
+  exception: string;
+
+  toggleException() {
+    const collapse = <any>(this.querySelector('#exceptionCollapse'));
```
maffoo: I think you can just type the variable instead:

``` typescript
const collapse: any = this.querySelector('#exceptionCollapse');
```

client-js/app/elements/labrad-nodes.ts
```diff
@@ -46,24 +64,52 @@ export class LabradInstanceController extends polymer.Base {
...
+  async doStart() {
     console.info(`Start: server='${this.name}', node='${this.node}'`);
-    this.api.startServer({node: this.node, server: this.name});
+    try {
+      await this.api.startServer({node: this.node, server: this.name});
+      this.set("server.errorString", "");
+      this.set("server.errorException", "");
+    } catch (e) {
+      console.error("Exception while starting server: ", e);
+      this.set("server.errorString", "An error occurred while starting the server.");
```
maffoo: Would be good to include the server name and node here in the message.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -46,24 +64,52 @@ export class LabradInstanceController extends polymer.Base {
...
+  async doStop() {
     console.info(`Stop: server='${this.name}', node='${this.node}'`);
-    this.api.stopServer({node: this.node, server: this.instanceName});
+    try {
+      await this.api.stopServer({node: this.node, server: this.instanceName});
+      this.set("server.errorString", "");
+      this.set("server.errorException", "");
+    } catch (e) {
+      console.error("Exception while stopping server: ", e);
+      this.set("server.errorString", "An error occurred while starting the server.");
```
maffoo: "...while stopping the server."

Also, as above, please include server name and node name.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -46,24 +64,52 @@ export class LabradInstanceController extends polymer.Base {
...
+  async doRestart() {
     console.info(`Restart: server='${this.name}', node='${this.node}'`);
-    this.api.restartServer({node: this.node, server: this.instanceName});
+    try {
+      await this.api.restartServer({node: this.node, server: this.instanceName});
+      this.set("server.errorString", "");
+      this.set("server.errorException", "");
+    } catch (e) {
+      console.error("Exception while restarting server: ", e);
+      this.set("server.errorString", "An error occurred while starting the server.");
```
maffoo: "... while restarting the server." + server name and node name, as above.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -74,7 +120,7 @@ export class LabradInstanceController extends polymer.Base {
           case 'STOPPED': this.updateButtonState(this.status); break;
           case 'STARTING': this.updateButtonState('RUNNING_ELSEWHERE'); break;
           case 'STARTED': this.updateButtonState('RUNNING_ELSEWHERE'); break;
-          case 'STOPPPING': this.updateButtonState('RUNNING_ELSEWHERE'); break;
+          case 'STOPPING': this.updateButtonState('RUNNING_ELSEWHERE'); break;
```
maffoo, line 131: LOL; good catch :-)

client-js/app/elements/labrad-nodes.ts
```diff
@@ -127,6 +173,8 @@ export class LabradNodeController extends polymer.Base {
     this.active = true;
     try {
       await this.api.refreshNode(this.name);
+    } catch (e) {
+      console.error("Exception while refreshing server: ", e);
```
maffoo: Maybe say "while refreshing node" instead of server. Also, include node name.

maffoo:
Nice! I like the way this looks to show the errors inline in the list; especially since you might get errors in lots of servers at once, so dialogs or popups would be annoying. Right now there's no way to dismiss an error unless you try to start the server again and it works. It would be nice to be able to do that; for example, suppose you click on the wrong button and try to start a known bad server; you don't actually want to start it, and it's never going to work so the error message will never go away (unless you reload the page).

Also, a couple of styling nits:
- the error box is hardcoded with colspan=3, which makes it not expand to full width if there is more than one node server. I'm not sure if there's a way to fix this other than binding a value to pass in the number of columns. Not a big deal if this is a pain.
- Having a top border of the error box sort of visually separates it from the instance controller above, IMO, and then the box fades in to the instance controller below. Again, not a big deal, but it would be nice to make it clear which server this is attached to. Maybe make the entire error box red, but put some background padding below it before the next instance controller. Up to you. Also, including the server name in the error message would make it clear which server this applies to.

adamcw:
1) I'll add a dismiss button.

2) I've set colspan to be 999, it's a bit of a hack, but having nodes as columns probably isn't the greatest UI choice to begin with, so it works until we find time to clean up the UI overall.

3) The purpose of the red line separator was to be in line with the Material Design recommendations for visually indicating errors:

https://material.google.com/patterns/errors.html#errors-user-input-errors

Making the whole thing red just seemed in-your-face. I think adding the server name/node name plus it being visually below the server is sufficient.

adamcw:
@maffoo PTAL

maffoo:
On issues: after dismissing, the error block never reappears on subsequent errors. Otherwise, this LGTM

adamcw:
Weird, it shows up again if the server successfully starts before it fails again (my `dying_test_server.py` randomly starts or fails rather than always failing). Will fix.

PR #251, opened 2016-08-01
Convert Remaining JavaScript Components to TypeScript

### LINKED ISSUE #220: Convert remaining javascript files to typescript

adamcw:
Fixes #220 by converting `labrad-setting.html`, `labrad-breadcrumbs.html` and `selectable-table.html` into components specified in separate TypeScript files rather than written in JavaScript within the HTML.


maffoo:
LGTM

PR #252, opened 2016-08-02
Allow Scalabrad-web to load in Firefox

### LINKED ISSUE #93: Not usable with client browsers of Firefox, Internet Explorer
When I try to access the scalabrad-web page using non-Chrome browsers (tried Firefox, Internet Explorer) with version -M1 from bintray, I get just a blank page. No login screen, nothing.

adamcw:
Fixes #93, somewhat.

Firefox blacklists my GPU/Graphics Drivers on my dev machine, so WebGL doesn't work, breaking the Grapher, this isn't an issue with the build though.

The remaining functionality appears to work, though there are some styling issues in places.

### DIFF
--- a/client-js/app/index.html
+++ b/client-js/app/index.html
@@ -46,6 +46,7 @@
 </head>
 
 <body unresolved class="fullbleed layout vertical">
+  <span id="browser-sync-binding"></span>
   <script src="scripts/bundle.js"></script>
 </body>
 </html>
--- a/client-js/app/index.html
+++ b/client-js/app/index.html
@@ -43,10 +43,11 @@
        websocket api connections should be made. Comments are stripped
        when we built the dist version, so this has no effect in prod. -->
   <!-- DEV_MODE_CONFIG -->
+
+  <link rel="import" href="/bundle.html">
 </head>
 
 <body unresolved class="fullbleed layout vertical">
   <span id="browser-sync-binding"></span>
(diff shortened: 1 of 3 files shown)


client-js/app/index.html
```diff
@@ -43,9 +43,11 @@
        websocket api connections should be made. Comments are stripped
        when we built the dist version, so this has no effect in prod. -->
   <!-- DEV_MODE_CONFIG -->
+
+  <link rel="import" href="/bundle.html">
 </head>
 
 <body unresolved class="fullbleed layout vertical">
-  <script src="scripts/bundle.js"></script>
+  <span id="browser-sync-binding"></span>
```
maffoo, line 51: Can this be a comment instead of an actual html element? We only need browser sync in dev mode, and comments get stripped out when we build for production mode.

maffoo:
Thanks for figuring this out! Would be nice to use an html comment for browser sync so it doesn't leak into prod mode. Otherwise LGTM.

client-js/app/index.html
adamcw, line 51: Probably, I'll try it out.

adamcw:
This doesn't actually seem to work for me anymore, it seems to replace the <span> (or a comment) just fine, but it still replaces the <body> tag in the comment breaking the app.

The Polyfill makes Firefox slow and unusable anyway, so I'm just going to close this PR until I get a chance to do Firefox support properly.

PR #253, opened 2016-08-02
Refresh connection info periodically to keep counts updated.

maffoo:
For now, just polls every 2 seconds and right after triggering a connection close. Would be nice to just stream the data, but that requires changes to the manager itself, so not worrying about that now.

### DIFF
--- a/client-js/app/elements/labrad-manager.ts
+++ b/client-js/app/elements/labrad-manager.ts
@@ -1,3 +1,4 @@
+import {Lifetime} from '../scripts/lifetime';
 import {ConnectionInfo, ManagerApi} from '../scripts/manager';
 
 @component('labrad-manager')
@@ -10,6 +11,17 @@ export class LabradManager extends polymer.Base {
   selectedConnection: {connId: number, name: string};
 
   mgr: ManagerApi;
+
+  private lifetime: Lifetime = new Lifetime();
+
+  attached() {
+    const interval = setInterval(() => this.updateConnections(), 2000);
+    this.lifetime.defer(() => clearInterval(interval));
+  }
+
+  detached() {
+    this.lifetime.close();
+  }
 
   closeConnection(event) {
     this.selectedConnection = {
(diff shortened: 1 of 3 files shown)


adamcw:
This loses the url information that is initially set by `activities.ts`

``` TypeScript
    var conns = await this.api.connections();
    var connsWithUrl = conns.map((c) => {
      var x = <any> c;
      if (c.server) {
        x['url'] = this.places.serverUrl(c.name);
      }
      return x;
    });
    var elem = <LabradManager> LabradManager.create();
    elem.connections = connsWithUrl;
```

client-js/app/elements/labrad-manager.ts
```diff
@@ -28,14 +40,14 @@ export class LabradManager extends polymer.Base {
...
-    const numConnections = this.connections.length;
-    for (let i = 0; i < numConnections; ++i) {
-      if (this.connections[i].id == id) {
-        this.splice('connections', i, 1);
-        break;
-      }
+  private async updateConnections(): Promise<void> {
+    const connections = await this.mgr.connections();
+    this.splice('connections', 0, this.connections.length);
+    for (let c of connections) {
```
adamcw, line 49: You can do this as:

``` TypeScript
this.splice('connections', 0, this.connections.length, ...connections);
```

maffoo, line 49: Nice! I think there are a few other places we splice that could use this trick too.

client-js/app/scripts/activities.ts
```diff
@@ -370,17 +370,16 @@ export class ManagerActivity implements Activity {
               private api: manager.ManagerApi) {}
 
   async start(): Promise<ActivityState> {
-    var conns = await this.api.connections();
-    var connsWithUrl = conns.map((c) => {
-      var x = <any> c;
+    var connections = await this.api.connections();
```
adamcw, line 373: Would it be nicer to deduplicate this code by just having `updateConnections` called on attach?

maffoo, line 373: Made some changes to dedupe this. The connection info is still loaded by the activity, but goes through a `setConnections` method instead of just setting the property so that the urls get updated in one place. That way the view doesn't render empty before the first time connection info is loaded.

adamcw:
LGTM :)

Issue #143, opened 2016-01-26
Sub-list of servers for the node browser


thchwhite:
There should be a way to create a custom sub-list of servers so one doesn't have to look through every server to find the ones you want.  This is especially relevant for the ADR's where most of the servers are unused.

maffoo:
One thing you should know about is the node now lets you set a list of servers that will be autostarted. The web UI doesn't currently support this, but if you just use the python client you will see a bunch of `autostart_*` settings that you can use to set up the list, and then just call `node.autostart()` to run them all. This will avoid having to go through the web interface at all.

Of course, having the web UI support this in some way will be nice. Basically, this you could add and remove servers from the autostart list, and start all of them with one button.

joshm-3b6b58:
Also, adding the filter/search dialog to the node browser would help this in the meantime.

joshm-3b6b58:
@thchwhite did you manage to find a workaround for this? Is this still an issue?

joshm-3b6b58:
@thchwhite is this still a problem?

thchwhite:
Yeah.  I thought I found a solution using symlinks, but it was super unstable.  I like the idea of the auto-start list, but often you need to restart just one server several times, and its still a pain to find.  Ideally you could have an option to only display the auto-start list.  Like how you can click a button on the old grapher to only display stared data sets.

jwenner:
Should this be moved to milestone v2.1, as I believe this goes beyond feature-equivalence with the old Delphi+Java node webpage system?

adamcw:
I'm personally inclined to resolve all the `PE-fixit` milestone items first. For a few reasons:
1. None of the remaining items are that difficult.
2. There are only two of them without `pr-pending` labels
3. I don't like the slippery slope of meeting milestone targets by simply reclassifying them.

On that note though, thanks for all your efforts in helping triage and classify these issues, it's a big help!

adamcw:
How do we envisage this feature working? I can implement a starring system, like the Grapher, but that's going to be Scalabrad-web wide. If a server is important to one person, is it important to everyone, or are people going to each want their own subset?

If we need multiple subsets for various use cases, then we probably want something more like a tag system, where you can then filter by tag.

Search/filter was mentioned. Would that just be an interim solution, or might that actually solve this problem more elegantly than a bunch of tags?

@thchwhite @maffoo @joshmutus

maffoo:
I think we should use the node autostart list for this.
[...]

PR #254, opened 2016-08-04
[#143, Part 1] Add buttons to assign servers to a node autostart list, and autostart all servers on a node.

adamcw:
Partial fix for #143. Added buttons to each server in the Node manager allowing the adding/removing of each server from the respective node's autostart list. Additionally added a button at the top of each node that will start all nodes on the autostart list.

### DIFF
--- a/client-js/app/elements/labrad-nodes.ts
+++ b/client-js/app/elements/labrad-nodes.ts
@@ -121,7 +121,6 @@ export class LabradInstanceController extends polymer.Base {
   }
 
 
-  @listen('autostart.click')
   async doAutostart() {
     console.info(`Autostart: server='${this.name}', node='${this.node}'`);
     try {
@@ -306,7 +305,6 @@ export class LabradNodes extends polymer.Base {
     } else {
       this.splice('info', idx, 1, msg);
     }
-    this.kick++;
   }
 
   onServerDisconnected(msg: ServerDisconnectMessage): void {
--- a/client-js/app/elements/labrad-nodes.ts
+++ b/client-js/app/elements/labrad-nodes.ts
@@ -49,6 +49,12 @@ export class LabradInstanceController extends polymer.Base {
   @property({type: Object, notify: true})
   server: ServerInfo;
 
+  @property({type: Boolean})
+  autostart: boolean;
+
+  @property({type: String})
+  autostartClass: string;
+
   @property()
   api: NodeApi;
 
@@ -183,6 +189,18 @@ export class LabradNodeController extends polymer.Base {
       await this.api.refreshNode(this.name);
     } catch (e) {
       console.error(`Exception while refreshing node: ${this.name}`, e);
+    } finally {
+      this.active = false;
+    }
+  }
+
+  @listen('autostart.click')
+  async onAutostart() {
+    this.active = true;
+    try {
+      await this.api.autostartNode(this.name);
+    } catch (e) {
+      console.error(`Exception while autostarting node: ${this.name}`, e);
     } finally {
       this.active = false;
     }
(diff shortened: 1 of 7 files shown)


client-js/app/elements/labrad-nodes.ts
```diff
@@ -118,6 +121,27 @@ export class LabradInstanceController extends polymer.Base {
   }
 
 
+  async doAutostart() {
```
maffoo: maybe call this `toggleAutostart`, since it just modifies the list, doesn't actually trigger the anything to be autostarted.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -118,6 +121,27 @@ export class LabradInstanceController extends polymer.Base {
...
+        await this.api.autostartAdd({node: this.node, server: this.name});
+        this.set('autostart', true);
+      }
+      this.set('server.errorString', '');
+      this.set('server.errorException', '');
+    } catch (e) {
+      const message = `${this.instanceName} (${this.node}): An error occured while starting the server`;
+      console.error(message, e);
+      this.set('server.errorString', message);
+      this.set('server.errorException', e.message);
```
maffoo, line 147: This inline error functionality is really nice!

server/src/main/scala/org/labrad/browser/NodeApi.scala
```diff
@@ -50,6 +50,27 @@ class NodeApi(cxn: LabradConnection)(implicit ec: ExecutionContext) extends Logg
...
+  }
+
+  def autostartList(node: String): Future[Seq[String]] = {
+    async {
+      val list = await { cxn.to(node).call("autostart_list") }
+      list.get[Seq[String]]
+    }
+  }
+
+  def autostartAdd(node: String, server: String): Unit = {
```
maffoo: It would be better to make the return type `Future[Unit]` so that we don't reply to the client until the call to the node is complete, and we send an error back to the client if the call to the node fails.

adamcw:
@maffoo PTAL.

server/src/main/scala/org/labrad/browser/NodeApi.scala
```diff
@@ -63,12 +63,16 @@ class NodeApi(cxn: LabradConnection)(implicit ec: ExecutionContext) extends Logg
     }
   }
 
-  def autostartAdd(node: String, server: String): Unit = {
-    cxn.to(node).call("autostart_add", Str(server))
+  def autostartAdd(node: String, server: String): Future[Unit] = {
+    async {
+      cxn.to(node).call("autostart_add", Str(server))
```
maffoo: Actually just saying async here doesn't help. The issue is that `.call` already returns a `Future`, so you need to await it inside the async block, or just return the Future directly. I guess it's actually a `Future[Data]`, and we can't send arbitrary labrad data over jsonrpc (alas), so would have to drop the resulting value. Either one of these would do the trick:

``` scala
def autostartAdd(...): Future[Unit] = {
  cxn.to(node).call(...).map { _ => () } // discard the Data
}

def autostartAdd(...): Future[Unit] = {
  async {
    await { cxn.to(node).call(...) }
  }
}
```

maffoo:
One small comment, then LGTM.

maffoo:
One other minor comment: I like the star buttons for adding to/removing from the autostart list, but for the button to actually trigger the autostart operation, it would be nice to have a different icon that is more "action-y". Maybe a play button?

adamcw:
My concern is that a play button isn't very autostart specific, it could easily mean "start node" or "start all servers".

maffoo:
What about [fast forward](https://design.google.com/icons/#ic_fast_forward)? It's two arrows, sort of conveys starting more than one thing. Also, the point of autostart is to be faster than starting each server individually.

maffoo:
Or [playlist play](https://design.google.com/icons/#ic_playlist_play)

adamcw:
Let's do `playlist-play`.

maffoo:
While we're on icons, what do you think about using an icon with [star outline](https://design.google.com/icons/#ic_star) instead of the circular outline with a star in the middle? I think that might look less visually cluttered. Not a big deal though.

adamcw:
I am using this star because it is the one used in the Grapher. If we want to change the stars, I think we do them all at once, and in a separate PR.

maffoo:
Ok. LGTM with the playlist-play icon.

PR #255, opened 2016-08-04
Allow logging in with OAuth in the web interface

maffoo:


client-js/app/scripts/app.ts
```diff
@@ -364,28 +413,39 @@ window.addEventListener('WebComponentsReady', () => {
...
+          await mgr.login({
+            username: (credential as Password).username,
+            password: (credential as Password).password,
+            manager: manager
+          });
+          break;
+      }
+    } catch (error) {
+      if (isPasswordError(error)) {
+        // if we had credentials, clear them out
```
adamcw: Can you fix this comment to be in line with style guide? Capital and full stop.

client-js/app/scripts/app.ts
```diff
@@ -364,28 +413,39 @@ window.addEventListener('WebComponentsReady', () => {
...
-        });
-      } catch (error) {
-        if (isPasswordError(error)) {
-          // if we had credentials, clear them out
-          storage.removeItem(key);
-        }
-        throw error;
+    }
+    try {
+      // TODO: in TS2.0 compiler will understand .kind; can then remove casts
```
adamcw: Full stop needed. Also, it's recommended that you do "TODO (username)" so that future people know who to contact regarding the TODO, it's not considered signing up to do the change.

client-js/app/scripts/credentials.ts
```diff
@@ -0,0 +1,52 @@
...
+export type Credential = Password | OAuthToken;
+
+/**
+ * Load credential from the given Storage object (session or local).
+ *
+ * If valid a credential is not found, return null.
+ */
+export function loadCredential(manager: string, storage: Storage): Credential {
+  var key = 'labrad-credentials';
+  if (manager) {
```
adamcw: Is manager supposed to be optional? If so, it should be last in the list.

client-js/app/scripts/credentials.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+/**
+ * Save credential to the given Storage object.
+ */
+export function saveCredential(manager: string, storage: Storage,
+                               credential: Credential) {
+  var key = 'labrad-credentials';
+  if (manager) {
```
adamcw, line 55: See above.

client-js/app/scripts/manager.ts
```diff
@@ -34,13 +34,21 @@ export interface ServerInfo {
   settings: Array<SettingInfo>;
 }
 
+export interface OAuthInfo {
```
adamcw, line 37: Would probably be nicer if this was:

``` TypeScript
export interface OAuthClientInfo {
  id: string;
  secret: string;
}
```

client-js/app/scripts/manager.ts
```diff
@@ -53,7 +61,7 @@ export interface ManagerApi {
   serverDisconnected: Observable<ServerDisconnectMessage>;
 }
 
-export class ManagerServiceJsonRpc extends rpc.RpcService implements ManagerApi {
+export class ManagerApiImpl extends rpc.RpcService implements ManagerApi {
```
adamcw: Seems weird to me that the interface gets a nice name, and then the actual class has a "Impl" suffix.

maffoo: That's because code that uses this is written in terms of the interface, so it gets the nice name. I realize that I named this inconsistently from other classes which use `Service` as a suffix instead of `Impl`; I like `Impl` better but didn't want to change code to match that is otherwise unrelated to this PR.

client-js/app/scripts/credentials.ts
maffoo: `manager` is not optional, but can be the empty string.

client-js/app/scripts/manager.ts
maffoo, line 37: The names `client_id` and `client_secret` are used throughout the OAuth apis, which is why I like using the full names here (appropriately camel-cased). Can change if you feel strongly.

client-js/app/scripts/app.ts
maffoo: Fixed.

client-js/app/scripts/app.ts
maffoo: Fixed.

client-js/app/scripts/app.ts
```diff
@@ -283,74 +277,129 @@ window.addEventListener('WebComponentsReady', () => {
       mkDvRoutes(i);
     }
   }
+
+  // create route for OAuth login callback
```
adamcw: Fix comment.

client-js/app/scripts/app.ts
```diff
@@ -283,74 +277,129 @@ window.addEventListener('WebComponentsReady', () => {
...
+  page(prefix + '/oauth2callback', (ctx, next) => {
+    finishOAuthLogin(ctx.querystring);
+  });
+
+  // create the rest of the application routes
```
adamcw: Fix comment.

client-js/app/scripts/app.ts
```diff
@@ -283,74 +277,129 @@ window.addEventListener('WebComponentsReady', () => {
...
 
   /**
    * Launch a dialog box to let the user log in to labrad.
    */
-  function loginWithDialog(mgr: manager.ManagerApi, manager: string): Promise<void> {
+  async function loginWithDialog(mgr: ManagerApi, manager: string): Promise<void> {
     var {obligation, promise} = obligate<void>();
     async function doLogin() {
-      var username = '',
+      var username = app.$.usernameInput.value,
```
adamcw: Why are we writing new code that uses `var` instead of `let` or `const`? If the line is already changing, probably best to just clean them up.

server/src/main/scala/org/labrad/browser/LabradConnection.scala
```diff
@@ -108,7 +185,12 @@ class LabradConnection(
 
         if (live) {
           // reconnect after some delay
```
adamcw: Fix comment.

client-js/app/scripts/app.ts
```diff
@@ -283,74 +277,129 @@ window.addEventListener('WebComponentsReady', () => {
...
-    var key = 'labrad-credentials';
-    if (manager) {
-      key += '.' + manager;
+  async function finishOAuthLogin(queryString: string) {
+    const params = decodeQueryString(queryString);
+    const state = JSON.parse(params['state']);
+    console.log(params);
+    console.log(state);
+
+    // now, exchange the access token for an ID token
```
adamcw: Fix comment.

client-js/app/scripts/credentials.ts
```diff
@@ -0,0 +1,52 @@
...
+    key += '.' + manager;
```
adamcw: Given we seem to be avoiding string concatenation, `key += .${manager}` be more appropriate, or even `key = ${key}.${manager}`.

client-js/app/scripts/manager.ts
adamcw: It might be better to avoid fragmenting standards already in place until we are ready to actually replace them all. 

If the timeline for a PR fixing all of something is more than 1-2 days away, I think it should follow the existing naming convention then replace it when the new PR is finally ready.

client-js/app/scripts/app.ts
```diff
@@ -283,74 +277,129 @@ window.addEventListener('WebComponentsReady', () => {
...
-   * Save credentials to the given Storage object, to be loaded by loadCreds.
-   */
-  function saveCreds(manager: string, storage: Storage, creds: Creds) {
-    var key = 'labrad-credentials';
-    if (manager) {
-      key += '.' + manager;
+  async function finishOAuthLogin(queryString: string) {
+    const params = decodeQueryString(queryString);
+    const state = JSON.parse(params['state']);
+    console.log(params);
```
adamcw: Remove `console.log`

client-js/app/scripts/manager.ts
adamcw, line 37: I'm ok with this then.

client-js/app/scripts/credentials.ts
adamcw: I guess the compiler should catch anything non-string going into this, so a more formal `if` would be redundant.

client-js/app/scripts/app.ts
maffoo: Fixed.

client-js/app/scripts/app.ts
maffoo: Fixed.

client-js/app/scripts/app.ts
maffoo: Fixed.

client-js/app/scripts/app.ts
maffoo: Fixed.

client-js/app/scripts/app.ts
maffoo: Fixed.

client-js/app/scripts/credentials.ts
maffoo: Fixed.

client-js/app/scripts/manager.ts
maffoo: Changed the name back. Will file an issue to clean up these names.

client-js/app/scripts/app.ts
```diff
@@ -278,36 +278,36 @@ window.addEventListener('WebComponentsReady', () => {
...
           manager: manager
         });
-        var credential: Password = {
+        const credential: Password = {
           kind: 'username+password',
           username: username,
           password: password
         };
-        var storage = rememberPassword ? window.localStorage
+        const storage = rememberPassword ? window.localStorage
```
adamcw, line 310: Minor point.. the ternary isn't aligned properly anymore.

client-js/app/scripts/credentials.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Load credential from the given Storage object (session or local).
+ *
+ * If valid a credential is not found, return null.
+ */
+export function loadCredential(manager: string, storage: Storage): Credential {
+  var key = 'labrad-credentials';
+  if (manager) {
+    key += '.' + manager;
```
adamcw: Sorry, forgot to mention this concatenation thing happens twice. It's also here.

client-js/app/scripts/credentials.ts
adamcw: Oh, it happens three times, you actually didn't fix this one.

client-js/typings/whatwg-fetch/whatwg-fetch.d.ts
```diff
@@ -0,0 +1,98 @@
...
+declare interface HeadersMap {
+	[index: string]: string;
+}
+
+declare class Headers {
+	constructor(headers?:Headers|HeadersMap)
+	append(name: string, value: string): void;
+	delete(name: string):void;
+	get(name: string): string;
+	getAll(name: string): Array<string>;
```
adamcw, line 53: `string[]`

client-js/typings/whatwg-fetch/whatwg-fetch.d.ts
```diff
@@ -0,0 +1,98 @@
...
+
+type ResponseType = "basic" | "cors" | "default" | "error" | "opaque" | "opaqueredirect";
+
+interface ResponseInit {
+	status: number;
+	statusText?: string;
+	headers?: HeaderInit;
+}
+
+declare type HeaderInit = Headers|Array<string>;
```
adamcw, line 90: `string[]`

client-js/app/elements/labrad-app.ts
```diff
@@ -26,6 +26,12 @@ export class LabradApp extends polymer.Base {
   @property({type: Array, value: []})
```
adamcw: Make this `value: () => []`

client-js/app/elements/labrad-app.ts
```diff
@@ -26,6 +26,12 @@ export class LabradApp extends polymer.Base {
...
   breadcrumbExtras: Array<{name: string; isLink: boolean; url?: string}>;
```
adamcw: Make this a type and change `Array` to `[]`?

client-js/app/scripts/app.ts
```diff
@@ -283,74 +277,127 @@ window.addEventListener('WebComponentsReady', () => {
...
+        'state': JSON.stringify({
+          path: location.pathname + location.search + location.hash,
+          manager: manager,
+          client_id: clientId,
+          client_secret: clientSecret,
+          redirect_uri: redirectUri,
+          remember_me: rememberMe
+        })
+      };
+      const baseUrl = 'https://accounts.google.com/o/oauth2/v2/auth'
```
adamcw, line 360: This says `v2` but line 381 says `v4` is this really correct?

adamcw:
So, mostly minor comments, some not directly related to this PR, but they'll probably get forgotten if done later.

Big issue though, I checked out the branch, and upon clicking "Login with Google" nothing happens..

client-js/app/scripts/credentials.ts
maffoo: LOL; oops. Moved the storage key generation into one spot in `credentials.ts` and fixed it there.

client-js/typings/whatwg-fetch/whatwg-fetch.d.ts
maffoo, line 53: This is straight from the DefinitelyTyped repo, so I'm going to leave it as is (this one didn't require any changes to work with es6 modules since it just polyfills a global function).

client-js/app/scripts/app.ts
maffoo, line 360: This is correct as written, there's no `v4` on the `accounts.google.com/o/oauth2` endpoint. Note that the line below for actually renewing the access code to get an id token is `www.googleapis.com/oauth2`. I guess these are just versioned separately. The documentation of the web oauth flow is [here](https://developers.google.com/identity/protocols/OAuth2UserAgent), although we're doing things slightly differently because we have `response_type=code` so that we can exchange the code for an id token. That is more typical of desktop flows, as described in the "Handling the response and exchanging the code" section [here](https://developers.google.com/identity/protocols/OAuth2InstalledApp).

maffoo:
@adamcw, PTAL at the latest changes.

adamcw:
You need to fix the defaults for the `allowOAuthLogin` and `allowUsernameLogin` so that it works even when the server doesn't return proper values.

maffoo:
This works for me when connecting to an pre-oauth manager as well as a newer manager both with and without oauth enabled. Can you describe in a bit more detail what problem you're seeing?

adamcw:
I get the following build errors

```
app/elements/labrad-app.ts(31,3): error TS2300: Duplicate identifier 'allowUsernameLogin'.
app/elements/labrad-app.ts(34,3): error TS2300: Duplicate identifier 'allowOAuthLogin'.
app/elements/labrad-app.ts(37,3): error TS2300: Duplicate identifier 'allowUsernameLogin'.
app/elements/labrad-app.ts(40,3): error TS2300: Duplicate identifier 'allowOAuthLogin'.
app/scripts/manager.ts(38,3): error TS2300: Duplicate identifier 'clientId'.
app/scripts/manager.ts(39,3): error TS2300: Duplicate identifier 'clientSecret'.
app/scripts/manager.ts(43,3): error TS2300: Duplicate identifier 'clientId'.
app/scripts/manager.ts(44,3): error TS2300: Duplicate identifier 'clientSecret'.
```

Then when trying to login (after restarting both the manager and the scalabrad-web server), I get the following login dialog:

screenshot from 2016-08-08 09 26 23

maffoo:
Hmm, that's strange; it builds fine for me. You're on 60765677281ab02da94f9e0a8e7652bb31ef3087 with no local modifications to the tree?

adamcw:
For anyone's future reference we resolved this as related to a incorrect merge after a force-push to the branch.

adamcw:
LGTM!

Issue #256, opened 2016-08-04
Change Array<Foo> to Foo[]


maffoo:
The style guide says to prefer `Foo[]` for array types instead of `Array<Foo>`. We've been changing things piecemeal with new code, but it would be good to just switch to the preferred notation in one fell swoop.

PR #257, opened 2016-08-04
[#143, Part 2] Added filtering of servers in Node view based on autostart status.

adamcw:
Partial fix for #143. Added controls for filtering the list of servers based on if they appear in an autostart list or not. 

This required refactoring the code to no longer use computed properties. Specifically, with the original way, when filtering the list, all state information of the servers would be lost as the list would just get recreated from scratch. The new code isn't substantially more complicated than the existing, despite having to properly keep the list state current.

This also fixes an unreported bug where the control states were incorrect on page load due to the server controls not communicating their state to one another.

### DIFF
--- a/client-js/app/elements/labrad-nodes.html
+++ b/client-js/app/elements/labrad-nodes.html
@@ -282,7 +282,8 @@
                       server={{server}}
                       instance-name={{node.instanceName}}
                       node={{node.name}}
-                      status={{node.status}} />
+                      status={{node.status}}
+                      autostart={{node.autostart}} />
                   </template>
                 </td>
               </template>
--- a/client-js/app/elements/labrad-nodes.html
+++ b/client-js/app/elements/labrad-nodes.html
@@ -226,17 +226,25 @@
         <thead>
           <th>Global Servers</th>
           <th></th>
-          <template is="dom-repeat" items="{{nodeNames}}" as="node">
+          <template is="dom-repeat"
+                    items="{{nodeNames}}"
+                    as="node"
+                    sort="sortNodes">
             <th><labrad-node-controller places={{places}} api={{api}} name={{node}}></labrad-node-controller></th>
           </template>
         </thead>
         <tbody>
-          <template is="dom-repeat" items="{{globalServersFiltered}}" as="server">
+          <template is="dom-repeat"
+                    items="{{globalServersFiltered}}"
+                    as="server">
             <tr>
-              <td class="name">{{server.name}}</td>
-              <td class="version">{{server.version}}</td>
-              <template is="dom-repeat" items="{{server.nodes}}" as="node">
-                <td class="controls">
+              <td class='name'>{{server.name}}</td>
+              <td class='version'>{{server.version}}</td>
+              <template is="dom-repeat"
+                        items="{{server.nodes}}"
+                        as="node"
+                        sort="sortNodes">
+                <td class='controls'>
                   <template is="dom-if" if="{{node.exists}}">
                     <labrad-instance-controller
                       places={{places}}
--- a/client-js/app/elements/labrad-nodes.html
+++ b/client-js/app/elements/labrad-nodes.html
@@ -146,16 +146,14 @@
 <dom-module id="labrad-nodes">
   <style>
     :host {
-      display: block;
+      height: 100%;
+      overflow: hidden;
+      display: flex;
     }
     #container {
-      display: flex;
-      flex-grow: 1;
+      width: 100%;
     }
     #left-column {
-      flex-basis: 100%;
-      flex-direction: column;
-      flex-grow: 1;
       border-right: 1px solid #AAA;
     }
     table {
@@ -204,7 +202,7 @@
     }
   </style>
   <template>
-   <div id="container" class="fit">
+   <div id="container">
    <paper-header-panel id="left-column">
     <div class="paper-header" id="buttons">
       <paper-icon-button
--- a/client-js/app/elements/labrad-nodes.html
+++ b/client-js/app/elements/labrad-nodes.html
@@ -146,6 +146,16 @@
     :host {
       display: block;
     }
+    #container {
+      display: flex;
+      flex-grow: 1;
+    }
+    #left-column {
+      flex-basis: 100%;
+      flex-direction: column;
+      flex-grow: 1;
+      border-right: 1px solid #AAA;
+    }
     table {
       border-collapse: collapse;
       width: 100%;
@@ -178,9 +188,40 @@
     td.controls {
       width: 350px;
     }
+    #buttons {
+      height: 50px;
+    }
+    #buttons paper-icon-button {
+      margin: 5px;
+    }
+    .autostartFilterOn {
+      color: #000;
+    }
+    .autostartFilterOff {
+      color: #aaa;
+    }
   </style>
   <template>
-    <div>
+   <div id="container" class='fit'>
+   <paper-header-panel id="left-column">
+    <div class="paper-header" id="buttons">
+      <paper-icon-button
+        hidden$={{isAutostartFiltered}}
+        class='autostartFilterOff'
+        icon="stars"
+        id="star"
+        on-click="autostartFilter"
+        title="Show Only Autostart Servers"></paper-icon-button>
+      <paper-icon-button
+        hidden$={{!isAutostartFiltered}}
+        class='autostartFilterOn'
+        icon="stars"
+        id="star"
+        on-click="autostartFilter"
+        title="Show All Servers"></paper-icon-button>
+    </div>
+
+    <div class='fit'>
       <table>
         <thead>
           <th>Global Servers</th>
@@ -251,5 +292,7 @@
         </tbody>
       </table>
     </div>
+   </paper-header-panel>
+   </div>
   </template>
 </dom-module>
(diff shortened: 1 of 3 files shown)


client-js/app/elements/labrad-nodes.html
```diff
@@ -180,9 +190,40 @@
...
+    }
   </style>
   <template>
-    <div>
+   <div id="container" class='fit'>
+   <paper-header-panel id="left-column">
+    <div class="paper-header" id="buttons">
+      <paper-icon-button
+        hidden$={{isAutostartFiltered}}
+        class='autostartFilterOff'
```
maffoo: Mixed single and double quotes; not a big deal, but my OCD is forcing me to comment :-)

client-js/app/elements/labrad-nodes.html
```diff
@@ -180,9 +190,40 @@
...
+        icon="stars"
+        id="star"
+        on-click="autostartFilter"
+        title="Show Only Autostart Servers"></paper-icon-button>
+      <paper-icon-button
```
maffoo, line 215: We should make an element that encapsulates this pattern of an icon toggle button that can change colors. Seems like that would be useful in a number of places. Not for this PR, obviously.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -231,15 +238,20 @@ export class LabradNodeController extends polymer.Base {
...
   name: string;
   version: string;
+  autostart: boolean;
   errorString: string;
   errorException: string;
-  nodes: Array<{name: string, exists: boolean, status?: string, instanceName?: string}>;
+  nodes: Array<NodeServerStatus>;
 }
 
+type NodeServerStatus = {name: string, exists: boolean, autostart?: boolean, status?: string, instanceName?: string};
```
maffoo: line too long?

client-js/app/elements/labrad-nodes.ts
```diff
@@ -231,15 +238,20 @@ export class LabradNodeController extends polymer.Base {
 interface ServerInfo {
   name: string;
   version: string;
+  autostart: boolean;
   errorString: string;
   errorException: string;
-  nodes: Array<{name: string, exists: boolean, status?: string, instanceName?: string}>;
+  nodes: Array<NodeServerStatus>;
```
maffoo: `NodeServerStatus[]` (or leave it for #256)

client-js/app/elements/labrad-nodes.ts
```diff
@@ -253,29 +265,72 @@ export class LabradNodes extends polymer.Base {
   @property({type: Number, value: 0})
   kick: number;
 
+  @property({type: Boolean, value: false, notify: true})
+  isAutostartFiltered: boolean;
+
+  @property({type: Object, value: []})
```
maffoo: `type: Array`. Also the next three properties. In addition, I think the recommendation with mutable initial values is to use an initializer function, to ensure that you don't share mutable values between element instances:

``` typescript
value: () => []
```

client-js/app/elements/labrad-nodes.html
adamcw: Yeah, my brain struggles with the swapping between TypeScript (single) and HTML (double). I was actually thinking to mention we may want to standardise HTML to single quotes to make it more consistent. I do believe the style guide allows either as long as it's consistent.

client-js/app/elements/labrad-nodes.html
adamcw, line 215: A helper icon/button would be very good, especially for setting `title` text. I'm not happy with my `hidden` hacks, but it was the nicest way I could find to do it without breaking the separation of script and style.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -63,8 +63,15 @@ export class LabradInstanceController extends polymer.Base {
     return places.serverUrl(instanceName);
   }
 
-  ready() {
-    this.updateButtonState(this.status);
+  attached() {
+    this.async(() => {
+      this.fire('labrad-instance-controller::ready', {
```
maffoo, line 68: Does the double colon here have any special meaning? Or are the names here arbitrary strings and this is just to namespace things?

adamcw, line 68: Arbitrary strings, I did it to namespace things. We could choose any operator, I've just always used `::` in the past.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -253,29 +265,72 @@ export class LabradNodes extends polymer.Base {
...
-   * DOM and ready to be cleaned up.
-   */
+  @listen('labrad-instance-controller::ready')
+  onLabradInstanceControllerReady(event) {
+    // If a new labrad-instance-controller has readied, we need to broadcast
+    // the state of each controller of that server type to all others of the
+    // same type. This informs existing servers of the new server state, and
+    // informs the new server of the existing servers' states.
+    const instances = Polymer.dom(this.root).querySelectorAll('labrad-instance-controller');
+    for (var i = 0; i < instances.length; i++) {
```
maffoo:
``` typescript
for (let instance of instances) {
  ...
}
```

client-js/app/elements/labrad-nodes.ts
```diff
@@ -253,29 +265,72 @@ export class LabradNodes extends polymer.Base {
...
     this.lifetime.close();
   }
 
+
   defined(x: any): boolean {
     return (typeof x !== 'undefined') && (x !== null);
   }
 
+
+  autostartFilter() {
```
maffoo: maybe `toggleAutostartFilter`?

client-js/app/elements/labrad-nodes.ts
```diff
@@ -331,152 +392,227 @@ export class LabradNodes extends polymer.Base {
   }
 
 
-  /**
-   * Find the index in the node info list of a node with the given name.
-   *
-   * If no such node is found, returns -1.
```
maffoo, line 337: I think the comment is helpful; at least that it returns -1 if not found.

client-js/app/elements/labrad-nodes.ts
adamcw: "Evaluated at call time

The default argument gets evaluated at call time, so unlike e.g. in Python, a new object is created each time the function is called."

In ECMAScript6, the default parameter doesn't suffer this problem. Do you have a citation for how this is an issue?

client-js/app/elements/labrad-nodes.ts
```diff
@@ -331,152 +392,227 @@ export class LabradNodes extends polymer.Base {
...
-  private _nodeIndex(name: string): number {
-    var idx = -1;
-    for (var i = 0; i < this.info.length; i++) {
-      if (this.info[i].name === name) {
+  private getNodeIndex(item: (NodeStatus | ServerDisconnectMessage), array: NodeStatus[]): number {
+    let idx = -1;
+    for (let i = 0; i < array.length; ++i) {
+      if (array[i].name === item.name) {
         idx = i;
         break;
```
maffoo: To match `getServerIndex`, could ditch `idx`, `return i;` here and then `return -1;` at the bottom.

client-js/app/elements/labrad-nodes.ts
adamcw, line 337: I figured it was self evident given this is how `indexOf` works, but I can put it back.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -331,152 +392,227 @@ export class LabradNodes extends polymer.Base {
...
+    this.updateNodeServerBinding('globalServersFiltered');
+    this.updateNodeServerBinding('localServersFiltered');
+
+    this.updateFilters();
+
+    this.kick++;
+  }
+
+
+  private _serverStatusMap(statuses: ServerStatus[]): Map<String, ServerStatus> {
```
maffoo: Leading underscore not needed.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -331,152 +392,227 @@ export class LabradNodes extends polymer.Base {
...
+    for (const node of this.info) {
+      onlineNodes.add(node.name);
+    }
+
+    // Backwards as we are splicing from the list as we iterate.
+    const nodes = this[serversName][serverIdx].nodes;
+    for (let nodeIdx = nodes.length - 1; nodeIdx >= 0; --nodeIdx) {
+      const node = nodes[nodeIdx];
+      if (!onlineNodes.has(node.name)) {
+        this.splice(`${serversName}.#${serverIdx}.nodes`, nodeIdx, 1);
```
maffoo, line 578: A comment about what is going on here would be helpful. That `#` is rather cryptic.

client-js/app/elements/labrad-nodes.ts
maffoo, line 337: Maybe we could name these `indexOfServer` and `indexOfNode`, then clear they work like `indexOf`, as you say. Not a big deal either way.

client-js/app/elements/labrad-nodes.ts
adamcw: You can use `const` with the `for..of` syntax.

``` TypeScript
for (const instance of instances) {
  ...
}
```

I have done this, but because of the cast, it's not exactly nicer..

maffoo:
There's a rendering bug with the blue footer when the list of servers gets long enough to flow offscreen. After fixing that, LGTM.

PR #258, opened 2016-08-04
[#143, Part 3] Change nodeNames in labrad-nodes to be a non-computed property.

adamcw:
Partial fix for #143. Attempting to move the code base away from computed properties and `kick` hacks, this keeps the same `nodeNames` functionality by keeping a list up-to-date rather than computing it on the fly.

### DIFF
--- a/client-js/app/elements/labrad-nodes.html
+++ b/client-js/app/elements/labrad-nodes.html
@@ -228,16 +228,24 @@
         <thead>
           <th>Global Servers</th>
           <th></th>
-          <template is="dom-repeat" items="{{nodeNames}}" as="node">
+          <template is="dom-repeat"
+                    items="{{nodeNames}}"
+                    as="node"
+                    sort="sortNodes">
             <th><labrad-node-controller places={{places}} api={{api}} name={{node}}></labrad-node-controller></th>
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-nodes.ts
```diff
@@ -262,9 +262,6 @@ export class LabradNodes extends polymer.Base {
   @property({type: Object})
   managerApi: ManagerApi;
 
-  @property({type: Number, value: 0})
-  kick: number;
```
maffoo, line 293: No more kick! Woot!

client-js/app/elements/labrad-nodes.html
```diff
@@ -228,16 +228,24 @@
         <thead>
           <th>Global Servers</th>
           <th></th>
-          <template is="dom-repeat" items="{{nodeNames}}" as="node">
+          <template is="dom-repeat"
+                    items="{{nodeNames}}"
+                    as="node"
+                    sort="sortNodes">
```
maffoo, line 234: Did not know you can sort on a `dom-repeat`. That's nice.

maffoo:
LGTM

PR #259, opened 2016-08-04
[#143, Part 4, Final] Fixes issue with nodes changing while view is filtered.

adamcw:
Implements #143. Currently the view breaks if you are filtering to just autostart servers and then a node comes or goes offline. This fixes this issue by ensuring the `globalServers` and `localServers` lists are also being kept up-to-date, as they form the basis of the post-filtered list.

### DIFF
--- a/client-js/app/elements/labrad-nodes.ts
+++ b/client-js/app/elements/labrad-nodes.ts
@@ -467,8 +467,15 @@ export class LabradNodes extends polymer.Base {
 
     this.updateNodeServerBinding('globalServersFiltered');
     this.updateNodeServerBinding('localServersFiltered');
+    this.updateNodeServerBinding('globalServers');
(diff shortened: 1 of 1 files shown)


client-js/app/elements/labrad-nodes.ts
```diff
@@ -480,11 +487,19 @@ export class LabradNodes extends polymer.Base {
...
+    // If a node goes offline that has a server marked as autostart that no
+    // other does, and we are currently filtering, then we need to update the
+    // filters to hide the invalid server(s).
+    if (this.isAutostartFiltered) {
+      this.updateFilters();
+    }
   }
 
 
+
```
maffoo: Probably don't need extra blank line.

maffoo:
LGTM

PR #261, opened 2016-08-05
Transformed `Array` to `[]` notation. Fixed `value: []`.

### LINKED ISSUE #256: Change Array<Foo> to Foo[]

adamcw:
Fixes #256. Cleaned up properties being directly declared to `[]` instead of `() => []` while I was at it.


maffoo:
LGTM

PR #262, opened 2016-08-05
[#182, Part 1] Refactor drawZoomRectangle to be a generic.

adamcw:
Implements #182.

Part 1: Refactors the drawZoomRectangle function so that it can be recycled to draw a selection rectangle for the fit functions as well.

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -2,6 +2,7 @@ import 'd3';
 import THREE from 'three';
 import {viridisData} from '../scripts/colormaps';
 import * as datavault from '../scripts/datavault';
+import {fitParabola, fitExponential} from '../scripts/fit';
 import {Lifetime} from '../scripts/lifetime';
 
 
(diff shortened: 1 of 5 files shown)


client-js/app/elements/labrad-plot.ts
```diff
@@ -1126,33 +1127,36 @@ export class Plot extends polymer.Base {
...
+      clip(y - this.margin.top, 0, this.height + offset)
+    ];
+  }
+
+
+  /**
    * Zoom into a selected rectangular region on the graph.
    */
-  private drawZoomRectangle(): void {
+  private drawRectangle(callback: Function): void {
```
maffoo: As an alternate to taking a callback function, you can have this return promise, something like. `Promise<{rect, x1, y1, x2, y2}>`. Then you can use this with async/await for inline code. And you can still call it in callback style if desired by calling `then` on the returned promise.

maffoo:
Consider using Promise instead of callbacks. Otherwise LGTM.

adamcw:
@maffoo I made the function a bit more generic at the same time, so probably needs another look before I merge it.

client-js/app/elements/labrad-plot.ts
```diff
@@ -1144,65 +1145,68 @@ export class Plot extends polymer.Base {
...
+          rect.style.display = 'block';
+        })
+        .on('mouseup', () => {
+          const [x, y] = this.mousePositionClipped(0);
+          d3.select(window)
+            .on('mousemove', null)
+            .on('mouseup', null);
+
+          if (x === originX || y === originY) {
+            return;
```
maffoo, line 1447: I this case should probably reject the promise.

maffoo:
One question about rejecting the promise if the box is degenerate (we might also want to allow this by pressing ESC mid-drag or something like that, though needn't be in this PR). Otherwise LGTM.

PR #263, opened 2016-08-05
[#182, Part 2] Add new data demos for parabola and exponential functions.

adamcw:
Implements #182.

Part 2: Add new data demos for parabola and exponential functions, used for testing fitting them.

### DIFF
--- a/server/src/test/scripts/data_demo.py
+++ b/server/src/test/scripts/data_demo.py
@@ -25,6 +25,37 @@ def demo(func):
     demos.append(func)
     return func
 
+@demo
+def demo_1d_exponential(dv):
+    dv.new('demo_1d_exponential', ['x [ms]'], ['y [V]'])
+    add_params(dv)
+
+    # Generate points between -5 and 5 in 0.1 increments
+    for i in [x * 0.1 for x in range(-50, 50)]:
+        x = i
+        y = pow(math.e, x) + random.random()*10
+        dv.add([x, y])
+        time.sleep(0.002)
+
+@demo
+def demo_1d_parabola(dv):
+    dv.new('demo_1d_parabola', ['x [ms]'], ['y [V]'])
+    add_params(dv)
+    for i in xrange(-1000, 1000):
+        x = i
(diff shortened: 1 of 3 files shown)


server/src/test/scripts/data_demo.py
```diff
@@ -25,6 +25,37 @@ def demo(func):
...
     return func
 
+@demo
+def demo_1d_exponential(dv):
+    dv.new('demo_1d_exponential', ['x [ms]'], ['y [V]'])
+    add_params(dv)
+
+    # Generate points between -5 and 5 in 0.1 increments
+    for i in [x * 0.1 for x in range(-50, 50)]:
+        x = i
```
maffoo: I would use `i` for the integer index and `x` for the float value:

``` python
for x in (i * 0.1 for i in xrange(-50, 50)):
    ...
```

The parens and `xrange` make this a generator expression instead of precreating a list.

server/src/test/scripts/data_demo.py
```diff
@@ -25,6 +25,37 @@ def demo(func):
...
+        y = pow(math.e, x) + random.random()*10
```
maffoo: `math.e**x`

maffoo: Also note that `math.e**(-x)` would be a lot more common for typical datasets, but it's fine either way of course.

server/src/test/scripts/data_demo.py
maffoo: Or more simply:

``` python
for i in xrange(-50, 50):
    x = i * 0.1
    ...
```

maffoo:
Minor nits; otherwise LGTM

server/src/test/scripts/data_demo.py
adamcw: Good call, not sure how I thought that was a good idea.

server/src/test/scripts/data_demo.py
adamcw: Today I Learned: *\* is actually faster than math.pow, good to know.

PR #264, opened 2016-08-05
[#182, Part 3] Changed titles on plotter functions to be cased correctly.

### LINKED ISSUE #182: Grapher: Allow for basic fitting

adamcw:
Fixes #182.

Part 3: Changed titles on plotter functions to be cased correctly.


maffoo:
LGTM

PR #265, opened 2016-08-05
Use 'star' and 'star-border' icons instead of 'stars'.

maffoo:
@adamcw

### DIFF
--- a/client-js/app/elements/labrad-grapher.html
+++ b/client-js/app/elements/labrad-grapher.html
@@ -123,8 +123,8 @@
       <div id="container">
         <paper-header-panel id="left-column">
           <div class="paper-header" id="buttons">
-            <paper-icon-button icon="star" id="star" on-click="starClicked" title="Show only starred items."></paper-icon-button>
-            <paper-icon-button icon="delete" id="trash" on-click="trashClicked" title="Show trashed items."></paper-icon-button>
+            <paper-icon-button icon="star" id="star" on-click="starClicked" title="Show Only Starred Items"></paper-icon-button>
+            <paper-icon-button icon="delete" id="trash" on-click="trashClicked" title="Show Trashed Items"></paper-icon-button>
           </div>
           <div id="listing">
             <iron-list items="{{filteredListItems}}" as="item" class='fit'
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-nodes.html
```diff
@@ -73,10 +73,10 @@
       top: 5px;
     }
     paper-icon-button.autostart {
-      color: #75787B;
+      color: #85888B;
```
adamcw: Change the colors back please, the grey doesn't match the other icons now.

client-js/app/elements/labrad-nodes.html
```diff
@@ -73,10 +73,10 @@
...
     }
     paper-icon-button.autostartOn {
-      color: #0F9D58;
+      color: #0FBD78;
```
adamcw: I chose a Google approved shade of green, and I preferred it.

client-js/app/elements/labrad-grapher.html
```diff
@@ -123,8 +123,8 @@
       <div id="container">
         <paper-header-panel id="left-column">
           <div class="paper-header" id="buttons">
-            <paper-icon-button icon="stars" id="star" on-click="starClicked" title="show starred items"></paper-icon-button>
-            <paper-icon-button icon="delete" id="trash" on-click="trashClicked" title="show trashed items"></paper-icon-button>
+            <paper-icon-button icon="star" id="star" on-click="starClicked" title="Show only starred items."></paper-icon-button>
+            <paper-icon-button icon="delete" id="trash" on-click="trashClicked" title="Show trashed items."></paper-icon-button>
```
adamcw: No other titles have full stops, but they are usually All Upper Title Case. I don't know if there is a style guide, but that's what I've been using.

adamcw:
I like the other icons better.. but if you think this is an improvement I'm not THAT attached to them.

client-js/app/elements/labrad-nodes.html
maffoo: Ok, reverted. I found this and the green too dark for my taste, but it's just that, taste :-)

maffoo:
I like having the star outline; I think it sets these off visually better from surrounding UI. But yes, this is all highly subjective...

adamcw:
These have grown on me already so, LGTM.

PR #268, opened 2016-08-09
[#182, Part 5] Add functionality to fit Parabola and Exponential functions.

adamcw:
Implements #182. Adds UI and functionality to the Grapher to allow the user to select a bounding region to fit either a parabola or exponential function.

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -435,7 +435,7 @@ export class Plot extends polymer.Base {
     const p = this;
 
     // Make room for the color bar if necessary.
-    if (p.numIndeps == 2) {
+    if (p.numIndeps === 2) {
       p.margin.right = PLOT_RIGHT_MARGIN + COLOR_BAR_WIDGET_SIZE;
     }
 
@@ -506,8 +506,8 @@ export class Plot extends polymer.Base {
             .style('text-anchor', 'middle')
             .text(p.yLabel);
 
-    // Color Bar Axis
-    if (p.numIndeps == 2) {
+    // Color Bar Axis.
+    if (p.numIndeps === 2) {
       p.zAxis = d3.svg.axis();
       p.zAxis.orient('right')
              .ticks(COLOR_BAR_NUM_TICKS)
@@ -523,7 +523,7 @@ export class Plot extends polymer.Base {
         })
       }
 
-      // The Color Bar Gradient
+      // The Color Bar Gradient.
       p.svg.append('defs')
              .append("linearGradient")
              .attr("id", "ColorBarGradient")
@@ -543,7 +543,7 @@ export class Plot extends polymer.Base {
       // for the overall app to make Polymer paths work.
       const gradientFill = `url('${location.href}#ColorBarGradient')`;
 
-      // Color Bar Rectangle
+      // Color Bar Rectangle.
       p.svg.append('rect')
              .attr('id', 'color-bar')
              .attr('fill', gradientFill)
@@ -584,7 +584,7 @@ export class Plot extends polymer.Base {
     const plotWidth = Math.max(plotBounds.width, PLOT_MIN_WIDTH);
     const plotHeight = Math.max(plotBounds.height, PLOT_MIN_HEIGHT);
 
-    // The inner dimensions of the plot, sans margins
+    // The inner dimensions of the plot, sans margins.
     this.width = plotWidth - this.margin.left - this.margin.right;
     this.height = plotHeight - this.margin.top - this.margin.bottom;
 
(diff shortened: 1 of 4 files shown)


client-js/app/elements/labrad-plot.ts
```diff
@@ -654,56 +685,185 @@ export class Plot extends polymer.Base {
...
+    return this.sceneObjects[0];
+  }
+
+
+  /**
+   * Plots a line given a set of data, adding it to the scene.
+   */
+  private plotLine(data: number[][], yColumn: number = 1,
+                   color: string = "#000", lineWidth: number = 1,
+                   lines: THREE.Line[]) {
```
maffoo: I would suggest just returning the line and letting the caller push it onto the appropriate list. Also note that it's best not to put required arguments after arguments with default values.

adamcw: I'll fix this in Part 7.

@maffoo See: https://github.com/labrad/scalabrad-web/pull/271/commits/3ff73c10957e03fca46aa940d989f0e7199547d2

client-js/app/elements/labrad-plot.ts
```diff
@@ -654,56 +685,185 @@ export class Plot extends polymer.Base {
...
+      const x = d[0];
+      if (x < this.fitBounds.xMin || x > this.fitBounds.xMax) {
+        continue;
       }
 
-      for (let row of data) {
-        dataPoints[offset] = row[0];
-        dataPoints[offset + 1] = row[i + 1];
-        offset += 3;
+      const datum = [x]
```
maffoo: Looks like this is not needed.

client-js/app/elements/labrad-plot.ts
```diff
@@ -654,56 +685,185 @@ export class Plot extends polymer.Base {
...
+    lines.push(line);
+  }
 
-      // Raw data is stored inside the geometry for more efficient
-      // reprojection.
-      const dataPoints = new Float64Array(length * 3);
-      let offset = 0;
+  /**
+   * Returns a dictionary where each trace (one-indexed) has been filtered to
+   * just include points that are within the currently selected fit area.
```
maffoo: This returns an array instead of a dictionary. Also, would be helpful to describe the shape of the output array a bit more; just saying "one-indexed" is a bit unclear.

client-js/app/elements/labrad-plot.ts
```diff
@@ -654,56 +685,185 @@ export class Plot extends polymer.Base {
...
+      return dataInFitBounds;
+    }
 
-      // Add the last point of data if maxima exists to avoid gaps.
-      if (this.lastData) {
-        dataPoints[offset] = this.lastData[0];
-        dataPoints[offset + 1] = this.lastData[i + 1];
-        offset += 3;
+    const d = data[0];
+    for (let i = 0; i < d.length; ++i) {
```
maffoo: Here you push `d.length` arrays, even though there are only `d.length-1` traces.

client-js/app/elements/labrad-plot.ts
```diff
@@ -151,6 +168,14 @@ export class Plot extends polymer.Base {
...
   private displaySurface: number = 2;
 
+  private fitMode: number = FIT_OPTION_NONE;
+
+  private lines: THREE.Line[] = [];
+  private linesFitParabolas: THREE.Line[] = [];
+  private linesFitExponentials: THREE.Line[] = [];
+
+  private fitParabolaCoefficients: {A: number, B: number, C: number}[] = [];
+  private fitExponentialCoefficients: {A: number, B: number}[] = [];
```
maffoo: These should include label, legend and units fields.

client-js/app/elements/labrad-plot.ts
```diff
@@ -132,6 +142,13 @@ export class Plot extends polymer.Base {
     left: PLOT_LEFT_MARGIN
   };
 
+  private fitBounds = {
+    xMin: 0,
+    xMax: 0,
+    yMin: 0,
+    yMax: 0
```
maffoo: I would prefer to use out-of-band sentinel values (e.g. `NaN`) here, but not sure if that complicates things elsewhere. I guess a degenerate box with zero area is also a fairly clear indication not to show the fit rect.

client-js/app/elements/labrad-plot.ts
```diff
@@ -1126,49 +1316,56 @@ export class Plot extends polymer.Base {
...
+      clip(y - this.margin.top - PLOT_TOP_OFFSET, 0, this.height + offset)
+    ];
+  }
+
+
+  /**
    * Zoom into a selected rectangular region on the graph.
    */
-  private drawZoomRectangle(): void {
+  private drawRectangle(rect: HTMLElement): Promise<RectangleBound> {
```
maffoo, line 1417: Didn't this change already happen in a previous PR? Not sure why it's showing up here too.

maffoo:
Some minor comments. Also, it looks like the base of this PR might be messed up since it seems to include some already-reviewed changes like the rect handling and additions to `data_demo.py`. Other than those things, LGTM.

PR #270, opened 2016-08-09
Bump polymer to version 1.6.x

maffoo:
The latest version of paper-checkbox wants polymer 1.6.x, so having it pinned to 1.5.x was breaking the teamcity build.


adamcw:
LGTM.

Issue #272, opened 2016-08-09
Make it possible to edit the registry with the keyboard only


maffoo:
Using the registry editor currently requires a lot of mousing around. It would be very nice to make it possible to use the keyboard. Here is a minimal set of keyboard controllability I'd like to add:
1. Use up and down arrow keys to highlight an item in the list.
2. Press `enter` to navigate into the highlighted directory (or up a level) or to edit the highlighted key.
3. When the edit box pops up, the edit area displaying the value should be focused so user can just start typing.
4. Press `shift+enter` to keep the edited value. It should also be possible to press `shift+enter` from the value edit area when creating a new key.

As part of this keyboard control, we should ensure that tab order works in a sensible way with the various dialog boxes, so that the user can tab between text fields and dialog buttons to control the dialogs. I think this _sort of_ works, but we should just try it out systematically.

Regarding number 1 above, we also have a filter box to restrict the set of keys that are shown. It would be good if this box could interact nicely with other keyboard controls so that it can be used without clicking on the box and then back on the list to select something. For example, pressing up arrow when the first element in the registry listing is highlighted could move the focus to the filter box; pressing down arrow from there would again highlight the first item in the listing. Alternately, we could just capture keystrokes other than up/down arrow and enter and use them to modify the filter text _as if_ those characters were typed in the box. I'm not sure the best thing to do here, so some experimentation is in order.

adamcw:
As of #279 all tab ordering for the dialogs in the registry are correct.

PR #275, opened 2016-08-09
[#272, Part 1] Add keyboard commands to traverse up and down registry keys.

adamcw:
Partially Implements #272. Added keyboard commands for traversing the registry directory structure:
- KEY_UP/KEY_DOWN: Move up and down the list of directories/keys.
- KEY_LEFT/KEY_BACKSPACE: Traverse to the parent directory.
- KEY_RIGHT/KEY_RETURN: Traverse to the child directory if a directory is selected.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -163,8 +163,6 @@
       <iron-a11y-keys target="[[target]]" keys="left backspace"
                       on-keys-pressed="cursorBack"></iron-a11y-keys>
       <table>
-
-
         <tbody is="selectable-table" id="combinedList" selected="{{selectedIdx}}">
           <template id="fullList" is="dom-repeat" items="{{listItems}}" filter="filterFunc">
             <tr name="{{item.name}}"
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -156,7 +156,15 @@
   </style>
   <template>
     <div id="reg-area">
+      <iron-a11y-keys target="[[target]]" keys="up down"
+                      on-keys-pressed="cursorMove"></iron-a11y-keys>
+      <iron-a11y-keys target="[[target]]" keys="right enter"
+                      on-keys-pressed="cursorTraverse"></iron-a11y-keys>
+      <iron-a11y-keys target="[[target]]" keys="left backspace"
+                      on-keys-pressed="cursorBack"></iron-a11y-keys>
       <table>
+
+
         <tbody is="selectable-table" id="combinedList" selected="{{selectedIdx}}">
           <template id="fullList" is="dom-repeat" items="{{listItems}}" filter="filterFunc">
             <tr name="{{item.name}}"
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -44,13 +44,18 @@ export class LabradRegistry extends polymer.Base {
   }
 
 
-  getDefaultSelectedItem(): number {
+  getListOffset(): number {
     return this.path.length > 0 ? 1 : 0;
   }
 
 
+  getDefaultSelectedItem(): number {
+    return getListOffset();
+  }
+
+
   cursorMove(e) {
-    const offset = this.path.length > 0 ? 1 : 0;
+    const offset = this.getListOffset();
     const length = this.keys.length + this.dirs.length + offset;
 
     switch (e.detail.combo) {
@@ -163,7 +168,8 @@ export class LabradRegistry extends polymer.Base {
   selectedItem: string;
 
   _computeSelectedType(selectedIdx: number): string {
-    var offset = this.path.length > 0 ? 1 : 0; // account for parent '..' entry
+    // Account for parent '..' entry.
+    const offset = this.getListOffset();
     if (this.dirs && selectedIdx < this.dirs.length + offset) {
       return 'dir';
     } else {
@@ -172,9 +178,10 @@ export class LabradRegistry extends polymer.Base {
   }
 
   _computeSelectedItem(selectedIdx: number): string {
-    var offset = this.path.length > 0 ? 1 : 0; // account for parent '..' entry
-    var dirNames = (this.dirs || []).map(it => it.name);
-    var keyNames = (this.keys || []).map(it => it.name);
+    // Account for parent '..' entry.
+    const offset = this.getListOffset();
+    const dirNames = (this.dirs || []).map(it => it.name);
+    const keyNames = (this.keys || []).map(it => it.name);
     return dirNames.concat(keyNames)[selectedIdx - offset];
   }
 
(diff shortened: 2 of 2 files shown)


client-js/app/elements/labrad-registry.html
```diff
@@ -156,7 +156,15 @@
...
     <div id="reg-area">
+      <iron-a11y-keys target="[[target]]" keys="up down"
+                      on-keys-pressed="cursorMove"></iron-a11y-keys>
+      <iron-a11y-keys target="[[target]]" keys="right enter"
+                      on-keys-pressed="cursorTraverse"></iron-a11y-keys>
+      <iron-a11y-keys target="[[target]]" keys="left backspace"
+                      on-keys-pressed="cursorBack"></iron-a11y-keys>
       <table>
+
+
```
maffoo: nit: remove blank lines

client-js/app/elements/labrad-registry.ts
```diff
@@ -42,15 +43,77 @@ export class LabradRegistry extends polymer.Base {
...
+    if (this.selectedIdx === null) {
+      return;
+    }
+
+    const item = this.$.combinedList.selectedItem;
+    const link = item.querySelector('a');
+
+    // If we have a link, we want to traverse down.
+    if (link) {
+      this.fire('app-link-click', {path: link.href});
```
maffoo: This is causes a full page reload, rather than just doing a route change within the app. You want to use `link.path` instead.

adamcw:
@maffoo PTAL

client-js/app/elements/labrad-registry.ts
```diff
@@ -42,15 +43,76 @@ export class LabradRegistry extends polymer.Base {
...
+    const length = this.keys.length + this.dirs.length + offset;
+
+    switch (e.detail.combo) {
+      case 'up':
+        if (this.selectedIdx === null || this.selectedIdx === 0) {
+          this.set('selectedIdx', null);
+        } else {
+          this.set('selectedIdx', this.selectedIdx - 1);
+        }
+        break;
```
maffoo, line 68: After scrolling up and down for a while on a long registry listing, the selection can go offscreen. It seems we might need to prevent the default event action for up and down keypresses. Of course then we need to implement scrolling in some other way so the listing scrolls to keep the selection on screen; I know there are libraries that can help with this kind of programmatic scrolling control, but I don't know which one is best.

client-js/app/elements/labrad-registry.ts
```diff
@@ -42,15 +43,76 @@ export class LabradRegistry extends polymer.Base {
...
   }
 
+
+  getDefaultSelectedItem(): number {
+    return this.path.length > 0 ? 1 : 0;
+  }
+
+
+  cursorMove(e) {
+    const offset = this.path.length > 0 ? 1 : 0;
```
maffoo: This duplicates the function above. Would be nice to DRY that up.

client-js/app/elements/labrad-registry.ts
```diff
@@ -42,15 +43,76 @@ export class LabradRegistry extends polymer.Base {
...
   /**
-   * on a path change, we deselect everything, empty filterText
+   * On path change, select the default element and clear the filter.
    */
   @observe('path')
   pathChanged(newPath: string[], oldPath: string[]) {
-    this.selectedIdx = null;
-    this.filterText = '';
+    this.set('selectedIdx', this.getDefaultSelectedItem());
+    this.set('filterText', '');
```
maffoo, line 117: It's too bad we have to go through these polymer APIs since we lose typesafety...

client-js/app/elements/labrad-registry.ts
```diff
@@ -91,11 +153,7 @@ export class LabradRegistry extends polymer.Base {
 
   @computed()
   selected(selectedIdx: number): boolean {
-    if (selectedIdx !== null) {
-        return true;
-      } else {
-        return false;
-      }
+    return (selectedIdx !== null);
```
maffoo, line 161: LOL

client-js/app/elements/labrad-registry.ts
adamcw, line 68: Can you clarify the issue? The later part of your statement makes it just sound like if you scroll off where you've selected, then use the keys, nothing happens because you are adjusting the selection off-screen. The earlier part makes it sound like the selection gets out of sync or something though, so it's unclear what you are saying.

client-js/app/elements/labrad-registry.ts
```diff
@@ -42,15 +43,76 @@ export class LabradRegistry extends polymer.Base {
...
+  }
+
+
+  cursorMove(e) {
+    const offset = this.path.length > 0 ? 1 : 0;
+    const length = this.keys.length + this.dirs.length + offset;
+
+    switch (e.detail.combo) {
+      case 'up':
+        if (this.selectedIdx === null || this.selectedIdx === 0) {
```
maffoo, line 63: Not sure how I feel about allowing `selectedIdx` to go null by hitting up arrow on the first entry in the list. It feels strange for the bottom of the list to be a "hard stop" that I can't move past, but the top of the list lets me fall off.

client-js/app/elements/labrad-registry.ts
maffoo, line 68: Sorry for being unclear. Right now when you hit the arrow keys the selection changes _and_ the default event handling for arrow keys kicks in to scroll the view. This _sort of_ keeps the selection on screen, but they can get out of sync so that the view keeps scrolling because of the default event handling but the selection is offscreen. I think it would be best to disable the default event handling for arrow key events and then programmatically control the scrolling to make sure the selected item stays visible.

client-js/app/elements/labrad-registry.ts
adamcw, line 63: That's because I plan to have going up off the top to select the filter text box... I just haven't done that bit yet!

maffoo, line 63: I'm wondering if maybe it's better to just use `Tab` to move the focus to the filter box, since that's a more common way of moving around a GUI. But I'm happy to reserve judgement until we see how things work together.

client-js/app/elements/labrad-registry.ts
adamcw: This isn't repetition, they are two bits of code that do the same thing for different reasons.

The offset is always. It just so happens that we've currently decided that the default selected item is also the first item after the offset. However, `getDefaultSelectedItem` could be changed to `return 0;` which would be an appropriate default, but would break the offset.

I can fork them both from a "hasPath" function, but that seems a little silly.

maffoo: I realize they're doing different things, it's mostly just jarring to see these two identical snippets right next to each other given that one is in a nice function and we're not using it a couple of lines below :-)

Given that we have this same offset calculation in a couple of other places as well (for purposes of getting a list offset, not a selected index), I think it would make sense to pull out into its own method. `getDefaultSelectedItem` can then call that method (and can still be changed later, as you note).

client-js/app/elements/labrad-registry.ts
adamcw, line 63: Perhaps, we can try it both ways and see which turns out better, for now I'll make going off the end impossible.

adamcw:
@maffoo PTAL

maffoo:
LGTM

Issue #278, opened 2016-08-10
Convert labrad-registry element to use iron-list


maffoo:
This will make it possible to use programmatic scrolling when using keyboard to change the list selection (as done in #277 for the grapher).

PR #279, opened 2016-08-10
[#272, Part 2, Final] Add keyboard controls for editing Registry keys.

adamcw:
Implements #272 (along with PR #275). Added keyboard commands for editing keys in the Registry.
- KEY_RIGHT/KEY_RETURN: Edit the selected key.
- KEY_SHIFT+KEY_ENTER: Submit edit.
- KEY_ESCAPE: Close dialog without submitting.

Other interactions:
- KEY_RIGHT/KEY_RETURN/KEY_LEFT/KEY_BACKSPACE: Disabled when search is selected.
- KEY_RETURN: Will unfocus the search when selected.
- KEY_UP/KEY_DOWN: Will still navigate when search is selected, and will blur the search if used.

### DIFF
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -221,6 +221,8 @@ export class LabradRegistry extends polymer.Base {
    * that contain substring in filterText
    */
   filterFunc(item) {
+    const selectedIdx = (this.$.search.value) ? 0 : this.getDefaultSelectedItem();
+    this.set('selectedIdx', selectedIdx);
     return item.name.match(this.regex);
   }
 
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -160,23 +160,23 @@ export class LabradRegistry extends polymer.Base {
     switch (dialog.id) {
       case 'newKeyDialog':
         this.doNewKey();
-        break
+        break;
 
       case 'newFolderDialog':
         this.doNewFolder();
-        break
+        break;
 
       case 'editValueDialog':
         this.doEditValue();
-        break
+        break;
 
       case 'renameDialog':
         this.doRename();
-        break
+        break;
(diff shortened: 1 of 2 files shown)


client-js/app/elements/labrad-registry.ts
```diff
@@ -108,6 +142,61 @@ export class LabradRegistry extends polymer.Base {
...
+    if (!dialog) {
+      return;
+    }
+
+    event.detail.keyboardEvent.preventDefault();
+
+    switch (dialog.id) {
+      case 'newKeyDialog':
+        this.doNewKey();
+        break
```
maffoo: Semicolon after break. Same in other cases.

client-js/app/elements/labrad-registry.ts
```diff
@@ -132,6 +221,8 @@ export class LabradRegistry extends polymer.Base {
    * that contain substring in filterText
    */
   filterFunc(item) {
+    const selectedIdx = (this.$.search.value) ? 0 : this.getDefaultSelectedItem();
+    this.set('selectedIdx', selectedIdx);
```
maffoo, line 225: It seems a little bit strange to update the selected index here, especially since this gets called in a loop when the items are updated. But this might be unavoidable given the way this works now with `listItems` being a computed property. It would be nice to preserve the selected index as much as possible, for example keeping an item selected after editing it, where currently this causes the selection to be reset to the top when the item is edited and the list gets repopulated.

maffoo, line 225: Probably easiest to address this as part of #278

adamcw, line 225: I already have a branch ready for #278 that eliminates this issue and removes all computed properties in the process.

maffoo:
LGTM. Looking forward to getting #278 in as well :-)

adamcw:
Merged manually.

PR #280, opened 2016-08-11
[#278] Change Registry to use iron-list.

adamcw:
Implements #278. Updated/Refactored the Registry to use an `iron-list` instead of a `table` with `dom-repeat`. This allows for smooth scrolling on the registry when using keyboard controls.

This also refactored the code to get rid of all computed properties, as well as the last instance of the `kick` hack. Also fixed problems with submitting forms via hitting enter in text fields, or while OK is selected. In addition, various style guide compliance changes.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -61,7 +61,7 @@
       -webkit-user-select: none;
     }
     .row .key {
-      width: 400px;
+      width: 16em;
       white-space: nowrap;
       padding-right: 2em;
       overflow: hidden;
@@ -259,8 +259,8 @@
                 <div class="key mono"
                      name="{{item.name}}"
                      draggable="true">
-                  <iron-icon icon="communication:vpn-key" item-icon></iron-icon>
-                  <span>{{item.name}}</span>
+                  <iron-icon icon="communication:vpn-key"item-icon></iron-icon>
+                  <span title="{{item.name}}">{{item.name}}</span>
                 </div>
                 <div class="value label-wide mono">
                   <span>{{item.value}}</span>
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -181,7 +181,7 @@
       <!-- Dialog Controls -->
       <iron-a11y-keys target="[[target]]" keys="esc"
                       on-keys-pressed="dialogCancel"></iron-a11y-keys>
-      <iron-a11y-keys target="[[target]]" keys="shift+enter"
+      <iron-a11y-keys target="[[target]]" keys="shift+enter enter"
                       on-keys-pressed="dialogSubmit"></iron-a11y-keys>
       <!-- Cursor Controls -->
       <iron-a11y-keys target="[[target]]" keys="up down"
@@ -277,10 +277,8 @@
 
     <paper-dialog id="newKeyDialog" modal>
       <h1>Create New Key</h1>
-      <form>
-        <paper-input id="newKeyInput" label="Key" autofocus></paper-input>
-        <paper-textarea id="newValueInput" label="Value"></paper-textarea>
-      </form>
+      <paper-input id="newKeyInput" label="Key" autofocus></paper-input>
+      <paper-textarea id="newValueInput" label="Value"></paper-textarea>
       <div class="buttons">
         <paper-button on-tap="doNewKey" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
@@ -289,9 +287,7 @@
 
     <paper-dialog id="editValueDialog" modal>
       <h1>Edit <span>{{selected.name}}</span></h1>
-      <form>
-        <paper-textarea id="editValueInput" label="Value" autofocus></paper-textarea>
-      </form>
+      <paper-textarea id="editValueInput" label="Value" autofocus></paper-textarea>
       <div class="buttons">
         <paper-button on-tap="doEditValue" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
@@ -300,9 +296,7 @@
 
     <paper-dialog id="newFolderDialog" modal>
       <h1>Create New Folder</h1>
-      <form>
-        <paper-input id="newFolderInput" label="Name" autofocus></paper-input>
-      </form>
+      <paper-input id="newFolderInput" label="Name" autofocus></paper-input>
       <div class="buttons">
         <paper-button on-tap="doNewFolder" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
@@ -311,10 +305,8 @@
 
      <paper-dialog id="dragDialog" modal>
       <h1><span id="dragOp"></span> <span id="dragClass"></span> "<span id="originName"></span>" from <span id="originPath"></span></h1>
-      <form>
-        <paper-input id="dragNameInput" label="New Name" always-float-label></paper-input>
-        <paper-input id="dragPathInput" label="New Path" always-float-label></paper-input>
-      </form>
+      <paper-input id="dragNameInput" label="New Name" always-float-label></paper-input>
+      <paper-input id="dragPathInput" label="New Path" always-float-label></paper-input>
       <div class="buttons">
         <paper-button on-tap="doDragOp" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
@@ -323,10 +315,8 @@
 
     <paper-dialog id="copyDialog" modal>
       <h1>Copy <span>{{selectedType}}</span> "<span>{{selected.name}}</span>"?</h1>
-      <form>
-        <paper-input id="copyNameInput" label="New Name" always-float-label autofocus></paper-input>
-        <paper-input id="copyPathInput" label="New Path" always-float-label></paper-input>
-      </form>
+      <paper-input id="copyNameInput" label="New Name" always-float-label autofocus></paper-input>
+      <paper-input id="copyPathInput" label="New Path" always-float-label></paper-input>
       <div class="buttons">
         <paper-button on-tap="doCopy" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
@@ -335,9 +325,7 @@
 
     <paper-dialog id="renameDialog" modal>
       <h1>Rename <span>{{selectedType}}</span> "<span>{{selected.name}}</span>"?</h1>
-      <form>
-        <paper-input id="renameInput" label="New Name" always-float-label autofocus></paper-input>
-      </form>
+      <paper-input id="renameInput" label="New Name" always-float-label autofocus></paper-input>
       <div class="buttons">
         <paper-button on-tap="doRename" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -282,7 +282,7 @@
         <paper-textarea id="newValueInput" label="Value"></paper-textarea>
       </form>
       <div class="buttons">
-        <paper-button on-click="doNewKey" dialog-confirm>OK</paper-button>
+        <paper-button on-tap="doNewKey" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
@@ -293,7 +293,7 @@
         <paper-textarea id="editValueInput" label="Value" autofocus></paper-textarea>
       </form>
       <div class="buttons">
-        <paper-button on-click="doEditValue" dialog-confirm>OK</paper-button>
+        <paper-button on-tap="doEditValue" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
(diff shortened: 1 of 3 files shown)


client-js/app/elements/labrad-registry.html
```diff
@@ -170,41 +172,75 @@
...
-                      on-keys-pressed="dialogSubmit"></iron-a11y-keys>
-      <!-- Search Controls -->
-      <iron-a11y-keys target="[[target]]" keys="enter"
-                      on-keys-pressed="searchSubmit"></iron-a11y-keys>
-      <table>
-        <tbody is="selectable-table" id="combinedList" selected="{{selectedIdx}}">
-          <template id="fullList" is="dom-repeat" items="{{listItems}}" filter="filterFunc">
-            <tr name="{{item.name}}"
+  <div id="outer">
+    <!-- Dialog Controls -->
```
maffoo, line 179: Is there a reason you reordered these shortcut elements to put the dialog stuff first?

client-js/app/elements/labrad-registry.html
```diff
@@ -170,41 +172,75 @@
...
-      <!-- Dialog Controls -->
-      <iron-a11y-keys target="[[target]]" keys="esc"
-                      on-keys-pressed="dialogCancel"></iron-a11y-keys>
-      <iron-a11y-keys target="[[target]]" keys="shift+enter"
-                      on-keys-pressed="dialogSubmit"></iron-a11y-keys>
-      <!-- Search Controls -->
-      <iron-a11y-keys target="[[target]]" keys="enter"
-                      on-keys-pressed="searchSubmit"></iron-a11y-keys>
-      <table>
-        <tbody is="selectable-table" id="combinedList" selected="{{selectedIdx}}">
```
maffoo, line 190: Now that this is changed you should be able to get rid of the `selectable-table` element altogether.

client-js/app/elements/labrad-registry.html
```diff
@@ -170,41 +172,75 @@
...
+    <paper-header-panel id="left-column">
+      <div class="paper-header" id="buttons">
+        <paper-toolbar>
+          <paper-button on-click="newKeyClicked">
+            <iron-icon icon="create"></iron-icon>
+            new key
+          </paper-button>
+          <paper-button on-click="newFolderClicked">
+            <iron-icon icon="folder-open"></iron-icon>
+            new dir
```
maffoo: When horizontal space gets tight, the "new key" and "new dir" captions render as two lines. We should use `&nbsp;`.

client-js/app/elements/labrad-registry.html
```diff
@@ -170,41 +172,75 @@
...
+            </paper-input>
+          </div>
+        </paper-toolbar>
+      </div>
+      <div id="listing">
+        <iron-list items="{{filteredListItems}}"
+                  id="combinedList"
+                  as="item"
+                  selected-item="{{selected}}"
+                  selection-enabled>
```
maffoo: nit: hanging indent is misaligned

client-js/app/elements/labrad-registry.html
```diff
@@ -170,41 +172,75 @@
...
+            <iron-icon icon="code"></iron-icon>
+            rename
+          </paper-button>
+          <paper-button on-click="deleteClicked" disabled="[[!selected]]">
+            <iron-icon icon="delete"></iron-icon>
+            delete
+          </paper-button>
+          <div id="search-bar">
+            <paper-input id="search" value="{{filterText::input}}"
+              noLabelFloat>
```
maffoo, line 219: The hanging indents here are a bit inconsistent, sometimes using 4 spaces, here 2, and sometimes aligning with attribute on the first line. I think 4 spaces or aligned would be fine, but we should be consistent (at least within a single .html if not throughout the codebase).

client-js/app/elements/labrad-registry.html
```diff
@@ -214,130 +250,87 @@
...
               </template>
               <template is="dom-if" if="{{item.isKey}}">
-                <td class="key mono"
-                    name="{{item.name}}"
+                <div class="key mono"
+                      name="{{item.name}}"
                     draggable="true">
-                  <iron-icon icon="communication:vpn-key" item-icon class="key"></iron-icon>
+                  <iron-icon icon="communication:vpn-key" item-icon></iron-icon>
                   <span>{{item.name}}</span>
```
maffoo: It would be nice to show the key name as a tooltip in case it overflows.

client-js/app/elements/labrad-registry.html
```diff
@@ -25,51 +20,69 @@
...
+    .row .dir.over {
+      background-color: #93d4d4;
     }
-    .label-wide {
+    .row .dir {
+      white-space: nowrap;
+      -webkit-user-select: none;
+    }
+    .row .key {
+      width: 400px;
```
maffoo: This seems a little larger than we need for typical key names. I would suggest making a bit smaller and expressing in character size units, something like `16em` seems reasonable. People tend to keep registry windows small, half-screen or less, and we should leave as much room as possible for the values.

client-js/app/elements/labrad-registry.html
```diff
@@ -214,130 +250,87 @@
...
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
 
     <paper-dialog id="editValueDialog" modal>
-      <h1>Edit <span>{{selectedItem}}</span></h1>
-      <form>
-        <paper-textarea id="editValueInput" label="Value" autofocus></paper-textarea>
-      </form>
+      <h1>Edit <span>{{selected.name}}</span></h1>
```
maffoo: Add quotes around item name to be consistent with other dialogs.

client-js/app/elements/labrad-registry.ts
```diff
@@ -1,6 +1,15 @@
...
 import {Places} from '../scripts/places';
 
+type ListItem = {
+  name: string,
+  isParent: boolean,
+  isDir: boolean,
+  isKey: boolean,
+  url?: string,
+  value?: string,
+};
```
maffoo, line 11: Per the typescript language spec: "Interface types have many similarities to type aliases for object type literals, but since interface types offer more capabilities they are generally preferred to type aliases." (https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.10). Not a big deal in this case though.

client-js/app/elements/labrad-registry.ts
```diff
@@ -111,20 +148,20 @@ export class LabradRegistry extends polymer.Base {
...
 
   cursorTraverse(event) {
-    if (this.selectedIdx === null || this.getOpenDialog() || this.$.search.focused) {
+    if (this.getSelectedIndex() === null || this.getOpenDialog() || this.$.search.focused) {
       return;
     }
 
     const item = this.$.combinedList.selectedItem;
+
+    // If we have an item, we want to traverse down.
```
maffoo: This comment seems a bit out of place right before doing something if we _don't_ have an item. I'd suggest just putting a doc string on the method to explain what it does, rather than inline comments.

client-js/app/elements/labrad-registry.html
adamcw, line 179: Yes. They fire in order they appear on the page, so if you reorder them, pressing Shift+Enter on the listing will open the dialog and leave it open. Without this reordering, the window would open then immediately close because first the directory Enter would trigger, then the Shift+Enter of the dialog would close it.

client-js/app/elements/labrad-registry.html
adamcw: I'll do this with CSS instead, a bit nicer.

client-js/app/elements/labrad-registry.ts
```diff
@@ -155,21 +192,29 @@ export class LabradRegistry extends polymer.Base {
       return;
     }
 
+    // New keys have a textbox, so we only want to submit on Shift+Enter,
+    // not on a solo enter keypress.
+    if (event.detail.combo !== 'shift+enter') {
+      if (dialog.id == 'newKeyDialog' || dialog.id === 'editValueDialog') {
```
maffoo: I would do the dialog check first, since the comment notes that this is a special case for those dialogs. Also, the comment should refer to both affected dialogs, or just be generic: `// For dialogs with a textbox, ...`

client-js/app/elements/labrad-registry.ts
```diff
@@ -556,56 +624,70 @@ export class LabradRegistry extends polymer.Base {
...
   /**
    * Copy the selected key or folder.
    */
   async doCopy() {
-    var newName =  this.$.copyNameInput.value;
-    var newPath = JSON.parse(this.$.copyPathInput.value);
+    const newName = this.$.copyNameInput.value;
+    const newPath = JSON.parse(this.$.copyPathInput.value);
+
+    const selectedType = this.getSelectedType();
```
maffoo: Can you just use `this.selected` here and its attributes `isDir`, `isKey`?

client-js/app/elements/labrad-registry.ts
```diff
@@ -202,226 +247,241 @@ export class LabradRegistry extends polymer.Base {
...
   }
 
-  @property({computed: '_computeSelectedType(selectedIdx)'})
-  selectedType: string;
 
-  @property({computed: '_computeSelectedItem(selectedIdx, kick)'})
-  selectedItem: string;
-
-  _computeSelectedType(selectedIdx: number): string {
+  private getSelectedType(): string {
```
maffoo: Should be able to get rid of this by using `this.selected` instead, as suggested below.

client-js/app/elements/labrad-registry.ts
```diff
@@ -202,226 +247,241 @@ export class LabradRegistry extends polymer.Base {
...
+
+    // If there was a previously selected item, say we were editing one, then
+    // we want to reselect it after the list is repopulated. To do this we
+    // fuzzy match the objects by requiring either the name or the value to be
+    // the same. This handles both the case of renaming a dir/key or changing
+    // the value of it.
+    if (selected) {
+      for (let i = 0; i < this.filteredListItems.length; ++i) {
+        const item = this.filteredListItems[i];
+        if (item.name == selected.name || item.value == selected.value) {
```
maffoo: It would be good to check the item type here, since it's possible to have a dir and key with the same name.

client-js/app/elements/labrad-registry.ts
```diff
@@ -202,226 +247,241 @@ export class LabradRegistry extends polymer.Base {
...
+    const selected = this.selected;
+    this.applyFilterToListItems();
+
+    let index = this.getDefaultSelectedItem();
+
+    // If there was a previously selected item, say we were editing one, then
+    // we want to reselect it after the list is repopulated. To do this we
+    // fuzzy match the objects by requiring either the name or the value to be
+    // the same. This handles both the case of renaming a dir/key or changing
+    // the value of it.
```
maffoo: I don't see how this could work with renaming a dir, since there's no value to match, or at least none that uniquely identifies it compared to other dirs. I don't think we have a good way to fix that at the moment, but should change the comment to reflect the current behavior.

client-js/app/elements/labrad-registry.ts
```diff
@@ -202,226 +247,241 @@ export class LabradRegistry extends polymer.Base {
...
-  }
-
-  /**
-   * Drag and drop logic
+   * Drag and drop logic.
    */
   @listen('dragenter')
   onDragEnter(event) {
-    event.preventDefault(); //I don't understand this.
+    event.preventDefault();
```
maffoo: So I take it you understand this now? :-)

client-js/app/elements/labrad-registry.ts
```diff
@@ -202,226 +247,241 @@ export class LabradRegistry extends polymer.Base {
...
   @listen('dragstart')
   startDrag(event) {
-    //detect start of drag event, grab info about target
-    var data: any; //I tried enum, but kept getting errors...
-    data = {path: this.path, name: event.target.name, kind: event.target.className.split(' ')[0]};
+    // Detect start of drag event, grab info about target.
+    const data = {
+      path: this.path,
+      name: event.target.name,
+      kind: event.target.className.split(' ')[0]
```
maffoo, line 431: ew. We should add an attribute for this; not for this PR though.

client-js/app/elements/labrad-registry.ts
```diff
@@ -556,56 +624,70 @@ export class LabradRegistry extends polymer.Base {
...
    */
   copyClicked() {
-    var dialog = this.$.copyDialog,
-        copyNameElem = this.$.copyNameInput,
-        copyPathElem = this.$.copyPathInput;
-    copyNameElem.value = this.selectedItem;
+    const dialog = this.$.copyDialog,
+          copyNameElem = this.$.copyNameInput,
+          copyPathElem = this.$.copyPathInput;
+    copyNameElem.value = this.$.combinedList.selectedItem.name;
```
maffoo: Can use `this.selected` here?

client-js/app/elements/labrad-registry.ts
```diff
@@ -614,18 +696,17 @@ export class LabradRegistry extends polymer.Base {
...
-          case "dir":
-            var resp = await this.socket.moveDir({path: oldPath, dir: oldName, newPath: newPath, newDir: newName});
+          case 'dir':
+            await this.socket.moveDir({path: oldPath, dir: oldName, newPath: newPath, newDir: newName});
             break;
 
-          case "key":
-            var resp = await this.socket.move({path: oldPath, key: oldName, newPath: newPath, newKey: newName});
+          case 'key':
+            await this.socket.move({path: oldPath, key: oldName, newPath: newPath, newKey: newName});
```
maffoo, line 714: It's always best to keep these kinds of formatting and other superficial changes in separate commits and/or PRs. If you have the separated out in the commit history I would suggest keeping them separate when you merge.

maffoo:
A few comments, but mostly this LGTM. Also tested and it works well. It would be nice to keep the main logic changes separate from the code cleanups in the commit history (single-quotes, formatting, const instead of var, etc.). If you have that stuff separated out in the commit history, would be good to not do a full squash when merging; if not, then not a big problem.

client-js/app/elements/labrad-registry.html
adamcw, line 190: I'll do this as a separate PR that we can merge once this one gets through.

client-js/app/elements/labrad-registry.ts
adamcw: Ahh, that shouldn't even be there, that's me missing something during a merge.

maffoo:
LGTM

PR #281, opened 2016-08-11
Fix rendering issue with breadcrumbs.

adamcw:
If you had a directory trailing underscores, like `default__` this would cause a really weird display error, where the trailing underscores moved to the start of the line before `Labrad`.

This fixes that.

screenshot from 2016-08-10 17 47 00

### DIFF
--- a/client-js/app/elements/labrad-app.html
+++ b/client-js/app/elements/labrad-app.html
@@ -62,8 +62,6 @@
     #location {
       white-space: nowrap;
       text-overflow: ellipsis;
-      overflow: hidden;
-      direction: rtl;
     }
 
     #tech-info {
--- a/client-js/app/elements/labrad-app.html
+++ b/client-js/app/elements/labrad-app.html
@@ -62,6 +62,11 @@
     #location {
       white-space: nowrap;
       text-overflow: ellipsis;
+      overflow: hidden;
+      direction: rtl;
+    }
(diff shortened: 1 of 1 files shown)


maffoo:
I had actually seen that bug, believe it or not, just hadn't dug into it to figure out what was going on, because yes, it's weird. The thing I was trying to achieve with the rtl direction and hiding the overflow was to hide the left side of the breadcrumbs--the root and early path segments--when the screen gets narrow, while keeping the parent and other nearby directories are still visible. By changing the direction, the right side gets hidden, so that if the screen is narrow, you can see the root, but not the parent or near ancestors.

I played around a bit to see if there's another way to handle this and found that it's only punctuation at the end of the last path segment that gets split off; if there is punctuation on interior path segments, they render just fine. So by adding one extra invisible list element containing text we can prevent the word-breaking glitch and keep the overflow direction as I described. I've pushed this on the [`u/maffoo/breadcrumb-fix` branch](https://github.com/labrad/scalabrad-web/commit/81030b9b3a218b205c23b2ced307e944db4402f3). Let me know what you think.

adamcw:
@maffoo I fixed this by no longer using rtl, and instead floating an inner sleeve to the right to obtain the desired "overflows to the left" functionality. No zero-alpha letters required.

maffoo:
@adamcw; nice! Much less hacky...

LGTM

Issue #274, opened 2016-08-09
Display errors if submitted registry values are invalid


maffoo:
Labrad data values are entered in the registry editor in text format and sent to the server to be parsed when adding or editing keys. Currently, if this parsing fails the error message is logged in the javascript console but from the user's perspective the edit is just silently ignored. Instead we should keep the edit key/new key dialog open and display the error so the user can make changes and try again.

PR #282, opened 2016-08-11
[#274] Show errors in registry rather than failing silently.

adamcw:
Implements #274. Errors that occur when performing actions in the registry will now show errors and keep the dialog boxes open rather than closing them silently.

### DIFF
--- a/client-js/app/elements/labrad-registry.html
+++ b/client-js/app/elements/labrad-registry.html
@@ -181,7 +181,7 @@
       <!-- Dialog Controls -->
       <iron-a11y-keys target="[[target]]" keys="esc"
                       on-keys-pressed="dialogCancel"></iron-a11y-keys>
-      <iron-a11y-keys target="[[target]]" keys="shift+enter"
+      <iron-a11y-keys target="[[target]]" keys="shift+enter enter"
                       on-keys-pressed="dialogSubmit"></iron-a11y-keys>
       <!-- Cursor Controls -->
       <iron-a11y-keys target="[[target]]" keys="up down"
@@ -277,10 +277,8 @@
 
     <paper-dialog id="newKeyDialog" modal>
       <h1>Create New Key</h1>
-      <form>
-        <paper-input id="newKeyInput" label="Key" autofocus></paper-input>
-        <paper-textarea id="newValueInput" label="Value"></paper-textarea>
-      </form>
+      <paper-input id="newKeyInput" label="Key" autofocus></paper-input>
+      <paper-textarea id="newValueInput" label="Value"></paper-textarea>
       <div class="buttons">
         <paper-button on-tap="doNewKey" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
@@ -289,9 +287,7 @@
 
     <paper-dialog id="editValueDialog" modal>
       <h1>Edit <span>{{selected.name}}</span></h1>
-      <form>
-        <paper-textarea id="editValueInput" label="Value" autofocus></paper-textarea>
-      </form>
+      <paper-textarea id="editValueInput" label="Value" autofocus></paper-textarea>
       <div class="buttons">
         <paper-button on-tap="doEditValue" dialog-confirm>OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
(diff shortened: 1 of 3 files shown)


maffoo:
This is looking pretty good; definitely a big improvement. A couple things I noticed:
- If there is an error, then I successfully make an edit or cancel, the error still appears the next time the dialog is opened.
- When editing a small key in a small window, the dialog starts quite small. It expands when the error message appears, but is no longer centered. Would be nice to keep things centered. This seems to be a general problem we have with polymer dialogs, so not sure if there's a generic way to fix it or if we have to always manually resize/recenter things.

adamcw:
I was aware of both of these issues, the former is definitely fixable. The latter I'm still brainstorming.

adamcw:
@maffoo PTAL

adamcw:
@maffoo PTAL

client-js/app/elements/labrad-registry.html
```diff
@@ -277,26 +289,38 @@
       <h1>Create New Key</h1>
       <paper-input id="newKeyInput" label="Key" autofocus></paper-input>
       <paper-textarea id="newValueInput" label="Value"></paper-textarea>
+      <div class="dialogError">
+        <template is="dom-if" if="{{errorTitle}}"><h4>{{errorTitle}}</h4></template>
+        <p>{{errorMessage}}</p>
```
maffoo: Should this be inside the template too?

client-js/app/elements/labrad-registry.html
```diff
@@ -277,26 +289,38 @@
...
       <div class="buttons">
-        <paper-button on-tap="doNewKey" dialog-confirm>OK</paper-button>
+        <paper-button on-tap="doNewKey">OK</paper-button>
         <paper-button dialog-dismiss>CANCEL</paper-button>
       </div>
     </paper-dialog>
 
     <paper-dialog id="editValueDialog" modal>
-      <h1>Edit "<span>{{selected.name}}</span>"</h1>
+      <h1>Edit <span>{{selected.name}}</span></h1>
```
maffoo: Leave quotes. Probably from the crazy merge.

client-js/app/elements/labrad-registry.html
```diff
@@ -277,26 +289,38 @@
...
       <paper-textarea id="editValueInput" label="Value" autofocus></paper-textarea>
+      <div class="dialogError">
+        <template is="dom-if" if="{{errorTitle}}"><h4>{{errorTitle}}</h4></template>
+        <p>{{errorMessage}}</p>
```
maffoo: Inside template.

client-js/app/elements/labrad-registry.html
adamcw: No, the errorTitle is optional. It's a question if its worth wrapping the entire thing in a "if errorMessage", but this is otherwise as intended.

client-js/app/elements/labrad-registry.html
```diff
@@ -277,26 +289,44 @@
       <h1>Create New Key</h1>
       <paper-input id="newKeyInput" label="Key" autofocus></paper-input>
       <paper-textarea id="newValueInput" label="Value"></paper-textarea>
+      <template is="dom-if" if="{{errorMessage}}">
+      <div class="dialogError">
```
maffoo: Indent to make the structure clear. (Same below)

client-js/app/elements/labrad-registry.ts
```diff
@@ -678,49 +707,44 @@ export class LabradRegistry extends polymer.Base {
...
-        }
-      } catch (error) {
-        this.handleError(error);
-      } finally {
-        this.$.pendingDialog.close();
+      if (dragType === "copy") {
+        this.$.toastCopySuccess.show();
+      } else {
         this.$.toastMoveSuccess.show();
       }
```
maffoo, line 742: I think with the pending dialog we probably don't need the toast as well.

adamcw, line 742: But it's cute!

adamcw, line 742: I'd be more inclined to have a Start and End toast than the pending dialog. What's the point of having the task be asynchronous if you lock the user out until it finishes? They can still work in another window anyway.

maffoo, line 742: I'm ok with killing the pending dialog. Probably best would be to show a spinner on the dialog you've just used to trigger an operation. Then if the operation fails, we show an error and the user can try again. If it succeeds, the dialog just closes, no more UI changes needed.

maffoo:
LGTM. We can figure out the UI for pending operations in a separate PR.

PR #283, opened 2016-08-11
Remove selectable-table as it is now redundant.

adamcw:
With the fixes for #278 the last instance of `selectable-table` was removed from the code base. This PR cleans up this redundant code by removing it.


maffoo:
LGTM

PR #284, opened 2016-08-11
Removed dependence on class to determine item kind in Registry.

adamcw:
Until now, the type of an item is inferred from the CSS class of the item when doing drag-and-drop, this is very fickle, so this PR adds a kind attribute to the `ListItem` type to do this in a far nicer way. It also grabs the closest row to the dragged element, so that drag and drop targets now work properly regardless of where you grab them.


maffoo:
LGTM

PR #285, opened 2016-08-11
Fix no item name when double-clicking to edit registry item.

adamcw:
Currently if you double click a registry item to edit it, because this selects and then unselects the item, the opened window will have no name associated with it. This PR fixes that by setting the name programatically when called rather than binding it to the currently selected item.


maffoo:
LGTM

PR #286, opened 2016-08-12
Change grapher filter logic to match delphi version.

maffoo:
The star and trash buttons work differently in the delphi version; star limits the displayed list to _only_ starred items, while trash shows trashed items which are normally hidden in addition to the rest of the items. This changes the filter functions to match that behavior.


adamcw:
LGTM.

PR #289, opened 2016-08-12
Only run keyboard action handlers if no modifiers are pressed.

maffoo:
Even though we register the key handlers for bare keys `c`, `f`, etc, they seem to fire when pressed with modifiers, at least on mac, so that for example pressing `Cmd+f` focuses the registry filter box and prevents the browser's find-on-page box from opening. Similarly, pressing `Cmd+r` to reload the page pops up a rename dialog before then actually reloading the page. We fix this by checking that the key event has no modifiers before dispatching to the appropriate action. See https://github.com/PolymerElements/iron-a11y-keys-behavior/issues/41.

### DIFF
--- a/client-js/app/elements/labrad-registry.ts
+++ b/client-js/app/elements/labrad-registry.ts
@@ -223,7 +223,15 @@ export class LabradRegistry extends polymer.Base {
       return;
     }
 
-    const key = event.detail.combo;
+    const key = event.detail.combo,
+          keyboardEvent = event.detail.keyboardEvent,
+          hasModifiers = (keyboardEvent.shiftKey || keyboardEvent.ctrlKey ||
+                          keyboardEvent.altKey || keyboardEvent.metaKey);
+
+    if (hasModifiers) {
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


client-js/app/elements/labrad-registry.ts
```diff
@@ -223,7 +223,15 @@ export class LabradRegistry extends polymer.Base {
       return;
     }
 
-    const key = event.detail.combo;
+    const key = event.detail.combo,
+          keyboardEvent = event.detail.keyboardEvent,
+          hasModifiers = (keyboardEvent.shiftKey || keyboardEvent.ctrlKey ||
```
adamcw, line 228: Kind of horrible that the component's default behavior doesn't support this..

adamcw:
Looked into it and couldn't find a nicer way to do this, so LGTM.

client-js/app/elements/labrad-registry.ts
maffoo, line 228: I agree; very non-intuitive. I guess we could go complain on the iron-a11y-keys element (or submit a fix...).

PR #290, opened 2016-08-12
Fix selection going offscreen for some window sizes.

maffoo:
If the window is sized such that the last list element is only partially visible, that last element is included in polymer's `lastVisibleIndex` count. This happens even if only a few pixels are visible, in which case we cannot see the text and the selection appears to go offscreen. By offsetting, we ensure that the view scrolls to keep the selection on a fully-visible item. If the window size happens to be such that the final element just fits in the view, this scrolls one step earlier than it needs to, but I think keeping the selected item fully visible is more important.


maffoo:
Replaced by #291

PR #291, opened 2016-08-12
Fix item visibility on Registry and Grapher at certain window sizes.

adamcw:
An alternate solution to that suggested in #290. Scrolling on the registry and grapher at some screen resolutions would lead to the item at the bottom not always being in view. This could also happen with the top element if the list had been offset manually. This PR scrolls by the distance between the element and where it should be rather than scrolling to fixed indices.


maffoo:
LGTM

PR #292, opened 2016-08-12
Added a throttle to selectedDataset updates on Grapher.

adamcw:
Currently every time you select an item in the Grapher, the `selectedDataset` computed property makes a call to the server and then renders all the info about that dataset. With the introduction of keyboard controls, holding a key leads to a very rapid shuffle through the selected datasets which hammers the server and lags the UI due to all the DOM updates.

This PR fixes that by adding a 100ms throttle to the `selectedDataset` update. The call to the server is delayed 100ms, and any future calls made within that window will cancel the request and create a new one. This means that you only call the server at the end of a scroll, making the UI much snappier and reducing calls to the server drastically.

### DIFF
--- a/client-js/app/elements/labrad-grapher.ts
+++ b/client-js/app/elements/labrad-grapher.ts
@@ -338,15 +338,15 @@ export class LabradGrapher extends polymer.Base {
 
     this.selectedThrottleTimeout = setTimeout(() => {
       this.selectedDataset = "";
-
       if (!this.selected) {
         return;
       }
 
-      if (this.selected.id.startsWith('dataset:')) {
-        const name = this.selected.id.substring(8);
-        this.fetchInfo(name);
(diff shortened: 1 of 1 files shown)


client-js/app/elements/labrad-grapher.ts
```diff
@@ -46,6 +51,8 @@ export class LabradGrapher extends polymer.Base {
   @property({type: String})
   connectionError: string;
 
+  selectedThrottleTimeout: any = null;
```
maffoo: The standard library `setTimeout` function just returns `number` (https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L18070).

maffoo:
Update the type on `selectedThrottleTimeout`, then LGTM.

PR #295, opened 2016-08-12
Change manager/node colors to be in line with the rest of the app.

adamcw:
Changes the styling of the Manager/Node panels slightly to make it more consistent with the rest of the application.


maffoo:
LGTM

PR #300, opened 2016-08-17
Raise an error if saved credentials do not have a known 'kind'

### LINKED ISSUE #299: Explicit logout required with v2 in browser window previously logged in to v1
Trying to load v2 results in blank window with no error message (see #297) and requires user to explicitly log out. Need to investigate further why this is happening.

maffoo:
This fixes a problem with logging in to scalabrad-web v2 in a browser window that had previously been logged in to v1, since credentials were previously stored without a 'kind' property. These old credentials will now be rejected, so that the user will have to log in again when connecting to v2, but will at least be presented with a login box instead of a blank screen.

Fixes #299

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -431,6 +431,9 @@ window.addEventListener('WebComponentsReady', () => {
             manager: manager
           });
           break;
+
+        default:
+          throw Error(`Unknown credential type: ${credential.kind}`);
       }
     } catch (error) {
       if (isPasswordError(error)) {


maffoo:
@joshmutus, is this something you can try out?

To test this I did the following:
- checkout `v1.1.0`, run it locally and log in to a browser window
- checkout `v2.0.0`, run it locally, and try to load it in the same browser window. You should see a blank window. There is a "not connected" error in the javascript console.
- checkout `u/maffoo/v2-login-fix` (this PR), run it locally, and load it in the same browser window. You should see a login box and be able to log in.

pklimov:
LGTM

PR #301, opened 2016-08-18
Fix splicing of lastData row in grapher

### LINKED ISSUE #298: Grapher sometimes misses points when datasets are updated live
Had reports of this in v2.0.0. Need to investigate to determine what is going on.

maffoo:
The second argument to splice is the number of items to delete from the array. When splicing in the lastData row we were deleting the next row, and if there was only one row to begin with this would result in a data array of length 1, causing createLine to return null and the data point not to be plotted. This problem becomes more pronounced when the data arrives slowly, one row at a time, as often happens when taking real- world experimental data. Added a slow demo scan to test this scenario.

Fixes #298

### DIFF
--- a/client-js/app/elements/labrad-plot.ts
+++ b/client-js/app/elements/labrad-plot.ts
@@ -859,7 +859,7 @@ export class Plot extends polymer.Base {
     this.dataLimits1D(data);
 
     if (this.lastData) {
-      data.splice(0, 1, this.lastData);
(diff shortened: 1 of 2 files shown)


maffoo:
@adamcw, can you take a look? Thankfully, the fix is super simple :-)

adamcw:
Glad this didn't turn out to be a complicated fix, I must have confused the remove length with the insertion length, oops! LGTM.

PR #304, opened 2016-09-15
Make oauth login more robust

maffoo:
It would be nice to have more structured error handling with proper error codes to distinguish different problems, but this is a quick fix to make oauth login flow work after oauth token has expired. Also fixes the calculation of the expiration time due to milliseconds vs. seconds issue.


kunalq (approved):
LGTM

PR #305, opened 2016-09-19
Add support for using OAuth access token to log in

maffoo:
Previously, we used id tokens for oauth login, but the manager supports access tokens as of labrad/scalabrad#75, so we add support for that here as well. Because the server gives us a proper expiration time for the access token, we can restart the oauth login flow automatically when the browser has an expired access token, which should reduce the number of times users have to click the "Login with google" button, especially users who keep scalabrad-web windows open for a long time.

### DIFF
--- a/client-js/app/scripts/app.ts
+++ b/client-js/app/scripts/app.ts
@@ -319,7 +319,8 @@ window.addEventListener('WebComponentsReady', () => {
     const authMethods = await mgr.authMethods({manager: manager});
     app.host = manager;
     app.allowUsernameLogin = authMethods.indexOf('username+password') >= 0;
-    app.allowOAuthLogin = authMethods.indexOf('oauth_token') >= 0;
+    app.allowOAuthLogin = authMethods.indexOf('oauth_token') >= 0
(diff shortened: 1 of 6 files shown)


client-js/app/scripts/app.ts
```diff
@@ -319,49 +319,57 @@ window.addEventListener('WebComponentsReady', () => {
...
-      window.location.href = url;
-    } catch (error) {
-      app.loginError = error.message || error;
-      setTimeout(() => app.$.passwordInput.$.input.focus(), 0);
-    }
+  /**
+   * Start OAuth login flow by redirecting to oauth login page. Returns a
+   * promise that never fires since there's nothing to do after the redirect.
+   */
+  async function startOAuthLogin(manager: string, oauthInfo: OAuthInfo, tokenType: string, rememberMe: boolean) {
```
maffoo, line 352: Moved references to `app` into the login dialog event listeners above. Otherwise, this is mostly just outdenting the existing function.

kunalq (approved):
LGTM :)

Issue #306, opened 2016-09-20
Dataset view fails on 0-dimensional datasets


maffoo:
Datasets can be created in the data vault with 0 independent variables. This breaks the dataset view and the grapher live view.

Issue #303, opened 2016-08-31
Cursor position y-coordinate in grapher live view incorrect


christopherquintana:
For example, it will display the lowest y-coordinate of the plot independent of position of mouse over the graph. It may be assuming the graph is in a different location than it really is.

PR #307, opened 2016-09-20
Miscellaneous plot fixes

maffoo:
Fixes a couple of issues with plots, including #303 and #306. Also fixes a three.js error that shows up in the console when data is coming in due to trying to add null lines to the scene graph.


kunalq (approved):
LGTM

PR #309, opened 2016-09-22
Update typescript to version 2.

maffoo:
Lets us remove some type casts when dealing with credentials which are a tagged union type distinguished by the `kind` member. We have to add a check when we deserialize credentials from JSON, to ensure that the object has a valid `kind` field, but from there the type system takes over.


kunalq:
LGTM :)

PR #314, opened 2016-10-12
Disable browserSync's ghostMode to avoid window interference.

### LINKED ISSUE #79: Events leak between open tabs/windows in dev mode
When running in dev mode with multiple windows open, events can leak between open windows. For example, if I open two windows and navigate to the same url in each (say, `http://localhost:3000/registry/`) and then tap the hamburger button in one window to open the drawer panel, it will actually slide open in both tabs. I would guess this is happening due to the way browserSync (used for live reloading) is loading the page. It doesn't happen in production mode as far as I can tell, so it's not a problem with our code, but it is kind of annoying when testing things with multiple windows (like drag and drop between windows). Low priority to look into this and see if there's a fix.

maffoo:
We're using browserSync for live reloading in development mode, but
by default it enables "ghostMode" which propagates clicks, key
presses and other UI interactions between the various open windows.
We disable this so that multiple open windows are isolated.

Fixes #79

### DIFF
--- a/client-js/gulpfile.js
+++ b/client-js/gulpfile.js
@@ -11,7 +11,7 @@ var tsc = require('gulp-typescript');
 var typescript = require('typescript');
 var del = require('del');
 var runSequence = require('run-sequence');
-var browserSync = require('browser-sync');
+var browserSync = require('browser-sync').create();
 var merge = require('merge-stream');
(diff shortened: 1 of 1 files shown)


kunalq (approved):
LGTM -- that's an unusual feature

PR #323, opened 2017-02-11
Fix calling methods on plot element

maffoo:
A DatasetActivity may not have a dataset if the data cannot be
plotted (for example, if the dataset is 0- or 3-dimensional).


kunalq (approved):
LGTM :)

PR #329, opened 2017-07-25
Add support for restarting outdated servers

maffoo:
This also cleans up the implementation of the LabradNodes element. Among other things, we move the fetching of extra information like the autostart list and list of outdated servers into the NodeApi so that the ui element can just use the status info it gets without making additional api calls. As a result of these cleanups the nodes element is much more robust, for example the displayed server versions update properly when the node config is refreshed. This was a persistent annoyance with the previous implementation since it made it even harder to understand what versions of server were running where.

The outdated server support relies on the node changes here: https://github.com/labrad/pylabrad/pull/343. The web interface will still work with older nodes, but the outdated restart button will do nothing.

Unfortunately, the diff is pretty large; sorry about that.

### DIFF
--- a/client-js/app/elements/labrad-nodes.ts
+++ b/client-js/app/elements/labrad-nodes.ts
@@ -275,7 +275,7 @@ interface ServerInfo {
   errorString: string;
   errorException: string;
   nodes: NodeServerStatus[];
-};
+}
 
 type NodeServerStatus = {
   name: string,
--- a/client-js/app/elements/labrad-nodes.ts
+++ b/client-js/app/elements/labrad-nodes.ts
@@ -34,6 +34,9 @@ export class LabradInstanceController extends polymer.Base {
   @property()
   node: string;
 
+  @property({type: String})
+  version: string;
+
   @property({type: Boolean})
   local: boolean;
 
@@ -250,6 +253,18 @@ export class LabradNodeController extends polymer.Base {
       this.active = false;
     }
   }
+
+  @listen('outdated.click')
+  async onRestartOutdated() {
+    this.active = true;
+    try {
+      await this.api.outdatedRestart(this.name);
+    } catch (e) {
+      console.error(`Exception while restarting outdated servers: ${this.name}`, e);
+    } finally {
+      this.active = false;
+    }
+  }
 }
 
 /**
@@ -257,16 +272,17 @@ export class LabradNodeController extends polymer.Base {
  */
 interface ServerInfo {
   name: string;
-  version: string;
-  autostart: boolean;
   errorString: string;
   errorException: string;
   nodes: NodeServerStatus[];
-}
+};
 
 type NodeServerStatus = {
   name: string,
   exists: boolean,
+  version?: string,
+  runningVersion?: string,
+  outdated?: boolean,
   autostart?: boolean,
   status?: string,
   instanceName?: string
@@ -278,7 +294,7 @@ type ServerFilterFunction = (item: ServerInfo) => boolean;
 @component('labrad-nodes')
 export class LabradNodes extends polymer.Base {
   @property({type: Array, notify: true, value: () => []})
-  info: NodeStatus[];
+  nodes: NodeStatus[];
 
   @property({type: Object})
   api: NodeApi;
@@ -315,7 +331,7 @@ export class LabradNodes extends polymer.Base {
     // the state of each controller of that server type to all others of the
     // same type. This informs existing servers of the new server state, and
     // informs the new server of the existing servers' states.
-    this.onServerStatus(event.detail);
+    this.broadcastStatusChange(event.detail);
   }
 
 
@@ -344,6 +360,15 @@ export class LabradNodes extends polymer.Base {
   }
 
 
+  private filterServersFunction(): ServerFilterFunction {
+    if (this.isAutostartFiltered) {
+      // Check if the server is configured to autostart on any node.
+      return (s) => s.nodes.some((node) => node.exists && node.autostart);
+    }
+    return (s) => true;
+  }
+
+
   // Listeners called on api object properties when they get set. This is
   // necessary because these properties are set asynchronously sometime after
   // the component is constructed, so we cannot access the values when the
@@ -369,21 +394,23 @@ export class LabradNodes extends polymer.Base {
 
 
   // Callbacks invoked when we receive remote messages.
-  async onNodeStatus(msg: NodeStatus) {
-    msg.autostartList = await this.api.autostartList(msg.name);
-    this.addItemToList(msg);
+
+  // Called when the status of a node changes.
+  async onNodeStatus(node: NodeStatus): Promise<void> {
+    this.updateNode(node);
   }
 
 
+  // Called when a server disconnects from labrad
   onServerDisconnected(msg: ServerDisconnectMessage): void {
     console.warn('Server disconnected:', msg.name);
 
     // If the disconnected server is a node, broadcast to all controllers that
     // the servers on the node are now offline.
-    for (const node of this.info) {
+    for (const node of this.nodes) {
       if (node.name == msg.name) {
         for (const server of node.servers) {
-          this.onServerStatus({
+          this.broadcastStatusChange({
             node: node.name,
             server: server.name,
             instance: '',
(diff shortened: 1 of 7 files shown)


server/src/main/scala/org/labrad/browser/ApiBackend.scala
```diff
@@ -78,11 +78,14 @@ class ApiBackend(config: LabradConnectionConfig)(implicit ec: ExecutionContext)
...
 
       CALL  org.labrad.node.allNodes       nodeApi.allNodes
+      CALL                 .getNodeStatus         .getNodeStatus
       CALL                 .refreshNode           .refreshNode
       CALL                 .autostartNode         .autostartNode
       CALL                 .autostartList         .autostartList
       CALL                 .autostartAdd          .autostartAdd
       CALL                 .autostartRemove       .autostartRemove
+      CALL                 .outdatedList          .outdatedList
+      CALL                 .outdatedRestart       .outdatedRestart
```
kunalq, line 88: I'm going to assume there isn't much feedback I can give here and gloss over these changes :+1:

client-js/app/elements/labrad-nodes.ts
```diff
@@ -250,23 +253,36 @@ export class LabradNodeController extends polymer.Base {
...
  */
 interface ServerInfo {
   name: string;
-  version: string;
-  autostart: boolean;
   errorString: string;
   errorException: string;
   nodes: NodeServerStatus[];
-}
+};
```
kunalq: I don't believe the semicolon is required, but I might be wrong here :)

client-js/app/elements/labrad-nodes.ts
```diff
@@ -408,15 +421,31 @@ export class LabradNodes extends polymer.Base {
...
+  }
+
+
+  // Broadcast status change to other instance controllers of the same server.
+  async broadcastStatusChange(msg: ServerStatusMessage): Promise<void> {
     const instances = Polymer.dom(this.root)
                              .querySelectorAll('labrad-instance-controller');
     for (const inst of instances) {
-      const instance = <any>inst;
+      const instance = inst as LabradInstanceController;
```
kunalq, line 448: Sorry for the dumb question, but is this casting a DOM element back into a Polymer component?

client-js/app/elements/labrad-nodes.ts
```diff
@@ -547,54 +564,47 @@ export class LabradNodes extends polymer.Base {
...
-      serverStatusMap.set(s.name, s);
+  private updateServerLists(): void {
+    const serverLists = [
+      'globalServersFiltered',
+      'globalServers',
+      'localServersFiltered',
+      'localServers'
+    ];
+    for (const listName of serverLists) {
+      const servers = this[listName];
```
kunalq, line 575: Oh interesting -- I didn't realize TypeScript preserved the names of properties as regular attributes of the class object.

client-js/app/elements/labrad-nodes.ts
```diff
@@ -626,47 +636,43 @@ export class LabradNodes extends polymer.Base {
...
+    for (let nodeIdx = 0; nodeIdx < server.nodes.length; nodeIdx++) {
+      const node = server.nodes[nodeIdx];
+      if (node.exists) {
+        const status = this.getServerStatus(node.name, server.name);
+        const version = status.outdated
+                      ? `running=${status.runningVersion}, latest=${status.latestVersion}`
+                      : status.version;
+        if (node.version !== version) {
+          this.set(`${listName}.#${serverIdx}.nodes.#${nodeIdx}.version`,
+                   version);
```
kunalq, line 650: This syntax is unfortunate, but I suppose there's really no easy way around it.

client-js/app/scripts/node.ts
```diff
@@ -44,17 +57,77 @@ export interface NodeApi {
...
-  allNodes(): Promise<NodeStatus[]> {
-    return this.call<NodeStatus[]>('allNodes', []);
+  // Fetch additional info needed along with the basic node status.
+  //
+  // Extra information like the autostart list and list of outdated servers is
+  // needed by the node interface but not included in status messages from the
+  // node server. This function grabs the extra data. We do this here in the
+  // NodeApi so other code can use the NodeStatus type without having to
+  // manually fetch additional data.
+  private async updateNodeStatus(nodeStatus: NodeStatus): Promise<NodeStatus> {
```
kunalq, line 81: For this method, it looks like it takes in a `NodeStatus`, mutates it (or at least its member list `servers`?), then returns that object. Is this for performance reasons? i.e. would creating a new `NodeStatus` each time allocate too many of these that would then need to be garbage collected?

server/src/main/scala/org/labrad/browser/NodeApi.scala
```diff
@@ -15,6 +15,9 @@ object NodeApi {
   case class NodeStatus(name: String, servers: Seq[ServerStatus])
   object NodeStatus { implicit val format = Json.format[NodeStatus] }
 
+  case class OutdatedServerInfo(name: String, instanceName: String, runningVersion: String, latestVersion: String)
+  object OutdatedServerInfo { implicit val format = Json.format[OutdatedServerInfo] }
```
kunalq, line 19: neat!

server/src/main/scala/org/labrad/browser/NodeApi.scala
```diff
@@ -43,6 +46,12 @@ class NodeApi(cxn: LabradConnection)(implicit ec: ExecutionContext) extends Logg
     }
   }
 
+  def getNodeStatus(node: String): Future[NodeStatus] = {
+    cxn.to(node).call("status").map { servers =>
+      NodeStatus(node, getServerStatuses(servers))
+    }
```
kunalq, line 52: I'm trying to understand what happens if this call fails -- I see the `call` in `rpc.ts`, which returns a `Promise`. If this call fails here, and the returned `Future` is set to failure/error, I think I might be missing where the client side error ends up (i.e. I can't find anywhere where the Promise has a `.catch(...)` errback).

kunalq (commented):
This looks like a lot of work, Matthew -- thanks for going through the trouble of adding this! My one concern is error handling -- I think maybe I'm missing it, but if there was a way to globally redirect all failed promises to at least log something in the Javascript console. Maybe by changing the JSON-RPC interface?

client-js/app/elements/labrad-nodes.ts
maffoo: removed

client-js/app/elements/labrad-nodes.ts
maffoo: In this case the object we get from the DOM actually _is_ a polymer component, the typescript compiler just doesn't know that because we got it through the untyped `querySelectorAll` interface. We need a cast to convince the compiler that this has a `.name` attribute, and I figure casting to the actual type is better than casting to `any` :-)

client-js/app/elements/labrad-nodes.ts
maffoo: Yeah, it's still just like javascript in that regard.

client-js/app/elements/labrad-nodes.ts
maffoo: Unfortunate is an understatement :-)

What I really want here is to use react, so we can just build up a complete model of the state of all the nodes after every event and then tell it to re-render the UI, and let _it_ figure out the minimal set of DOM changes that are needed to get the UI in sync. Here, we have to redirect all the state changes through polymer's `set` function with this complicated path syntax so that the polymer data binding infrastructure is aware of the changes we are making. It's really gross.

Incidentally, I looked into what it would take to drop react in here and it would require a bunch of changes to the build system to get that working, but stay tuned for future PRs :-)

client-js/app/scripts/node.ts
maffoo: I think creating a new `NodeStatus` object is not necessary because this update function is only used internally to the NodeApi class to update status messages with additional info, and the non-updated messages are never exposed anywhere, so we'd just be creating extra work for the GC by keeping them around (I don't think it'd be a huge amount of extra work, but still).

server/src/main/scala/org/labrad/browser/NodeApi.scala
maffoo: The only place `getNodeStatus` gets called on the client is [here](https://github.com/labrad/scalabrad-web/pull/329/files#diff-b404fdff99d6a1c62690fd08a74dac31R435). Any exceptions raised on the server will result in a failed scala Future, which then sends the error over json-rpc and results in a failed typescript Promise, which then gets raised as an exception in our async function when we await that promise. In this particular case, we log an error to the js console. Some day we'll get around to surfacing all these errors in the UI :-)

client-js/app/elements/labrad-nodes.ts
kunalq: Might not be worth the effort, but is it possible to mutate the array normally, then call [`notifySplices`](https://www.polymer-project.org/2.0/docs/devguide/model-data#notifysplices) to refresh the whole thing at once?

kunalq (approved):
Thanks for the explanations! LGTM :)

client-js/app/elements/labrad-nodes.ts
maffoo: That doesn't work for changing properties on the elements within the arrays, which is what we're doing here. Also, if I'm reading the [docs](https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#method-notifySplices) right we have to provide the `notifySplices` method with a list of splice objects describing the changes, so it's probably easier to just use the polymer splice methods to modify the arrays in the first place.

PR #339, opened 2020-06-26
Upgrade elements to Polymer 3

gschaffner:
This removes almost all HTML imports from the repo.

I'm also specifying specific `@polymer/{paper,iron}-{list,etc}` explicitly in `package.json` since requiring `@polymer/{paper,iron}-elements` appears to be unsupported now.

There is still one HTML import in the repo: https://github.com/labrad/scalabrad-web/blob/aef3429f91b146ee367dbf00709139afa81c1352/client-js/app/index.html#L38


client-js/app/elements/app-link.ts
```diff
@@ -1,8 +1,11 @@
...
+import {customElement, property} from "@polymer/decorators";
+
 const MOUSE_MAIN_BUTTON = 0;
 
-@component('app-link')
-@extend('a')
-export class AppLink extends polymer.Base {
+@customElement('app-link')
+// @extend('a')  // TODO: https://polymer-library.polymer-project.org/3.0/docs/devguide/custom-elements#extending-elements
+export class AppLink extends PolymerElement {
```
gschaffner, line 8: The fix for this might be to `extend HTMLLinkElement` instead of `PolymerElement`. However: "Polymer does not currently support extending built-in elements." (https://polymer-library.polymer-project.org/3.0/docs/devguide/custom-elements#extending-elements)

client-js/app/elements/expandable-input.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  //holds the shortened value which renders as 'value...' if longer than cols
+  @property({type: String, computed: 'shortenValue(value)'})
+  valueShort: String;
+
+  function() {
+    //This focus-changed event handles the revealing and hiding of the static field
+    this.addEventListener('focused-changed', this.fieldFocus);
+    this.$.long.$.input.$.textarea.addEventListener('keypress', this.keyPressed);
+  }
```
gschaffner, line 55: expandable-input and some other elements used `this.` frequently in their `expandable-input.html`'s `<script> Polymer({ ... }) </script>` block. In attempting to update, I just copied the use of `this.` from the `<script> Polymer({ ... }) </script>` blocks. However, I am not sure that this copying has led to the correct use of `this.` in the `.ts` files.

gschaffner:
Should we keep element HTML outside of the `.ts` files? It looks like this may be possible: https://stackoverflow.com/questions/46393169/import-template-from-separate-html-file-in-polymer-3

client-js/app/elements/app-link.ts
maffoo: Yeah, we'll have to just try to match as much of the behavior of the built-in `<a>` element as possible, since we can't actually extend `<a>`.

maffoo (approved):
This is great work; thanks! I'm going to merge and work on getting the polymer3 branch to build.

maffoo:
Re: having separate `.html` files, I originally thought that would be better, but I've kinda gotten used to this style of inline templates and keeping the template close to the code. One advantage is that it makes us less dependent on a particular build system since we _just_ have to compile typescript, not also process special kinds of includes. I'd say stick with this style for now.

